PR #9, opened 2023-06-09
updated package.json name, version, description

frasergr:


frasergr:
Sorry this is no good, I thought it was a simple change but there's other parts to this I hadn't tested!

PR #19, opened 2023-06-10
fix: convert insert loop to 1 insert stmt

jwaltz:
I was getting errors when inserting a large .docx (`"wordCount": 225518, "token_count_estimate": 47635`) file's vectorIds into the document_vectors table:

`Error: SQLITE_BUSY: database is locked`

I took a look at the `bulkInsert()` method in `server/models/vectors.js` and noticed it was executing `INSERT` statements in a loop. Sqlite supports batch inserts so I converted the `INSERT` statement to a single insert with multiple parameters to account for all of the vectorIds.

My LSP formatter/linter might have changed a bit of the formatting of the method signatures but I hope the PR is still valuable/valid. Let me know if I can fix anything.

Thanks!

### DIFF
--- a/server/models/vectors.js
+++ b/server/models/vectors.js
@@ -12,14 +12,13 @@ const DocumentVectors = {
   createdAt TEXT DEFAULT CURRENT_TIMESTAMP,
   lastUpdatedAt TEXT DEFAULT CURRENT_TIMESTAMP
   `,
-  db: async function () {
+  db: async function() {
     const sqlite3 = require("sqlite3").verbose();
     const { open } = require("sqlite");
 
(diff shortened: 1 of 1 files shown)


timothycarambat:
@jwaltz any chance you can run `yarn lint` from the root to lint all files then re-commit those changes? Theres a lot of delta and its mostly formatting so its hard to see the true delta.

Appreciate you 🙏

jwaltz:
@timothycarambat sure thing, linted and added commit.

timothycarambat (approved):
You are awesome - this will help resolve a ton of issues 🙏

PR #24, opened 2023-06-10
refactor: convert chunk embedding to one API call

jwaltz:
I was having really long embed times for a couple large .docx documents ingested through the `/hotdir`. One had 256 chunks and took many minutes to fully embed, sometimes never finishing at all.

I converted the `embedChunk()` method to accept a list of chunks instead of just one chunk, and make a single call to OpenAI's embedding API endpoint, which seems to have sped up the embed process dramatically. The refactored `embedChunks()` required some changes to its usage in `addDocumentToNamespace()` as well. I included these changes for the 3 vector db options at `/server/utils/vectorDbProviders/[vectorDb]/index.js`

Let me know what you think and if this is helpful. Again, apologies for my editor's lsp changing some of the method signatures and line-length formatting, hope it is ok.

### DIFF
--- a/server/utils/vectorDbProviders/chroma/index.js
+++ b/server/utils/vectorDbProviders/chroma/index.js
@@ -11,7 +11,7 @@ const { toChunks, curateSources } = require("../../helpers");
 
 const Chroma = {
   name: "Chroma",
-  connect: async function() {
+  connect: async function () {
     if (process.env.VECTOR_DB !== "chroma")
       throw new Error("Chroma::Invalid ENV settings");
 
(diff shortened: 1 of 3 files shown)


timothycarambat:
LFG - will review. Thank you!

jwaltz:
@timothycarambat no problem, linted and committed.

AntonioCiolino:
is there a reason this isn't merges (other than merge conflicts)? Anything that could enhance speed would be helpful.

timothycarambat:
Moved to #153

PR #62, opened 2023-06-15
Remove `maxLength` in chat input box to allow larger prompts

preeteshjain:
Can't send larger messages because of this artificial limit.

ChatGPT also doesn't have any upper limit to what we can send as message.

### DIFF
--- a/frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
@@ -55,7 +55,6 @@ export default function PromptInput({
             onKeyDown={captureEnter}
             onChange={onChange}
             required={true}
-            maxLength={240}
             disabled={inputDisabled}
             onFocus={() => setFocused(true)}
             onBlur={(e) => {


timothycarambat:
The issue with doing this will be that when we have to do manual similarity searches across a vector space we have to pre-chunk the query if its over a certain length to vectorize. Turning this off like this will result in someone sending a message that will fail because the query will be too long to do a similarity search without pre-chunking for when we are not using langchain - which is just lance

preeteshjain:
Okay, now I understand. But to really make this an alternative to ChatGPT, we need to increase this limit.

CerxMe:
Split the query into multiple parts to make prompts and then generate text summarizations.
Use those to piece together whatever the original message says and *feed that in* to generate results.

AntonioCiolino:
to work around this (b ecause 240 was pretty unusable) I changed it to 500 on my local instance. 500 is generally ok.

franzbischoff:
500 seems to work ok here too.
@preeteshjain I get the problem and I've increased the value by myself before actually looking here and see what it was so small. I'm not in the team, but I would like to say that "make this an alternative to ChatGPT" is not the goal here. This project (for now at least) has the big plus on making easy to make your own personal knowledge dataset and "talk to your documents". Until now this is the best project I've seen around that actually works and only requires one paying API: the openai's.

PR #82, opened 2023-06-18
bump LanceDB [vectordb] to latest for stability reasons

timothycarambat:


timothycarambat:
0.1.8 looks to be broken and non-responsive to any client commands on Mac arm64. Will try in Docker container next to see if working on that platform

PR #126, opened 2023-06-27
History per workspace

AntonioCiolino:
Merge for Adding chat history window into the code base.

### DIFF
--- a/frontend/src/components/Modals/MangeWorkspace/Settings/index.jsx
+++ b/frontend/src/components/Modals/MangeWorkspace/Settings/index.jsx
@@ -140,6 +140,40 @@ export default function WorkspaceSettings({ workspace }) {
                 onChange={() => setHasChanges(true)}
               />
             </div>
+
+            <div>
+              <div className="flex flex-col gap-y-1 mb-4">
+                <label
+                  htmlFor="name"
+                  className="block text-sm font-medium text-gray-900 dark:text-white"
+                >
+                  Chat History
+                </label>
+                <p className="text-xs text-gray-600 dark:text-stone-400">
+                  Chat history: how far back in teh current discussion will 
+                  chat pull responses.
+                  <br />
+                  Default 20. Anything more than 45 is likely to cause significant failures
+                  as well as higher costs.
+                  <br />
+                  Recommended: 20
+                </p>
+              </div>
+              <input
+                name="openAiHistory"
+                type="number"
+                min={0.0}
+                max={45.0}
+                step={1}
+                onWheel={(e) => e.target.blur()}
+                defaultValue={workspace?.openAiHistory ?? 20}
+                className="bg-gray-50 border border-gray-300 text-gray-900 text-sm rounded-lg focus:ring-blue-500 focus:border-blue-500 block w-full p-2.5 dark:bg-stone-600 dark:border-stone-600 dark:placeholder-gray-400 dark:text-white dark:focus:ring-blue-500 dark:focus:border-blue-500"
+                placeholder="20"
+                required={true}
+                autoComplete="off"
+                onChange={() => setHasChanges(true)}
+              />
+            </div>
           </div>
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/Modals/MangeWorkspace/Settings/index.jsx
```diff
@@ -140,6 +140,40 @@ export default function WorkspaceSettings({ workspace }) {
...
+                  Chat History
+                </label>
+                <p className="text-xs text-gray-600 dark:text-stone-400">
+                  Chat history: how far back in teh current discussion will 
+                  chat pull responses.
+                  <br />
+                  Default 20. Anything more than 45 is likely to cause significant failures
+                  as well as higher costs.
+                  <br />
+                  Recommended: 20
```
timothycarambat, line 159: Lets reword this to be more concise
```
The number of previous chats that will be included in the response's short-term memory.
Recommend: 20. Anything more than 45 is likely to lead to continuous chat failures depending on message size.

File: server/models/workspace.js [hunk 2]
```diff
@@ -104,9 +111,10 @@ const Workspace = {
       this.writable.includes(key)
     );
     const values = Object.values(data);
+    // console.log(validKeys);
```
timothycarambat, line 114: can we remove this commented out console log?

server/utils/chats/index.js
```diff
@@ -104,7 +104,9 @@ async function chatWithWorkspace(workspace, message, chatMode = "chat") {
       error: null,
     };
   } else {
-    const rawHistory = await WorkspaceChats.forWorkspace(workspace.id, 20);
+    var chat_memory = workspace?.openAiHistory;
+    // console.debug(chat_memory);
```
timothycarambat, line 108: can we remove this commented out console log?

timothycarambat: Also for JS code we utilize `camelCase` naming convention. I also think a name like `messageLimit` would be more indicative of what it is doing as chat memory sound like actual memory objects.

server/models/workspace.js
```diff
@@ -42,6 +44,11 @@ const Workspace = {
                                  END`,
         doif: true,
       },
+      {
+        colName: "openAiHistory",
+        execCmd: `ALTER TABLE ${this.tablename} ADD COLUMN openAiHistory INTEGER DEFAULT 20`,
+        doif: false,
+      },
```
timothycarambat, line 51: 👍

frontend/src/components/Modals/MangeWorkspace/Settings/index.jsx
```diff
@@ -140,6 +140,40 @@ export default function WorkspaceSettings({ workspace }) {
...
+                </p>
+              </div>
+              <input
+                name="openAiHistory"
+                type="number"
+                min={0.0}
+                max={45.0}
```
timothycarambat, line 166: These can be whole numbers since step is 1! Not a problem if left though

timothycarambat (changes requested):
Functions well and i tested on my end - just some superficial changes but otherwise looks good once those are down!

frontend/src/components/Modals/MangeWorkspace/Settings/index.jsx
AntonioCiolino: Changed

frontend/src/components/Modals/MangeWorkspace/Settings/index.jsx
AntonioCiolino: updated to ints

server/models/workspace.js
AntonioCiolino: deleted.

server/utils/chats/index.js
AntonioCiolino: camelCased and removed comment

PR #134, opened 2023-06-29
Twitter Feature

AntonioCiolino:
Commit to enable twitter in the collector.

### DIFF
--- a/collector/scripts/twitter.py
+++ b/collector/scripts/twitter.py
@@ -14,11 +14,15 @@ def twitter():
     searchQuery = input("Search term, or leave blank to get user tweets (blank to ignore): ")
     tweetCount = input("Gather the last number of tweets: ")
     
-    #read your API kes to call the api.
-    consumer_key = os.environ["TW_CONSUMER_KEY"]
-    consumer_secret = os.environ["TW_CONSUMER_SECRET"]
-    access_token = os.environ["TW_ACCESS_TOKEN"]
-    access_token_secret = os.environ["TW_ACCESS_TOKEN_SECRET"]
+    # Read your API keys to call the API.
+    consumer_key = os.environ.get("TW_CONSUMER_KEY")
+    consumer_secret = os.environ.get("TW_CONSUMER_SECRET")
+    access_token = os.environ.get("TW_ACCESS_TOKEN")
+    access_token_secret = os.environ.get("TW_ACCESS_TOKEN_SECRET")
+
+    # Check if any of the required environment variables is missing.
+    if not consumer_key or not consumer_secret or not access_token or not access_token_secret:
+        raise EnvironmentError("One of the twitter API environment variables are missing.")
 
     # Pass in our twitter API authentication key
     auth = tweepy.OAuth1UserHandler(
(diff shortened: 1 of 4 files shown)


timothycarambat:
@AntonioCiolino Does this still work even though Elon killed the free Twitter tier??

AntonioCiolino:
I expect so but i didn’t test it yesterday. But… this does still require an API key so it probably isn’t really anonymous.

AntonioCiolino:
Plus he’ll likely roll it back…lol

collector/scripts/twitter.py
```diff
@@ -0,0 +1,95 @@
...
+import json
+from .utils import tokenize, ada_v2_cost
+
+def twitter():
+    #get user and number of tweets to read
+    username = input("user timeline to read from (blank to ignore): ")
+    searchQuery = input("Search term, or leave blank to get user tweets (blank to ignore): ")
+    tweetCount = input("Gather the last number of tweets: ")
+    
+    #read your API kes to call the api.
```
timothycarambat, line 17:
```suggestion
    #read your API keys to call the api.
```

We should check the existence of the keys before invocation and throw an error if any are missing - right now we just try to do the OAuth and hope for the best - if anything the `auth` and `api` should be inside of the try-catch block as well so at least that gets handled.

collector/scripts/twitter.py
```diff
@@ -0,0 +1,95 @@
...
+  url = f"http://twitter.com/anyuser/status/{row['id']}"
+  title = f"Tweet {row['id']}"
+  meta = {
+    'url': url,
+    'title':  title,
+    'description': 'Tweet from ' + row["Screen Name"],
+    'published': row["Date Created"].strftime('%Y-%m-%d %H:%M:%S'),
+    'wordCount': len(row["Tweet"]),
+  }
+  return "Tweet JSON Metadata:\n"+json.dumps(meta)+"\n\n\nText Content:\n" + row["Tweet"] if metadata_only == False else meta
```
timothycarambat, line 95: I am fairly sure this function does not work because `http://twitter.com/anyuser/status/{row['id']}` will try to resolve the `@anyuser` handle doing this - which won't work since the users handle can be anything.

timothycarambat (changes requested):

timothycarambat:
Iffy to merge this because we cannot test it since the API doesn't work without paying hundreds/thousands to do it. Maybe we should just kill the feature entirely and leave twitter unsupported and/or find a reliable way to web scrape as many others do.

collector/scripts/twitter.py
AntonioCiolino: This does work since I’ve used it to scrape. This is a known workaround in Twitter; it’ll redirect to the right user ID you know the i I d.

collector/scripts/twitter.py
AntonioCiolino: Good point.

collector/scripts/twitter.py
timothycarambat: Ah, good to know. Had no idea!

AntonioCiolino:
updated and validate that the process is still working with API keys.

timothycarambat (approved):

PR #148, opened 2023-07-15
Adding Digital Ocean Deployment using Terraform

ajosegun:
This PR introduces a new Terraform configuration that enables users to easily deploy a private AnythingLLM instance on DigitalOcean.


timothycarambat:
The 🐐 returns with another deployment PR. Thank you 🙏 

Ping me in Discord for the Contributor role if you are in there - for sure earned it doing this stuff, IaC is such a headache sometimes.

digitalocean/terraform/main.tf
Swigler, line 1: hello, when trying to use this I am getting the following error

Error: Invalid function argument
│ 
│   on main.tf line 33, in locals:
│   33:   env_content = file("../../docker/.env")  
│     ├────────────────
│     │ while calling file(path)
│ 
│ Invalid value for "path" parameter: no file exists at "../../docker/.env"; this function works only with files that are distributed as part
│ of the configuration source code, so if this file will be created by a resource in this configuration you must instead obtain this result
│ from an attribute of that resource.
╵
 Am I missing something?

timothycarambat: You are not running the terraform template from inside the directory the template is in - as you can see it looks two directories up for the `docker` folder. Change that line if the .env is in the same directory

PR #153, opened 2023-07-20
[Fork] Batch embed by jwaltz 

timothycarambat:
Original Author: [jwaltz](https://github.com/jwaltz)

> I was having really long embed times for a couple large .docx documents ingested through the /hotdir. One had 256 chunks and took many minutes to fully embed, sometimes never finishing at all.
> 
> I converted the embedChunk() method to accept a list of chunks instead of just one chunk, and make a single call to OpenAI's embedding API endpoint, which seems to have sped up the embed process dramatically. The refactored embedChunks() required some changes to its usage in addDocumentToNamespace() as well. I included these changes for the 3 vector db options at /server/utils/vectorDbProviders/[vectorDb]/index.js
> 
> Let me know what you think and if this is helpful. Again, apologies for my editor's lsp changing some of the method signatures and line-length formatting, hope it is ok.

Had to modify this code as it would break the application since we do both multiple embeds and singular text embeds so each vector database needs to have this interface and support it since during chat mode we manually embed the query.


timothycarambat:
@jwaltz I dont know what your Discord handle is but happy to add you as a contributor! Apologies for the delay in merging this as its a massive quality improvement.

AntonioCiolino:
In excited to try this out!

jwaltz:
@timothycarambat I don't use discord much but am happy to have contributed nonetheless. Keep up the good work!

PR #155, opened 2023-07-20
LanceDB upgrade and improved functionality.

timothycarambat:
LanceDB 0.1.12 can now support count of rows and deletion of rows via the Node package.
- Bumped LanceDB version
- Added support for counting and deletion of documents
- Removed Delete warning on frontend when using lancedb

### Tests
- [x] Testing locally
- [x] Tested local Docker.

**note**: 0.1.15 does not work locally and is missing some kind of dependency and 0.1.12 and is highest non-broken distribution of the vectordb npm package. LanceDB is **not supported for windows or WSL systems by the core team**.


server/utils/vectorDbProviders/lance/index.js
```diff
@@ -42,8 +42,21 @@ const LanceDb = {
     await this.connect();
     return { heartbeat: Number(new Date()) };
   },
+  tables: async function () {
+    const fs = require("fs");
+    const { client } = await this.connect();
+    const dirs = fs.readdirSync(client.uri);
+    return dirs.map((folder) => folder.replace(".lance", ""));
+  },
```
timothycarambat, line 50: The native `tableNames()` function does not work so we roll our own.

PR #161, opened 2023-07-28
improve citations to show all text chunks referred and expand the citation to view full referenced text

timothycarambat:
Screen Shot 2023-07-27 at 10 30 58 PM


timothycarambat:
Also, fix some issues in light-mode. Why did nobody notice these things!! Ah well :)

PR #178, opened 2023-08-04
Actually enable x-scrolling when citations overflow.

efocht:
Only 3 and a half citations are displayed in the chat history. When there are more, there is no scroll bar to access them. Tiny fix to enable a horizontal scroll bar in this case. Obviously this was intended in the original code but the `no-scroll` was wrong.

### DIFF
--- a/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/Citation/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/Citation/index.jsx
@@ -22,7 +22,7 @@ export default function Citations({ sources = [] }) {
 
   return (
     <div className="flex flex-col mt-4 justify-left">
-      <div className="no-scroll flex flex-col justify-left overflow-x-scroll ">
+      <div className="flex flex-col justify-left overflow-x-scroll ">
         <div className="w-full flex overflow-x-scroll items-center gap-4 mt-1 doc__source">
           {combineLikeSources(sources).map((source) => (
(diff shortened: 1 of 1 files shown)


timothycarambat:
`.no-scroll` does not prevent `overflow-x-scroll` it just stops the scroll bar from showing. If you were to place your cursor over the items and scroll left right (like on a trackpad) it would work as intended. I think the issue may be just hiding the scroll bar on Windows since then the control is missing for them.

timothycarambat:
```
.no-scroll::-webkit-scrollbar {
  display: none !important;
}

/* Hide scrollbar for IE, Edge and Firefox */
#chat-history,
#chat-container,
.no-scroll {
  -ms-overflow-style: none !important;
  /* IE and Edge */
  scrollbar-width: none !important;
  /* Firefox */
}
```

efocht:
I'm sorry, my explanation is probably wrong. The impact is the same :-(

I am using Firefox under Linux and am not able to scroll to the right to access the citations that are hidden. I use a keyboard and a mouse, no trackpad. I recon this would work on a laptop with trackpad scrolling with 2 fingers. Or a tablet. But not on a desktop.

timothycarambat:
Out of curiosity - if you go to the area where it shows the 3 1/2 citations and hold `Shift` and then scroll - it does scroll, right?

Not saying that's the solution but want to be sure scrolling works at all for non-x-axis scrolling devices

timothycarambat:
Also, merged this! Ping me in [Discord](https://discord.gg/6UyHPeGZAC) if you are in there and get a contributor role!

PR #183, opened 2023-08-10
Feature/metadata

franzbischoff:
As discussed with @timothycarambat , this implements the PyMuPDF that is quick and handles Unicode better.
Also, allows us to use actual PDF metadata as 'title', 'author', etc, to populate the embedded metadata for further references when searching.

- One thing I find out is that the window that shows the references of the searches is not handling Unicode on 'title' field, but it is on the text body. This should be fixed.

- Known bug: lanceDB insists in assume the Scheme is different when differerent filetypes are uploaded for embedding. Pinecone handle this ok.

I've reviewed the documentation and the actual tables created by lanceDB and could not figure out.

The rest works fine.


franzbischoff:
A fix to my fix, please change this line:

`page.page_content.replace("\n ", " ").replace("  ", " ")`

by this:

`page.page_content.replace("\n ", " ").replace("\uFFFD", " ").replace("  ", " ")`

There are PDF's that inserts wierd stuffs as space and is recognized as "unknown".

.editorconfig
franzbischoff, line 1: This is a proposal change. Useful for dev, but can be dropped

collector/scripts/gitbook.py
franzbischoff, line 1: lint

collector/scripts/watch/convert/as_docx.py
franzbischoff, line 1: lint and added some fields, better check

collector/scripts/watch/convert/as_markdown.py
franzbischoff, line 1: added some fields, better check

collector/scripts/watch/convert/as_mbox.py
franzbischoff, line 1: added some fields, better check

collector/scripts/watch/convert/as_text.py
franzbischoff, line 1: Can't remember why this got duplicated! Please, check.
added some fields, better check

docker/Dockerfile
franzbischoff, line 1: This was reviewed and fixed according to hadolint rules/best practices

collector/scripts/watch/convert/as_pdf.py
franzbischoff, line 1: Here is the main thing.
some fields were added to the metadata, thus the other "convert" scripts were reviewed.

timothycarambat:
moved to #241

Issue #165, opened 2023-08-01
Appearance Settings Page


timothycarambat:
As the admin in a multi-user instance or in a single-user instance of AnythingLLM there should be a page for "Appearance" that allows the user to
- Update the logo in the sidebar of the app. Store locally
- Change the Default opening messages
  - each item should have a `text` key and an option `actions` array of buttons `(text + ur propsl)`

timothycarambat:
There should be a page on the URL `/system/appearance` that is accessible for all in single user mode. In multi user mode this page will only be available to the `admin` role users.

The file upload input should only allow image-type files and suggest the correct dimensions for the appropriate image size.
On upload, the image should be saved into a new folder `/server/storage/assets` and the value of the filename be saved as a key in the `System` table of the DB.

On page load we should try to load their logo and if it fails to load for any reason we should fallback to the new shiny AnythingLLM logo.

PR #186, opened 2023-08-11
Logo customization for single-user & multi-user modes

shatfield4:
connect #165


timothycarambat (changes requested):
See my commit before this comment

frontend/src/components/Modals/Settings/index.jsx
```diff
@@ -132,6 +134,12 @@ function SettingTabs({ selectedTab, changeTab, settings, user }) {
           />
         </>
       )}
+      <SettingTab
+        displayName="Appearance"
+        tabName="appearance"
+        icon={<Eye className="h-4 w-4 flex-shrink-0" />}
+        onClick={() => window.open(paths.appearance())}
+      />
```
timothycarambat, line 142: This tab should only appear when *not* in multi use mode. Not show in all modes.

timothycarambat (changes requested):

PR #193, opened 2023-08-15
Custom default messages implementation for single and multi-user modes

shatfield4:
connect #165

New table created to store custom welcome messages and renders custom messages if they are set in appearance settings. If they are not set in appearance settings, it displays the default AnythingLLM welcome messages.


frontend/src/models/admin.js
```diff
@@ -216,6 +216,26 @@ const Admin = {
         return { success: false, error: e.message };
       });
   },
+  setWelcomeMessages: async function (messages) {
+    return fetch(`${API_BASE}/system/set-welcome-messages`, {
+      method: "POST",
+      headers: {
+        ...baseHeaders(),
+        "Content-Type": "application/json",
```
timothycarambat, line 224: Do we need to specify the `application/json` here? All other calls are working without this?

frontend/src/models/admin.js
```diff
@@ -216,6 +216,26 @@ const Admin = {
...
+      },
+      body: JSON.stringify({ messages }),
+    })
+      .then((res) => {
+        if (!res.ok) {
+          throw new Error(res.statusText || "Error setting welcome messages.");
+        }
+        return { success: true, ...res.json() };
```
timothycarambat, line 232: if success is a tag that should exist, return it from the response then. Can then just simplify this to
```
.then((res) => {
  if(!res.ok) throw....
  return res.json()
}
```

frontend/src/models/system.js
```diff
@@ -187,6 +187,40 @@ const System = {
...
+        return null;
+      });
+  },
+  setWelcomeMessages: async function (messages) {
+    return fetch(`${API_BASE}/system/set-welcome-messages`, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+      },
+      body: JSON.stringify({ messages }),
```
timothycarambat, line 211: setting welcome message, even in single user mode should add base headers because without that - even on a password protected instance anyone could update the messages if they knew the endpoint

server/endpoints/system.js
```diff
@@ -477,6 +478,45 @@ function systemEndpoints(app) {
...
+          response.locals.multiUserMode &&
+          response.locals.user?.role !== "admin"
+        ) {
+          return response.sendStatus(401).end();
+        }
+        if (!request.body || !Array.isArray(request.body.messages)) {
+          return response.status(400).json({
+            message: "Invalid message format. Expected an array of messages.",
+          });
+        }
```
timothycarambat, line 507: Oh this is why you are specifying the content-type in the POST.

Just use `reqBody` function  - can then restructure the intended result automatically.
`const { messages = [] } = reqBody(request);`

server/models/welcomeMessages.js
```diff
@@ -0,0 +1,89 @@
+const WelcomeMessages = {
+  tablename: "welcome_messages",
```
timothycarambat, line 2: Good to use plural for tables

frontend/src/pages/System/Appearance.jsx
```diff
@@ -68,66 +81,181 @@ export default function Appearance() {
...
+    if (!success) {
+      setErrorMsg(error);
+      return;
+    }
+    setSuccessMsg("Successfully updated welcome messages.");
+    setHasChanges(false);
+  };
+
+  const handleBackNavigation = () => {
+    window.location.href = "/";
```
timothycarambat, line 121: `window.location = paths.home()`

timothycarambat (approved):

PR #246, opened 2023-09-21
Renamed all indicies to vectors to avoid confusion of vocab

### LINKED ISSUE #245: Replace all references to 'indicies'  to 'vectors'
We should rename any references to 'indicies' to 'vectors' to avoid confusion.

shatfield4:
resolves #245

Renaming all mention of the word 'indicies' to 'vectors' to avoid naming confusion.


timothycarambat (changes requested):
I actually didn't mean for you to change `vector` in the backend code. I intended to just change this piece of text only 😆 in the sidebar.

The backend code doesn't need to change, but we can keep these changes anyway since it's not a big lift & does help with clarity!

Screen Shot 2023-09-21 at 1 40 48 AM

shatfield4:
> I actually didn't mean for you to change `vector` in the backend code. I intended to just change this piece of text only 😆 in the sidebar.
> 
> The backend code doesn't need to change, but we can keep these changes anyway since it's not a big lift & does help with clarity!
> 
> Screen Shot 2023-09-21 at 1 40 48 AM

fixed

timothycarambat (approved):

PR #247, opened 2023-09-22
changing the build behavior in the aws cloudformation template from […

santini-cyber:
…scripts-user, always] to [scripts-user, once-per-instance] so the userdata script is not run every time the server boots.


timothycarambat:
If you are in the Mintplex Labs Discord you can get a contributor role if you want one! Just ping me in the discord via DM or in a channel

PR #256, opened 2023-09-29
Prevent updates of specific keys via API

### LINKED ISSUE #255: API ENV block updates on permissions
TLDR: 
Prevent non-admins in multi-user mode from pushing updates via the API
Dont allow password updates when in multi-user mode.

https://huntr.dev/bounties/00ec6847-125b-43e9-9658-d3cace1751d6/

timothycarambat:
prevent non-admin from updating settings
prevent password updates in multi-user mdoe

resolves #255


timothycarambat:
@raltheo

PR #263, opened 2023-10-01
Feature/devcontainer

franzbischoff:
This PR does the following:

- Added configuration for creating a devcontainer with the environment ready for starting server:dev and frontend:dev
- Changes the location of eslint config in order to be global. In fact, Eslint is mostly useful when employed with the IDE extension. The package lint script uses prettier, not eslint.
- Creates a prettier config file and ignore file, so both CLI and IDE extensions have the same formatting rules.
- Tailwind is now integrated into the IDE.

Adjustments:
- settings.json needs the eslint config, so VSCode doesn't complain about the old format.
- package-lock.json is removed in favor of yarn.lock. The project uses yarn, not npm.
- Some adjustments were made on all "package.json" files.
- Added SERVER_ADDRESS in server/.env.example to fix a bug when using WSL.

Others:
- Most (99%) of changes in files are lint/format, not code changes.
- The markdown files are also formatted by the IDE, and CLI if done manually. Some markdown files had to be reviewed because the original code (that wasn't following the standard) was misformatted by prettier. From now on, these files can be re-formatted normally.
- Dockerfile wasn't changed, just linted/reformatted
- The "collector" folder and the Python code are still the same. In the next PR I'll think about it.

Final words:

These standardizations of code format don't require to use of the devcontainer. One can still run the server and frontend and configure the extensions by hand in the system environment. Nevertheless, the devcontainer is a safer and quicker way to start developing.

I've tested in several ways and it is working. But who knows... 🤞


franzbischoff:
Works on gh workspaces too!
image

frontend/package-lock.json
franzbischoff, line 1: Do we really need to keep npm and yarn locks at the same time?

collector/scripts/watch/convert/as_mbox.py
franzbischoff, line 1: This has been improved to deal with lots of e-mail formats, etc. Even though, some encoded strings may need to be caught in later versions.

franzbischoff:
For a better world

PR #275, opened 2023-10-11
v2 document picker

shatfield4:
Changes Made
• Revamped document picker UI


frontend/src/components/Modals/MangeWorkspace/Documents copy/DocumentRow/index.jsx
timothycarambat, line 1: Are these files even needed anymore?

shatfield4: Oh, these are removed in the latest commit

PR #295, opened 2023-10-26
Added telemetry for onboarding completion

shatfield4:


server/endpoints/workspaces.js
```diff
@@ -21,13 +21,22 @@ function workspaceEndpoints(app) {
   app.post("/workspace/new", [validatedRequest], async (request, response) => {
     try {
       const user = await userFromSession(request, response);
-      const { name = null } = reqBody(request);
+      const { name = null, onboardingComplete } = reqBody(request);
```
timothycarambat, line 24: set default to false if the key doesnt exist in request

server/endpoints/workspaces.js
```diff
@@ -21,13 +21,22 @@ function workspaceEndpoints(app) {
...
       const { workspace, message } = await Workspace.new(name, user?.id);
       await Telemetry.sendTelemetry("workspace_created", {
         multiUserMode: multiUserMode(response),
         LLMSelection: process.env.LLM_PROVIDER || "openai",
         VectorDbSelection: process.env.VECTOR_DB || "pinecone",
       });
+
+      if(onboardingComplete) {
```
timothycarambat, line 32: lets do an explict type +value match for `true`

server/endpoints/workspaces.js
```diff
@@ -21,13 +21,22 @@ function workspaceEndpoints(app) {
...
+        await Telemetry.sendTelemetry("onboarding_complete", {
+          multiUserMode: multiUserMode(response),
+          LLMSelection: process.env.LLM_PROVIDER || "openai",
+          VectorDbSelection: process.env.VECTOR_DB || "pinecone",
+        });
```
timothycarambat, line 37: You can drop the options - all we need is the event here I think.

timothycarambat (changes requested):

PR #297, opened 2023-10-26
devcontainer v1

franzbischoff:
This PR does the following:

- Added configuration for creating a devcontainer with the environment ready for starting server:dev and frontend:dev
- Changes the location of eslint config in order to be global. In fact, Eslint is mostly useful when employed with the IDE extension. The package lint script uses prettier, not eslint.
- Creates a prettier config file and ignore file, so both CLI and IDE extensions have the same formatting rules.
- Tailwind is now integrated into the IDE.

Adjustments:
- settings.json needs the eslint config, so VSCode doesn't complain about the old format.
- package-lock.json is removed in favor of yarn.lock. The project uses yarn, not npm.
- Some adjustments were made on all "package.json" files.
- Added SERVER_ADDRESS in server/.env.example to fix a bug when using WSL.

Final words:

These standardizations of code format don't require to use of the devcontainer. One can still run the server and frontend and configure the extensions by hand in the system environment. Nevertheless, the devcontainer is a safer and quicker way to start developing.

I've tested in several ways and it is working. But who knows... 🤞


franzbischoff (commented):
reviewd

.devcontainer/README.md
```diff
@@ -0,0 +1,85 @@
...
+If you plan to develop the collector, currently (maybe not in a few months), you'll need to install the `pre-release` versions of the extensions to work as they should since MS is moving features around their extensions.
+
+Make sure you have installed the following extensions as `pre-release`:
+
+- Python (ms-python.python)
+- Pylance (ms-python.vscode-pylance)
+- Pylint (ms-python.pylint)
+- Black formatter (ms-python.black-formatter)
+- Python Debugger (ms-python.debugpy)
+
```
timothycarambat, line 67: Do we still need to support this now that we removed all python code from the codebase?

timothycarambat (changes requested):
Looks like this will work great, i think the docs need a touch up because obviously we made a ton of changes while this was underway - namely that the collector was changed from Python to NodeJS, which should be easier now to run alongside everything else hopefully.

franzbischoff (commented):
done

franzbischoff (commented):
reviewed

timothycarambat (approved):

PR #308, opened 2023-10-29
mbox parsing improvements v1

franzbischoff:
better parse for e-mails.
deals with several encodings and multipart issues.
It can still miss some weird encoding that may be used on fields other than "Subject" and the e-mail content, but it seems a case-to-case to deal with in further versions.


franzbischoff:
Wow, a PR autobot :-D

1- Nice!
2- Very nice :-D
3- Agree! But I just repeat what was done :-D
4- new, old format?

Issue #313, opened 2023-10-31
LM Studio Integration?


abcnow:
I wonder if you can install anything llm with LM Studio and use your LLM instead of ChatGPT? Thanks in advance!

timothycarambat:
Related #118 

Thanks for the suggestion! Will be doing a bunch of localLLM support soon!

franzbischoff:
PR #335

timothycarambat:
Resolved by f499f1ba59f2e9f8be5e44c89a951e859382e005

PR #335, opened 2023-11-06
Using OpenAI API locally

franzbischoff:
This PR implements the basics for using the OpenAI API for connecting to a locally compatible server like gpt4all or lmstudio.ai while keeping separate configurations for the embedding model.

the apps mentioned above are just those I used by chance. Both work very well, but lmstudio.ai allows you to use your GPU along with the CPU. It has a nice browser look that allows you to discover models published on Huggingface.

Refer to issue #313


netandreus:
@franzbischoff I'm really looking forward to this release. How can I use LocaAI with this branch?
Where should I specify LocalAI url?

netandreus:
Anything_local_ai

Anything_local_ai_embedding

@franzbischoff Can you please add support LocalAI for embeddings also? It would be very nice.

netandreus:
Hello, @franzbischoff ! Thank you for quick reply. I try to use master branch at actual top commit (https://github.com/Mintplex-Labs/anything-llm/commit/6e72519cafbb603b7559e6c47e36836454e69576) but still don't see ability to use LocalAI for Embeddings. Only OpenAI and Azuer OpenAI are available. Can you please add it?

Screenshot 2023-11-10 at 10 35 49

PR #351, opened 2023-11-10
Create manager role and limit default role

shatfield4:
Created a "Manager" role that has all access that the "Admin" role has except for changing 
- LLM Preference 
- Vector DB
- Embedding Preference

"Default" role
- Now limited to chatting only
- Users cannot modify or upload any docs and cannot change any settings


shatfield4:
Still WIP, fixing single user mode

Issue #352, opened 2023-11-10
/v1/workspace/{slug}/update-embeddings not working


LooproInc:
Hello! I’ve been able to successfully use all other API endpoints except for the embedding API. After a successful file upload to the workspace (visible on the frontend), the embedding continually returns {‘workspace’: None}. Could you please assist in resolving this issue?

tlandenberger:
Hi @LooproInc I've looked into the code and was able to reproduce the error. Looks like the return type of _addDocuments_ function in _documents.js_ is wrong when no documents were added.
So changing line 39 in _document.js_ from
`if (additions.length === 0) return;`
to 
`if (additions.length === 0) return {failed: [], embedded: []};`
fixed the issue for me. Could you check if this solves your problem and then we can create a pull request?

timothycarambat:
closed by https://github.com/Mintplex-Labs/anything-llm/pull/353

PR #353, opened 2023-11-10
fix: adjust return type of addDocuments in case of no additions

tlandenberger:
Fixing a bug when the _/v1/workspace/{slug}/update-embeddings_ endpoint was called with no additions to the documents.
Solves issue #352


timothycarambat:
Ah great catch! Thanks for the contribution! If you want a contributor badge in the discord ping me!

PR #361, opened 2023-11-13
LocalAI for embeddings

tlandenberger:
Provide option to use LocalAI as local embedding engine. LocalAI must be running with installed embedding model. The selected model for embedding needs to be set in the _.env_ file.


Amejonah1200:
👀👀👀👀👀👀👀

PR #400, opened 2023-11-29
Add down arrow to scroll user to bottom of page

shatfield4:
Create down arrow that appears when chat is not scrolled to the bottom


timothycarambat:
@shatfield4 Can you fix merge conflicts and split the arrow show function from the PFP stuff? Too much going on to parse for just PFP support and the arrow and I know I'm going to miss something here

PR #413, opened 2023-12-06
[Feature] AnythingLLM use locally hosted Llama.cpp and GGUF files for inferencing

timothycarambat:
- Add support for the 2,000+ [GGUF models](https://huggingface.co/models?library=gguf&sort=trending) that can be used as the LLM of choice. Making AnythingLLM now 100% privately running.
- Support sync chat and streaming from models.
- Add documentation to support this feature
- Hide selection when on hosted domain


Amejonah1200:
Is GPU Acceleration possible for this feature?

timothycarambat:
> Is GPU Acceleration possible for this feature?

[node-llama-cpp using CUDA is supported](https://withcatai.github.io/node-llama-cpp/guide/CUDA#using-node-llama-cpp-with-cuda) 
[Apple Metal is also supported](https://withcatai.github.io/node-llama-cpp/guide/Metal) 

And once built it will use GPU acceleration for the llama client, however for docker instances this is CPU only so some additional work will need to be done to bind GPUs to the instance so they can be used.

PR #422, opened 2023-12-11
feat: enable HTML uploads from UI

### LINKED ISSUE #418: Document Upload file type HTML
Extend the pulling from website, to support uploading of HTML files.
I have had better luck with converting PDF to HTML rather than using the PDF upload, because my method uses OCR to preserve formatting, which does not happen with the PDF uploads.

timothycarambat:
resolves #418


timothycarambat:
Screen Shot 2023-12-11 at 2 38 58 PM

PR #448, opened 2023-12-14
fixes/#101-repo-doucument loader

pramodgoud7:
Fixed the loader issue

### DIFF
--- a/collector/scripts/link.py
+++ b/collector/scripts/link.py
@@ -6,6 +6,7 @@ from .link_utils import append_meta, AsyncHTMLSessionFixed
 from .utils import tokenize, ada_v2_cost
 import requests
 from bs4 import BeautifulSoup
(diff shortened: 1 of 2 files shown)


timothycarambat (changes requested):
Hey there, thanks for contributing! I just wanted to touch on a few issues with this PR

1. The branch is named `bug/` but the PR title is `fixes/`. This is neither, as this is a `feat/`!

2. In `collector/scripts/link.py` you import a library in EJS format, but this is a python file. Also this does not install the dependency or even have it made available for use by the client side. 

3. `frontend/.env.example` is edited and is unrelated to the PR, it should be checked out.

If this PR is a WIP then just let me know, but in its current state this does not add the functionality you wished to add to AnythingLLM.

timothycarambat:
Closing as the migration from python loader to nodeJS loader breaks this PR totally.

PR #460, opened 2023-12-16
truncate file row using truncate package

kschmelter13:
Not entirely sure if this is correct (havent made a local build) but seems like it since custom truncate hasnt been used in a little


timothycarambat:
Ah, so this would not be correct (and its unclear so this is on us), but that truncate function from utils is actually more like "middle truncation".

So verylongstringthatoverflows => verylo...ows
as opposed to native truncate npm library which is tail truncation only.

Closing this PR but will make another to clarify that function and why it exists 👍

Issue #523, opened 2024-01-04
Add Support for HTTPS to Server


pritchey:
For some self-hosted environments, everything much utilize encryption even if it rides the loopback interface.  Therefore, the server which executes must support both running in HTTP and HTTPS mode.

This pull request modifies server/index.js to provide users the ability to enable HTTPS by making a change to the .env (configuration) file.  The configuration file mod adds a flag to indicate the desire to enable HTTPS and the path/filename for the key and cert to be used.

Logging output has been updated to include whether or not the server is being launched as an HTTP or HTTPS server, as well as when the code is about to load the key/cert file.

PR #524, opened 2024-01-04
523-Added support for HTTPS to Server.

pritchey:
Details provided in the issue ticket (#523).  This pull request adds the ability to enable HTTPS capability in the server for those that require it for self-hosted deployments.


timothycarambat:
resolves #523

Issue #204, opened 2023-08-21
addDocumentToNamespace Cannot destructure property 'id' of 'chunk.payload' as it is undefined.


franzbischoff:
Following the series #197 ...

On Qdrant, when trying to use cached embeddings...

```
Cached vectorized results of my-awesome-pdf-1691674380/my-awesome-pdf-pg9-95185230-2b4f-47ba-9639-ff8b0b8539ea.json found! Using cached data to save on embed costs.
addDocumentToNamespace Cannot destructure property 'id' of 'chunk.payload' as it is undefined.
```

Result: namespace is created, but vectors are not.

PR #526, opened 2024-01-04
Issue #204 Added a check to ensure that 'chunk.payload' exists and contains the 'id' property

isayandev:
I added a check (if (chunk.payload && chunk.payload.id)) to ensure that 'chunk.payload' exists and contains the 'id' property before attempting to destructure it.
If the check fails, an error message is logged, and you can handle the situation accordingly.


timothycarambat:
👋 Thank you @isayandev for this contribution. If not already, you can ping me in Discord and I can give you a contributor badge!

resolves #204

Issue #540, opened 2024-01-05
Possible to add "STOP" button if user does not like the AI's response? 


jameschen83:
In ChatGPT, user can click "stop generation" button to stop LLM streaming. Is it available in AnythingLLM, or Possible to add this feature in AnythingLLM? 

If no, I plan to fork the branch to add the feature.

duc-gp:
When I would pick this up to implement, how would I start and how is the process regarding UI / UX? Are there any design specs for this or can the developer propose something in their Pull request?

timothycarambat:
We do not have a design for this yet. The main limitation here is that all this would do is disconnect the client from the response stream - it would not terminate the request at the LLM side - so an infinite response loop would still continue on the LLM side and it would stay occupied until it finished.

If the client just wants to "stop" seeing the output then we can break the response stream but I think this will lead to more issues with localLLMs that go wild and keep streaming and the user sends a new prompt and now the chat is hanging because the LocalLLM is still stream the old response.

duc-gp:
> We do not have a design for this yet. The main limitation here is that all this would do is disconnect the client from the response stream - it would not terminate the request at the LLM side - so an infinite response loop would still continue on the LLM side and it would stay occupied until it finished.
> 
> If the client just wants to "stop" seeing the output then we can break the response stream but I think this will lead to more issues with localLLMs that go wild and keep streaming and the user sends a new prompt and now the chat is hanging because the LocalLLM is still stream the old response.

I dig a little bit into the code, so I think a way to disconnect the client would be to declare a abortcontroller instance outside of https://github.com/Mintplex-Labs/anything-llm/blob/f4b09a8c794f4e1ae0a21a2f47790b680579e087/frontend/src/models/workspace.js#L76 and passing it, so it can be called from outside where its used? Would this be the correct way to implement this?

But then I end up having still a blinking cursor at the end, see screenshot:
![Screenshot 2024-02-17 at 23 06 09](https://github.com/Mintplex-Labs/anything-llm/assets/40763918/699805eb-9ff1-4d01-ad79-9a493f97a56d)

Then I thought about passing a chatResult with `type: "finalizeResponseStream"` but for that I would need the `chatId` ?

Can you provide some guidance on this topic, maybe I'm on a complete wrong way?

Thanks

duc-gp:
> has it been added? i cant see it.

[...]

PR #550, opened 2024-01-06
Issue #540 add stop responding

Satyabrata31:
#540

### DIFF
--- a/server/endpoints/chat.js
+++ b/server/endpoints/chat.js
@@ -1,3 +1,4 @@
+const expresss = require("express");
 const { v4: uuidv4 } = require("uuid");
 const { reqBody, userFromSession, multiUserMode } = require("../utils/http");
 const { Workspace } = require("../models/workspace");
@@ -11,6 +12,8 @@ const {
   writeResponseChunk,
 } = require("../utils/chats/stream");
 
+let shouldRespond = true;
+
 function chatEndpoints(app) {
   if (!app) return;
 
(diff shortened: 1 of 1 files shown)


server/endpoints/chat.js
```diff
@@ -11,13 +12,23 @@ const {
   writeResponseChunk,
 } = require("../utils/chats/stream");
 
+let shouldRespond = true;
+
```
timothycarambat, line 16: This variable would be globally controlled in the API, which would break in any multi-user instance since if two chats were running for two different user connections and user one canceled the response, user two's steam would stop at the same time.

server/endpoints/chat.js
```diff
@@ -1,3 +1,4 @@
+const expresss = require("express");
```
timothycarambat, line 1: Why was this re-imported?

`chatEndpoints(app) {` `app` is the passed-in express server reference to build these endpoints, but I don't see where other references to express are required in this change

server/endpoints/chat.js
```diff
@@ -11,13 +12,23 @@ const {
...
   app.post(
     "/workspace/:slug/stream-chat",
     [validatedRequest],
     async (request, response) => {
+      if (!shouldRespond){
+        response.status(200).json({
+          id: uuidv4(),
+          type: "success",
+          message: "Stop responding logic executed successfully."
+        });
```
timothycarambat, line 29: The ouput of the response object is very important for the client UI to respond
```javascript
response.status(500).json({
                  id: uuidv4(),
                  type: "abort", // success is not a valid type to be handled.
                  textResponse: null,
                  sources: [],
                  close: true,
                  error: `Stop responding logic executed successfully`,
                });
```

timothycarambat: See https://github.com/Mintplex-Labs/anything-llm/blob/master/frontend/src/utils/chat/index.js

timothycarambat (changes requested):
With the current changes:
- There is a globally controlled variable to stop streaming that would break multi-user streaming
- The client response for streaming to halt returns an unhandled object
- The stop request does not fully cancel the model provider request, so this would simply drop all future inferencing packets, which the end effect would be the same for the client, but the LLM provider would still have produced the tokens to fulfill the entire request and will be occupied if a new request is sent while busy.

PR #618, opened 2024-01-18
A startup script for Anything LLM

lolsZz:
**Main Script**:
- The script starts by checking for required dependencies.
- It then navigates to the project directory.
- It sets the API base URL for the project (reverse proxy for gemini-pro).
- It counts the number of services that are already running.
- Based on the number of running services, it presents options to the user:
- If all services are running, it asks the user if they want to restart the application or close all services.
- If some services are running, it restarts all services.
- If no services are running, it starts all services.
- The script then starts the server, collector, and front-end services using the `start_service()` function.
- It logs a message indicating that all services have been started and provides the location of the log files for output.

 ### Pull Request Type

- [ x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #xxx

### What is in this change?

### Additional Information

### Developer Validations

### DIFF
--- a/startup.sh
+++ b/startup.sh
@@ -1,0 +1,130 @@
+#!/bin/bash
+
+# Define the project directory
+PROJECT_DIR="$HOME/anything-llm"
+
+# Load configuration from an external file (if exists)
+CONFIG_FILE="$PROJECT_DIR/startup.config"
+if [ -f "$CONFIG_FILE" ]; then
+    source "$CONFIG_FILE"
+else
+    # Define the ports and associated services if no config file is found
+    declare -A ports=( ["server"]=3001 ["collector"]=8888 ["frontend"]=3000 )
+fi
+
+# Log function with timestamp
+log() {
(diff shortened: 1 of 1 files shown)


timothycarambat:
While this is an impressive script I think its particular use may be unique or niche to how you ideally want to run AnythingLLM for dev. Considering the impact of this would be only for running AnythingLLM in development mode, the recommended mode of just running each command in a terminal tab and having stdout/stderr being dumped to the terminal is sufficient for development. 

Am I reading this PR correctly or is there an overt advantage to this kind of startup script? Additionally, the frontend proxies requests to an untrusted vercel application that is running a proxy around Google's API restrictions for Gemini use - which would be a ToS violation so we cannot commit that to the core repo.

Lastly, I do not think the proxy would work because the requests to gemini are sent from the backend, not the frontend so they would not even be proxied.

Closing this issue due to security concerns and will re-open if I am incorrect in my understanding of what this scripts purpose is.

PR #655, opened 2024-01-29
Patch lanceDB not deleting vectors from workspace

### LINKED ISSUE #594: [BUG]: Embeddings not removed.

Even though I may have added and saved one embedding in a workspace and later realised it was not giving me correct results, I removed it. But in the workspace it was still given as citation - and yes it was giving answers from that only.

The embeddings are not getting removed when I choose to remove from a workspace, my only solution was to remove the work space completely and make a new one.

timothycarambat:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #594

### What is in this change?

- documentVectors self-sanitize on delete of parent document
- patch lanceDB not deleting vectors from workspace

RCA:
When adding a new document to a workspace for lanceDB that also created the table, the `submissions` item would have a different `id` given to it by the `collector` that would overwrite the known document id that was stored in the database - making deletion of it impossible.

This only impacted vectors that were totally new to the system (uncached) as cached documents would be able to be properly tracked. 

If users have this issue, the easiest full solve is just to delete the workspace and it will fully remove the vectors. They will persist in the DB, but be orphaned which is a negligible impact.

### Developer Validations

### DIFF
--- a/server/utils/vectorDbProviders/lance/index.js
+++ b/server/utils/vectorDbProviders/lance/index.js
@@ -207,9 +207,9 @@ const LanceDb = {
 
           vectors.push(vectorRecord);
           submissions.push({
+            ...vectorRecord.metadata,
             id: vectorRecord.id,
             vector: vectorRecord.values,
-            ...vectorRecord.metadata,
           });
           documentVectors.push({ docId, vectorId: vectorRecord.id });
         }
--- a/server/models/documents.js
+++ b/server/models/documents.js
@@ -106,6 +106,9 @@ const Document = {
         await prisma.workspace_documents.delete({
           where: { id: document.id, workspaceId: workspace.id },
         });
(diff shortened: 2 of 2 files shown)


server/utils/vectorDbProviders/lance/index.js
```diff
@@ -207,9 +207,9 @@ const LanceDb = {
 
           vectors.push(vectorRecord);
           submissions.push({
+            ...vectorRecord.metadata,
```
timothycarambat, line 210: spread first so that our values overwrite whatever might be colliding in metadata - namely `vector` and `id` key. This way what we define as a property will carry throughout the creation and upsert of the vectors.

PR #665, opened 2024-02-01
Fixed docker-compose report error: Additional property name is not al…

phuang07:
…lowed

 ### Pull Request Type

- [ ] ✨ feat
- [x ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #xxx

### What is in this change?

Remove invalid property in the docker-compose.yaml 

### Additional Information
When run `docker-compose up ` there's an error: Additional property name is not allowed.
This patch fix this.

### Developer Validations

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -1,7 +1,5 @@
 version: "3.9"
 
-name: anythingllm
-
 networks:
   anything-llm:
     driver: bridge


timothycarambat:
Are you using docker-compose 3.9 or higher? `name` is a relatively new key from what I understand

phuang07:
> Are you using docker-compose 3.9 or higher? `name` is a relatively new key from what I understand

I am on mac, runing docker desktop with docker Compose version v2.0.0-rc.3.

timothycarambat:
Due to the properties that can be required by the current DockerFile (like cap-add) and other properties, I recommend just upgrading your docker engine so that it is compatible. Without this specification we are opening our repo to docker engine compatibility issues for those running much older docker engines that may not support capabilities we require

You can see the compatibility for[ docker version <> compose spec here](https://docs.docker.com/compose/compose-file/compose-file-v3/#compose-and-docker-compatibility-matrix)

PR #681, opened 2024-02-06
Fix white screen bug when deleting chats under Workspace Chats menu

### LINKED ISSUE #669: [BUG]: White screen after deleting a workspace chat from settings

Local development

### What happened?

When you are at the page "Workspace Chats", and you have more than one
page of chats, and you delete a chat, and then use the button "next
page" or "previous page", the app crashes with a white screen.
Restarting the app shows that the chat(s) was/were deleted.

This only happens when there are more than 20 chats sent and we see the next and previous pages for pagination.

shatfield4:
… on delete/fix backend invalid json error

 ### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #669 

### What is in this change?

- Fix a bug where if we have more than 20 workspace chats and we delete a chat then click previous or next buttons, a white screen would show until we refresh the page
- Changed from using a ref to delete the row to just filtering out the chats in the parent component

### Additional Information

### Developer Validations

### DIFF
--- a/frontend/src/pages/GeneralSettings/Chats/index.jsx
+++ b/frontend/src/pages/GeneralSettings/Chats/index.jsx
@@ -142,6 +142,10 @@ function ChatsContainer() {
     setOffset(offset + 1);
   };
 
+  const handleDeleteChat = (chatId) => {
+    setChats((prevChats) => prevChats.filter((chat) => chat.id !== chatId));
+  };
+
   useEffect(() => {
     async function fetchChats() {
       const { chats: _chats, hasPages = false } = await System.chats(offset);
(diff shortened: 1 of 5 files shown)


server/endpoints/system.js
```diff
@@ -774,7 +774,7 @@ function systemEndpoints(app) {
       try {
         const { id } = request.params;
         await WorkspaceChats.delete({ id: Number(id) });
-        response.sendStatus(200).end();
+        response.json({ success: true });
```
timothycarambat, line 777: Why change this if we dont use it on the frontend?

frontend/src/pages/GeneralSettings/Chats/index.jsx
```diff
@@ -142,6 +142,10 @@ function ChatsContainer() {
     setOffset(offset + 1);
   };
 
+  const handleDeleteChat = (chatId) => {
+    setChats((prevChats) => prevChats.filter((chat) => chat.id !== chatId));
+  };
```
timothycarambat, line 147: 👍 makes sense, the max this list can even be is 20 anyway

PR #683, opened 2024-02-06
[CHORE] Hide security tab on Settings Sidebar/Refactor SettingsSidebar component

### LINKED ISSUE #682: [FEAT]: Hide Security Tab on Setting Sidbar

Once a system is set to multi-user mode there is no way to set the system to anything else since multi-user mode is a one-way operation.

Once in multi-user, we should just hide the security option on the sidebar to clear up space.

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #682 

### What is in this change?

- When the instance is in multiusermode, we hide the security settings tab because there is no point in showing this screen since you cannot modify anything after multiusermode is enabled.
- Refactor SettingsSidebar component so that we do not have a separate mobile sidebar anymore and render the same options on both.

### Additional Information

### Developer Validations

### DIFF
--- a/frontend/src/components/SettingsSidebar/index.jsx
+++ b/frontend/src/components/SettingsSidebar/index.jsx
@@ -29,8 +29,8 @@ import { isMobile } from "react-device-detect";
 
 export default function SettingsSidebar() {
   const { logo } = useLogo();
-  const sidebarRef = useRef(null);
   const { user } = useUser();
+  const sidebarRef = useRef(null);
   const [showSidebar, setShowSidebar] = useState(false);
   const [showBgOverlay, setShowBgOverlay] = useState(false);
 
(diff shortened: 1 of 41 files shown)


frontend/src/components/SettingsSidebar/index.jsx
```diff
@@ -24,11 +24,21 @@ import {
...
   const { user } = useUser();
+  const [multiUserModeEnabled, setMultiUserModeEnabled] = useState(false);
+
+  useEffect(() => {
+    async function fetchIsMultiUserMode() {
+      const multiUserModeEnabled = await System.isMultiUserMode();
+      setMultiUserModeEnabled(multiUserModeEnabled);
+    }
+    fetchIsMultiUserMode();
+  }, []);
```
timothycarambat, line 41: You dont need to check this. If the user has a role property it is multi user

frontend/src/components/SettingsSidebar/index.jsx
```diff
@@ -169,14 +179,16 @@ export default function SettingsSidebar() {
...
+                {!multiUserModeEnabled && (
+                  <Option
+                    href={paths.settings.security()}
+                    btnText="Security"
+                    icon={<Lock className="h-5 w-5 flex-shrink-0" />}
+                    user={user}
+                    flex={true}
+                    allowedRole={["admin", "manager"]}
+                  />
+                )}
```
timothycarambat, line 191: Lets have a new property on the Option that allows us to set it as hidden with  a boolean property

timothycarambat (changes requested):
Forgot to update the mobile sidebar as well. This seems to occur each time we update the setting sidebar. Lets refactor this component to where the Mobile and Desktop layout are rendering the same options, but they keep their specific container styles.

timothycarambat (approved):

PR #685, opened 2024-02-07
Refactor handleStream to LLM Classes

### LINKED ISSUE #684: [CHORE]: Refactor `handleStreamResponse`

The `handleStreamResponse` function is a disaster right now. We should handle stream responses from the LLM provider if the function exists (meaning that it is a special case).

If the function does not exist in the class then we fall back to the OpenAI catchall functionality. This will simplify the context and which how streamed responses are handled.

timothycarambat:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #684 

### What is in this change?

Refactors all instances of `handleStreamResponses` to a single class method on the LLMProvider itself.

### Additional Information

### Developer Validations

- [x] Tested all providers


timothycarambat:
ready to merge

PR #700, opened 2024-02-09
[FEAT] Improved CSV chat exports

### LINKED ISSUE #697: [FEAT]: Improved CSV Chat Exports

Currently, the CSV export of chats is not very helpful to an admin. The workspace chat export as a csv will likely be used for auditing, review, RLHF, or just general filtering and usage analysis.

The csv currently only exports the message and response, which are important but we should be as verbose as possible when doing the CSV export.

We should have the following columns in the table at a **minimum**:
- `id` 
- `username`
- `workspace_name`
- `prompt`
- `response`
- `sent_at`

This format will be more helpful when we implement metrics on chats like token use, chat_model, and more. So this implementation should be easily extensible to add new metrics without much lift.

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #697 

### What is in this change?

- Improve CSV chat exports to include more data
- CSV exports now include the following columns: 

  - ```id```
  - ```username```
  - ```workspace_name```
  - ```prompt```
  - ```response```
  - ```sent_at```

### Additional Information

### Developer Validations


server/utils/helpers/chat/convertTo.js
```diff
@@ -5,16 +5,18 @@ const { Workspace } = require("../../../models/workspace");
...
+async function convertToCSV(preparedData) {
+  const rows = ["id,username,workspace,prompt,response,sent_at"];
+  for (const item of preparedData) {
+    // Escape double quotes and wrap content in double quotes
+    const escapedPrompt = `"${item.prompt
+      .replace(/"/g, '""')
+      .replace(/\n/g, " ")}"`;
+    const escapedResponse = `"${item.response
+      .replace(/"/g, '""')
+      .replace(/\n/g, " ")}"`;
```
timothycarambat, line 17: Break this into a non-exported function in this file and wrap the relevant data in it for better readability.

server/utils/helpers/chat/convertTo.js
```diff
@@ -33,10 +35,27 @@ async function convertToJSONL(workspaceChatsMap) {
     .join("\n");
 }
 
-async function prepareWorkspaceChatsForExport() {
+async function prepareWorkspaceChatsForExport(type = "jsonl") {
```
timothycarambat, line 38: Do a validation check here to ensure that the type is in the `exportMap` key set.

server/utils/helpers/chat/convertTo.js
```diff
@@ -5,16 +5,18 @@ const { Workspace } = require("../../../models/workspace");
...
+  for (const item of preparedData) {
+    // Escape double quotes and wrap content in double quotes
+    const escapedPrompt = `"${item.prompt
+      .replace(/"/g, '""')
+      .replace(/\n/g, " ")}"`;
+    const escapedResponse = `"${item.response
+      .replace(/"/g, '""')
+      .replace(/\n/g, " ")}"`;
+    const row = `"${item.id}","${item.username}","${item.workspace}",${escapedPrompt},${escapedResponse},"${item.sent_at}"`;
+    rows.push(row);
```
timothycarambat, line 19: It would probably be more readable to define a data array that then makes the string
```javascript
const record = [
item.id,
escape(item.username), // we dont validate this to not contain quotes and better safe??
escape(item.workspace), // this needs to be escape because we dont validate this on backend.
escape(item.prompt)
escape(item.response)
item.sent_at
].join(',')

row.push(record)
```

server/utils/helpers/chat/convertTo.js
```diff
@@ -33,10 +35,27 @@ async function convertToJSONL(workspaceChatsMap) {
...
+        username: chat.user ? chat.user.username : "unknown user",
+        workspace: chat.workspace ? chat.workspace.name : "unknown workspace",
+        prompt: chat.prompt,
+        response: responseJson.text,
+        sent_at: chat.createdAt,
+      };
+    });
+
+    return preparedData;
+  }
```
timothycarambat, line 57: im not crazy about adding this here but I cant think of better way to do this or position it

timothycarambat (changes requested):

timothycarambat (approved):

PR #708, opened 2024-02-10
[FEAT] RLHF on response messages

### LINKED ISSUE #595: [FEAT]:  RHLF like feature.

An option for Vetting the queries-answers by different users and send it to the workspace embedding. I have this immense resource of question answers that can be used further to train the LLM - although I don't want to go fine-tuning way.

An option for the user to give thumbs-up or down; visible to the manager/ administrator in Workspace Chat pane - and the Administrator can send (after some editing) this to workspace embeddings so that some kind of RHLF happens. But these should be kept separate from the documents uploaded - for better folder management. Thanks

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #595 

### What is in this change?

- Implement RLHF in chats where user can thumbs up or down messages
- Create column in workspace_chats table to keep track of feedback score

### Additional Information

### Developer Validations

### DIFF
--- a/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
@@ -12,7 +12,16 @@ import createDOMPurify from "dompurify";
 const DOMPurify = createDOMPurify(window);
 const HistoricalMessage = forwardRef(
   (
-    { uuid = v4(), message, role, workspace, sources = [], error = false },
+    {
+      uuid = v4(),
+      message,
+      role,
+      workspace,
+      sources = [],
+      error = false,
+      feedbackScore,
+      chatId,
+    },
     ref
   ) => {
     return (
@@ -60,7 +69,11 @@ const HistoricalMessage = forwardRef(
           {role === "assistant" && !error && (
             <div className="flex gap-x-5">
               <div className="relative w-[35px] h-[35px] rounded-full flex-shrink-0 overflow-hidden" />
-              <Actions message={DOMPurify.sanitize(message)} />
+              <Actions
+                message={DOMPurify.sanitize(message)}
+                feedbackScore={feedbackScore}
+                chatId={chatId}
+              />
             </div>
           )}
           {role === "assistant" && <Citations sources={sources} />}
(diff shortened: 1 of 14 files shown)


shatfield4:
Currently works when refreshing page after sending chat or loading existing chat from historical messages.

WIP - Fix for after chat message is sent it needs to pass the chatId from last message to Actions component in order to update result.

frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
```diff
@@ -12,7 +12,16 @@ import createDOMPurify from "dompurify";
...
   (
-    { uuid = v4(), message, role, workspace, sources = [], error = false },
+    {
+      uuid = v4(),
+      message,
+      role,
+      workspace,
+      sources = [],
+      error = false,
+      feedbackScore,
```
timothycarambat, line 22: set default value, even if null

frontend/src/models/workspace.js
```diff
@@ -60,6 +60,16 @@ const Workspace = {
       .catch(() => []);
     return history;
   },
+  updateChatFeedback: async function (chatId, feedback) {
+    const result = await fetch(`${API_BASE}/workspace/chat-feedback`, {
+      method: "POST",
+      headers: baseHeaders(),
+      body: JSON.stringify({ chatId, feedback }),
```
timothycarambat, line 67: add chatID as url param with workspace slug as well.

Then we can pre-check the workspace x user ownership and then update on backend.

server/endpoints/workspaces.js
```diff
@@ -321,6 +321,26 @@ function workspaceEndpoints(app) {
...
 
+  app.post(
+    "/workspace/chat-feedback",
+    [validatedRequest],
+    async (request, response) => {
+      try {
+        const { chatId, feedback } = reqBody(request);
+        const result = await WorkspaceChats.updateFeedbackScore(
+          Number(chatId),
+          Number(feedback)
```
timothycarambat, line 332: Do the casting inside of the method here, would prevent accidental non-cast if we use it later.

PR #758, opened 2024-02-20
Enable ability to do full-text query on documents

### LINKED ISSUE #310: Summarization actions
Given the default nature of AnythingLLM is a Q/A against a document set, we should also enable an "action" via the prompt that can then do rote summarization of all text data attached to a specific workspace - if possible we should also enable filtering of summarization on specific documents in case their workspace contains many large documents.

`/summary <?doc> <prompt/>`

### LINKED ISSUE #729: [FEAT]: Document Pinning

Users should be able to "pin" a document in a workspace where the entire context of the document (context permitting) is placed into the system message.

Doing this will override almost all other documents and settings for a workspace since we will need to maximize context window space and provide ample room for responses.

timothycarambat:
Show alert modal on first pin for client
Add ability to use pins in stream/chat/embed

 ### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #729 
resolves #310

### What is in this change?

Enable the ability for the user to "pin" documents to the workspace. This does a full-text context injection into the prompt + vector-searching to ensure that important documents are always attached to a context window. This can resolve almost any relevancy problems depending on the model being used + the document content.

### Additional Information

### Developer Validations


timothycarambat (commented):
RTM

PR #835, opened 2024-02-28
[FIX] : Docker compose has a bad image reference

### LINKED ISSUE #834: [Docs]: Docker compose has a bad image reference

Docker (local)

### What happened?

git auth error on deploying. 

compose file has image: anything-llm:latest but should be mintplexlabs/anythingllm:latest

docker run or docker up

CRAZy-Monk3Y:
### Pull Request Type

- [x] ✨ feat
- [x] 🐛 fix
- [x] ♻️ refactor
- [x] 💄 style
- [x] 🔨 chore
- [x] 📝 docs

resolves #834

### What is in this change?

As suggested in the issue changed image name in docker compose file

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -9,7 +9,7 @@ networks:
 services:
   anything-llm:
     container_name: anything-llm
-    image: anything-llm:latest
+    image: mintplexlabs/anythingllm:latest
     build:
       context: ../.
       dockerfile: ./docker/Dockerfile


timothycarambat:
Closing this because this docker-compose is actually for development purposes and because of that we dont actually need the image tag at all. It was in there as a mistake, but this file should remain untouched as it is only used for quickly building/validating new changes in docker

PR #847, opened 2024-03-01
[FEAT] Ability to set workspace profile image

### LINKED ISSUE #706: [FEAT]: Ability to set workspace profile image

We currently allow images to be set for user profiles. We should allow workspaces to also have an image set which will be stored in an additional field on the `workspaces` table. This can be set by anyone who can modify that workspaces settings.

_this issue is blocked by #705_

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #706 

### What is in this change?

- Add the functionality to set the profile image for the assistant per workspace

### Additional Information

### Developer Validations


frontend/src/components/UserIcon/index.jsx
```diff
@@ -16,6 +18,16 @@ export default function Jazzicon({ size = 10, user, role }) {
...
+  useEffect(() => {
+    async function fetchWorkspacePfp() {
+      const pfpUrl = await Workspace.fetchPfp(user?.uid);
+      setWorkspacePfp(pfpUrl);
+    }
+    if (role === "assistant") {
+      fetchWorkspacePfp();
+    }
+  }, [role, user?.uid]);
+
```
timothycarambat, line 30: This will fire hundreds of times on a response chunk and will constantly make a call to the backend for the PFP. On only a small response it called over 80 times. We should be local-caching this image aggressively (maybe even when the workspace is GET on the chat page).

The image request should only be called once and then retrieved from cache then on until some expiration date.


Lastly, UseEffects need to return a function so rewriting this as 
```jsx
useEffect(() => {
    async function fetchWorkspacePfp() {
     if(role !== "assistant") return false;
      const pfpUrl = await Workspace.fetchPfp(user?.uid);
      setWorkspacePfp(pfpUrl);
    }
      fetchWorkspacePfp();
  }, [role, user?.uid]);
  ```
  
  makes more sense here

File: frontend/src/components/UserIcon/index.jsx [hunk 2]
```diff
@@ -26,6 +38,13 @@ export default function Jazzicon({ size = 10, user, role }) {
           className="absolute top-0 left-0 w-full h-full object-cover rounded-full bg-white"
         />
       )}
+      {role === "assistant" && workspacePfp && (
```
timothycarambat, line 41: When would WorkspacePfp be present and the role not be an assistant?

server/endpoints/workspaces.js
```diff
@@ -422,6 +431,127 @@ function workspaceEndpoints(app) {
...
+      } catch (error) {
+        console.error("Error processing the logo request:", error);
+        response.status(500).json({ message: "Internal server error" });
+      }
+    }
+  );
+
+  app.post(
+    "/workspace/:slug/upload-pfp",
+    [validatedRequest, flexUserRoleValid([ROLES.all])],
```
timothycarambat, line 472: I do not think `ROLES.all` is the correct permission set here. A default user should not be able to set this as they cannot view workspace settings already

server/endpoints/workspaces.js
```diff
@@ -422,6 +431,127 @@ function workspaceEndpoints(app) {
...
+      } catch (error) {
+        console.error("Error processing the profile picture upload:", error);
+        response.status(500).json({ message: "Internal server error" });
+      }
+    }
+  );
+
+  app.delete(
+    "/workspace/:slug/remove-pfp",
+    [validatedRequest, flexUserRoleValid([ROLES.all])],
```
timothycarambat, line 519: Same permission oversight here

timothycarambat (changes requested):

PR #956, opened 2024-03-22
handle expired token being null and prevent constant rerender

### LINKED ISSUE #947: [BUG]: Crash when navigating to localhost.

Docker (local)

### What happened?

Container starts fine, but when navigating to localhost:3001 it throws an error and the container stops.

Log:
2024-03-21 14:21:38 /app/server/node_modules/bcrypt/bcrypt.js:166
2024-03-21 14:21:38         throw new Error('data and hash arguments required');
2024-03-21 14:21:38               ^
2024-03-21 14:21:38 
2024-03-21 14:21:38 Error: data and hash arguments required
2024-03-21 14:21:38     at Object.compareSync (/app/server/node_modules/bcrypt/bcrypt.js:166:15)
2024-03-21 14:21:38     at validatedRequest (/app/server/utils/middleware/validatedRequest.js:41:15)
2024-03-21 14:21:38 
2024-03-21 14:21:38 Node.js v18.19.1

### Started the container and mounted local folder:

> $env:STORAGE_LOCATION="$HOME\Documents\anythingllm"; `
> If(!(Test-Path $env:STORAGE_LOCATION)) {New-Item $env:STORAGE_LOCATION -ItemType Directory}; `
> If(!(Test-Path "$env:STORAGE_LOCATION\.env")) {New-Item "$env:STORAGE_LOCATION\.env" -ItemType File}; `
> docker run -d -p 3001:3001 `
> --cap-add SYS_ADMIN `
> -v "$env:STORAGE_LOCATION`:/app/server/storage" `
> -v "$env:STORAGE_LOCATION\.env:/app/server/.env" `
> -e STORAGE_DIR="/app/server/storage" `
> mintplexlabs/anythingllm;

### Navigated to localhost:3001 in browser.

timothycarambat:
When using AnythingLLM with a password or multi-user auth state the default token expiry is set to 30 days expiration (TBD custom config for this value).

If you are logged in and do not re-visit the system for some time where the token then expires on revisit the token will fail to verify and will return a `null` value, which will fatally crash the `bcrypt` `compareHash` function. Passing an empty `''` instead when invalid will then cause an infinite re-render on the login screen.

Additional work was required to have this work when the session is fully expired:
- Invalidate token as normal
- Send the user to log in, but **prevent** auto login via a query param set during redirect due to invalid auth.
- User can log in again and get a new session token.

---------
- Tested as all roles in multi-user
- Tested with just password
- Tested with no password instance

------
resolves #947

### DIFF
--- a/frontend/src/components/Modals/Password/index.jsx
+++ b/frontend/src/components/Modals/Password/index.jsx
@@ -34,7 +34,7 @@ export default function PasswordModal({ mode = "single" }) {
   );
 }
 
-export function usePasswordModal() {
+export function usePasswordModal(notry = false) {
   const [auth, setAuth] = useState({
     loading: true,
     requiresAuth: false,
@@ -47,7 +47,7 @@ export function usePasswordModal() {
 
       // If the last validity check is still valid
       // we can skip the loading.
-      if (!System.needsAuthCheck()) {
+      if (!System.needsAuthCheck() && notry === false) {
         setAuth({
           loading: false,
           requiresAuth: false,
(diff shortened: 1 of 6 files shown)


timothycarambat:
resolves #947 947

PR #961, opened 2024-03-23
[FEAT] Check port access in docker before showing a default error

### LINKED ISSUE #959: [BUG]: Running anything llm container with --network host is hardcoded to give error

Docker (local)

### What happened?

image

The following docker command is fully functional, and allows me to use localhost rather than a docker internal name to access ollama port since we are setting network to be host when running anythingLLM:

```
docker run -d -p 3001:3001 --name "anythingllm" \                            
--cap-add SYS_ADMIN --network host \
-e STORAGE_DIR="/app/server/storage" \
mintplexlabs/anythingllm
```

However, even though it can connect and even find the models using `http://localhost:11434`, it seem to be hard-coded to give an error:

```
Failed to save LLM settings: Localhost, 127.0.0.1, or 0.0.0.0 origins cannot be reached from inside the AnythingLLM container. Please use host.docker.internal, a real machine ip, or domain to connect to your service.
```

Run a service such as ollama on your host machine:
``` 
# install ollama
curl -fsSL https://ollama.com/install.sh | sh

# run using deepseek-coder
ollama run deepseek-coder
```

Run anything LLM in a docker container, with network set to host
```
docker run -d --name "anythingllm" \  
--cap-add SYS_ADMIN --network host \
-e STORAGE_DIR="/app/server/storage" \
mintplexlabs/anythingllm
```

In AnythingLLM: Change the LLM url using Ollama to: http://localhost:11434
- Notice that it can find and load the models just fine  

When you press Save changes, a hardcoded error message occurs regardless if you can connect or not.

## Suggested Solution
Rather than checking if the URL matches localhost, 127.0.0.1, etc., the first attempt should attempt to load and connect when saving. If saving fails, then, and only then should you check for a localhost/127.0.0.1/etc. match before returning the same error message.

ivanskodje:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #959 

### What is in this change?

In validDockerizedUrl (which only runs in docker environments) we now do a port check before we simply _assume_ that the user cannot connect to the port. When you run docker in host network mode, it is perfectly fine to use localhost URLs.

### Additional Information

- Since host.docker.internal is not available in Linux systems, I have added the Docker URL for Linux in the error message.
- Added handling of async functions in a separate method, so you can now make a validation function async if needed

### Developer Validations

### DIFF
--- a/server/utils/helpers/updateENV.js
+++ b/server/utils/helpers/updateENV.js
@@ -1,5 +1,3 @@
-const isPortInUse = require("./portAvailabilityChecker");
-
 const KEY_MAPPING = {
   LLMProvider: {
     envKey: "LLM_PROVIDER",
@@ -468,18 +466,19 @@ async function validDockerizedUrl(input = "") {
   if (process.env.ANYTHING_LLM_RUNTIME !== "docker") return null;
 
   try {
+    const { isPortInUse, getLocalHosts } = require("./portAvailabilityChecker");
+    const localInterfaces = getLocalHosts();
     const url = new URL(input);
     const hostname = url.hostname.toLowerCase();
     const port = parseInt(url.port, 10);
 
+    // If not a loopback, skip this check.
+    if (!localInterfaces.includes(hostname)) return null;
     if (isNaN(port)) return "Invalid URL: Port is not specified or invalid";
 
-    const isPortAvailableFromDocker = await isPortInUse(port);
-    if (!isPortAvailableFromDocker) {
-      if (["localhost", "127.0.0.1", "0.0.0.0"].includes(hostname)) {
-        return "localhost, 127.0.0.1, or 0.0.0.0 origins cannot be reached from inside the AnythingLLM container. Please use host.docker.internal (for linux use 172.17.0.1), a real machine ip, or domain to connect to your service.";
-      }
-    }
+    const isPortAvailableFromDocker = await isPortInUse(port, hostname);
+    if (isPortAvailableFromDocker)
+      return "Port does not seem to running a service reachable on loopback address from inside the AnythingLLM container. Please use host.docker.internal (for linux use 172.17.0.1), a real machine ip, or domain to connect to your service.";
   } catch (error) {
     console.error(error.message);
     return "An error occurred while validating the URL";
--- a/server/utils/helpers/updateENV.js
+++ b/server/utils/helpers/updateENV.js
@@ -477,25 +477,14 @@ async function validDockerizedUrl(input = "") {
     const isPortAvailableFromDocker = await isPortInUse(port);
     if (!isPortAvailableFromDocker) {
       if (["localhost", "127.0.0.1", "0.0.0.0"].includes(hostname)) {
-        return "Localhost, 127.0.0.1, or 0.0.0.0 origins cannot be reached from inside the AnythingLLM container. Please use host.docker.internal, a real machine ip, or domain to connect to your service.";
+        return "localhost, 127.0.0.1, or 0.0.0.0 origins cannot be reached from inside the AnythingLLM container. Please use host.docker.internal (for linux use 172.17.0.1), a real machine ip, or domain to connect to your service.";
       }
     }
   } catch (error) {
-    console.error(error.message)
-    return "An error occurred while validating the URL"
+    console.error(error.message);
+    return "An error occurred while validating the URL";
   }
 
-  return null;
-}
-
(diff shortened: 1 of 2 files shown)


timothycarambat:
Just to double check this is essentially a patch for those who run docker with a different network config and when doing so:
1. Check if the port is running _anything_
2. If its empty -> go to error because there is no service there
3. If its live -> service is live and reachable from within the container.


I feel like this opens some issues whether we change it or not, because all it will take is someone to run LMStudio on 8888 and then docker would see the collector running locally even if it cant reach host 8888. However, this is a common config misstep for a lot of people no matter what we do as someone will always do it differently and mess it up.

Either way, just want to clarify I understand the use case here.

ZynthCode:
> Just to double check this is essentially a patch for those who run docker with a different network config and when doing so:
> 
> 1. Check if the port is running _anything_
> 2. If its empty -> go to error because there is no service there
> 3. If its live -> service is live and reachable from within the container.
> 
> I feel like this opens some issues whether we change it or not, because all it will take is someone to run LMStudio on 8888 and then docker would see the collector running locally even if it cant reach host 8888. However, this is a common config misstep for a lot of people no matter what we do as someone will always do it differently and mess it up.
> 
> Either way, just want to clarify I understand the use case here.

You've got it right.

I'm starting to think maybe the forceful validation wasn't the best move. It looks like we were trying to nudge folks towards using Docker's internal IPs if they're in a container. Maybe a heads-up notice when running in Docker would've been a lot better, giving them a tip instead of a hard "No, you can't use localhost or your local IP in Docker." validation step. 

At the very least, I'm hoping this approach will reduce frustration for the time being. 😸

timothycarambat:
It was really never the right choice i just got sick of closing issues and copy-pasting the same comment 1 million times a day because people never read the !IMPORTANT label on the `Recommended way to use Docker` in the docs 😢 

So that was the only reason that ever even existed and who knows how many ctrl+v's its saved me to date haha

ZynthCode:
@timothycarambat  @ me if you need / wish for any changes :+1:

ivanskodje:
> Just to double check this is essentially a patch for those who run docker with a different network config and when doing so:
> 
> 1. Check if the port is running _anything_
> 2. If its empty -> go to error because there is no service there
> 3. If its live -> service is live and reachable from within the container.
> 
> I feel like this opens some issues whether we change it or not, because all it will take is someone to run LMStudio on 8888 and then docker would see the collector running locally even if it cant reach host 8888. However, this is a common config misstep for a lot of people no matter what we do as someone will always do it differently and mess it up.
> 
> Either way, just want to clarify I understand the use case here.

Just give me a call if you need any changes to the PR. You understood what this does perfectly. :star:

timothycarambat:
@ivanskodje  if you are in Discord im happy to give you a contributor badge if you hang out in there. Just ping me

server/utils/helpers/updateENV.js
```diff
@@ -468,18 +466,19 @@ async function validDockerizedUrl(input = "") {
...
 
   try {
+    const { isPortInUse, getLocalHosts } = require("./portAvailabilityChecker");
+    const localInterfaces = getLocalHosts();
     const url = new URL(input);
     const hostname = url.hostname.toLowerCase();
     const port = parseInt(url.port, 10);
 
+    // If not a loopback, skip this check.
+    if (!localInterfaces.includes(hostname)) return null;
```
ivanskodje, line 476: A tip @timothycarambat :

Unless you are planning or expect to be using `getLocalHosts` separately in the near future, you might be better off getting the boolean check from within portAvailabilityChecker.ts directly.

Rather than asking for data to work with by using `getLocalHosts` to fetch a local variable localInterfaces, we can instgead expose the boolean check itself, and expose a function such as for example `isLoopbackAvailable(hostname)`.

This way we can remove an unnecessary variable and remove the comment that is currently needed to explain what we are checking.

For example:
```
const { isPortInUse, isLoopbackAvailable } = require("./portAvailabilityChecker");

...

if (!isLoopbackAvailable(hostname)) return null;

```

timothycarambat: I am expecting to use it elsewhere right now! Might use it in the `desktop` branch for another config of a future LLM or Image-gen provider that can run locally!

PR #1091, opened 2024-04-12
[FEAT] #1089 Collector: allow more file type extensions

### LINKED ISSUE #1089: [FEAT]: Allow more coding-related filetypes which are still txt files.

Uploading documents should allow more source code files from common programming languages, like C#, Pascal (.lua already works).

Specifically, the mime.js could use some expansion of allowed file extensions, like '.cs', '.csproj', '.pas'.

tobitege:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1089 

### What is in this change?

The constants.js file in the Collector contains a list of allowed file extensions/mimes.
These have been expanded for some common development languages, like C, C#, Javascript, Lua, Pascal, Python and Typescript.
They're all treated as text files.

### Additional Information

Tried yarn lint, but it fails in WSL2 with command prettier not found and I'm not a Linux guy.

### Developer Validations

### DIFF
--- a/collector/utils/constants.js
+++ b/collector/utils/constants.js
@@ -1,7 +1,7 @@
 const WATCH_DIRECTORY = require("path").resolve(__dirname, "../hotdir");
 
 const ACCEPTED_MIMES = {
-  "text/plain": [".txt", ".md", ".org", ".adoc", ".rst"],
+  "text/plain": [".txt", ".md", ".org", ".adoc", ".rst", ".c", ".cs", ".h", ".js", ".lua", ".pas", ".py", ".ts"],
   "text/html": [".html"],
 
   "application/vnd.openxmlformats-officedocument.wordprocessingml.document": [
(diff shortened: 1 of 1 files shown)


timothycarambat:
Before merging this, i wanted to verify some functionality since I dont have all those filetypes on hand, but the default behavior of unknown files is to attempt to parse as UTF-8! So this functionality likely already exists and it was implied instead of written because we will never be able to add every file required.

Here you can see if the file extension is not _specifically supported_ because it does not require some special library to parse (https://github.com/Mintplex-Labs/anything-llm/blob/102689a77abb3271f6362855c4d90607c76f29b6/collector/processSingleFile/index.js#L50-L64) we then use `isTextType` from out internal Mime.js type detector (https://github.com/Mintplex-Labs/anything-llm/blob/master/collector/utils/files/mime.js)

You can see here this exists for a multitude of files (including python, c and ts). The whole reason this flow exists is because
```
// the .ts extension maps to video/mp2t because of https://en.wikipedia.org/wiki/MPEG_transport_stream
// which has had this extension far before TS was invented. So need to force re-map this MIME map.
```

So instead of merging this PR, we could just add the extensions to here and they will always be forced to text, however
`".c", ".cs", ".h", ".js", ".lua", ".pas", ".py"` all should upload and parse as text already with no additional code.

If any of those fail, we should add them to https://github.com/Mintplex-Labs/anything-llm/blob/102689a77abb3271f6362855c4d90607c76f29b6/collector/utils/files/mime.js#L24 and not in the "SUPPORTED FILETYPES` const.

tobitege:
Thanks for the detailed response!  

Let me re-check that the filetypes work for me or not and get back to you here soon-ish.
Cheers

tobitege:
@timothycarambat so LUA files can be uploaded, but .cs and .pas files are rejected, like this:
grafik

I'll look into changing my PR according to your remarks re: mime.js then :)

PR #1191, opened 2024-04-26
[FEAT] Website depth scraping data connector

### LINKED ISSUE #1190: [FEAT]: Website scraping depth

We should create a data connector that allows scraping websites with a configurable depth

- Should scrape all `<a>` tags on the site and look for all links that match the original domain name then scrape X levels deep
- Should all be done using the existing puppeteer instance that is already used for scraping links

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1190 

### What is in this change?

- Create data connector that will scrape to X depth of links on site
- Only finds links with matching domain name on site to scrape only links that are on the same website

### Developer Validations


shatfield4:
@timothycarambat, refactored based on what we discussed.

- Creates array of all links so we know how many links before main scraping starts
- Passes the array to bulk scraping function

PR #1313, opened 2024-05-08
[FIX] Patch usernames to be case insensitive when creating users and logging in

### LINKED ISSUE #1312: [BUG]: Usernames should not be case-sensitive when logging in

Docker (remote machine)

### What happened?

Today, you can register as USERNAME even if someone else is already registered as username. Username should be case insensitive and unique.

1. Create a user with username "admin
2. Create another user with username "Admin" starting with uppercase "A"
This should not be allowed, an error should be displayed that the user already exists.

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1312 

### What is in this change?

- Convert all usernames to lowercase in backend to make usernames case insensitive 

### Developer Validations

### DIFF
--- a/frontend/src/pages/Admin/Users/NewUserModal/index.jsx
+++ b/frontend/src/pages/Admin/Users/NewUserModal/index.jsx
@@ -15,7 +15,13 @@ export default function NewUserModal({ closeModal }) {
     for (var [key, value] of form.entries()) data[key] = value;
     const { user, error } = await Admin.newUser(data);
     if (!!user) window.location.reload();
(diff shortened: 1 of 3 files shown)


giovanniscalar:
In addition a `usernameCheck` function should be added to apply different validations such as username cannot use spaces.
image

PR #1341, opened 2024-05-10
[FEAT] Implement regenerate response button

### LINKED ISSUE #1276: [FEAT]: Regenerate button

I really miss a "Regenerate" button just before "copy to buffer" button in desktop version. It will erase last LLM answer and resubmit history to LLM. 

It would be also very nice if there would be ability to edit user input and erase question and answer all together from the history.

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1276 

### What is in this change?

- Implement a new action for regenerating assistant response in workspace chat

### Developer Validations

### DIFF
--- a/frontend/src/components/Modals/Password/MultiUserAuth.jsx
+++ b/frontend/src/components/Modals/Password/MultiUserAuth.jsx
@@ -1,7 +1,6 @@
 import React, { useEffect, useState } from "react";
 import System from "../../../models/system";
 import { AUTH_TOKEN, AUTH_USER } from "../../../utils/constants";
-import useLogo from "../../../hooks/useLogo";
 import paths from "../../../utils/paths";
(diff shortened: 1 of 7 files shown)


koolong:
Hi, is it possible to expose this regenerate response function to developer API? I noticed that this function uses delete-chats internal API. Or perhaps expose this delete-chats API to developer API?

timothycarambat:
@koolong Please make a feat request. In theory we should be able to allow you to specify a chatID and a `regenerate` endpoint? Or maybe some other form factor?

Can we understand a bit more about the use case so we can design a better API expierence?

PR #1352, opened 2024-05-11
prevent accidental lockout from restricted chars in single pass mode

### LINKED ISSUE #1289: [BUG]: An admin password containing `, ' or # does not work

Docker (local)

### What happened?

I had used a pre-generated password from keepass that contained a `` ` ``  e.g. ``H$byw&52nb`7``
Upon restart the password was not recognised until I tried `H$byw&52nb`

- Do first time setup Docker image version of AnythingLLM
- Choose an admin password containing a `` ` `` 
- Log out, close browser tab, stop docker container
- Start docker container and try to log in again using the saved password
- Failure to validate credentials

timothycarambat:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1289

### What is in this change?

Prevent a password-protected instance from being self-owned by accidentally using `.env` file-protected chars in their password since they will be either removed or escaped and therefore the password will no longer match or will be broken on instance reboot as loading from the `.env` on boot will only boot what the system can read as opposed to what the user actually wanted the password to be.

Regex: `/^[a-zA-Z0-9_\-!@$%^&*();]+$/`

### Developer Validations

### DIFF
--- a/frontend/src/pages/GeneralSettings/Security/index.jsx
+++ b/frontend/src/pages/GeneralSettings/Security/index.jsx
@@ -190,6 +190,7 @@ function MultiUserMode() {
   );
 }
 
+const PW_REGEX = new RegExp(/^[a-zA-Z0-9_\-!@$%^&*();]+$/);
 function PasswordProtection() {
   const [saving, setSaving] = useState(false);
   const [hasChanges, setHasChanges] = useState(false);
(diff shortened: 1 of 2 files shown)


Propheticus:
works
image

PR #1477, opened 2024-05-21
Improve RAG responses via source backfilling

### LINKED ISSUE #1240: [CHORE]: Improve RAG retention via history context

When we send a chat like `What is AnythingLLM` the system will find relevant context, if any and append it under the system prompt.
```shell
{SYSTEM}
<context>
<context>
# No history
<user prompt>
```

However, on follow up request, where we may not get a vector search result due to ambiguity like `tell me some features`
We will get a flow like

```shell
{SYSTEM}
# NO CONTEXT!
<reply from prompt 1 from LLM>
<user prompt>
```

This puts 100% of the response result on the `<reply from prompt 1 from LLM>` response where we could leverage the previous citations in the history to help offload the responsibility and scope of the answering.

timothycarambat:
resolves #1240
 
 ### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 🔨 chore
- [ ] 📝 docs

> [!WARNING]
> This PR enables `query` to work with chat history now but **only** in the app. Behavior in the embed is the same.

In an effort to improve RAG results without adding unwarranted or "conditional" middleware like ReRankers, there is a lot of room for improvement of the traditional RAG pipeline AnythingLLM uses. This PR includes the ability to **backfill** sources during a chat session.

## What is backfilling?

Currently, AnythingLLM will append `Context` snippets to the `system` prompt on each chat. These contexts come from the attached VectorStores `.similaritySearch` method. While this works for direct questions this does not scale well as chat's become higher context or more vague. This leads to a bad RAG experience in a traditional chat, but is far worse in `query` where a follow-up query, that is contextually relevant, is deemed irrelevant.

## Example

**prompt 1**: "What is anythingllm?" 
-  possibly get 4 good sources and get a good LLM response as it is operating with high context.

**prompt 2**: "Tell me some features" 
  - This question is appropriate, topical, and a decent inquiry but standalone is quite vague when viewed outside of the context of the chat.
  - Possible get 0 - 1 maybe relevant sources
  - We now must rely on the _maybe_ existent snippet for source and hope the previous LLM response has something worth looking at

_introduced in this pr_

**prompt 1**: "What is anythingllm?" 
-  possibly get 4 good sources and get a good LLM response as it is operating with high context.

**prompt 2**: "Tell me some features" 
  - Possible get 0 - 1 maybe relevant sources
  - We now can backfill the context window in this order of priority
     - Pinned documents
     - This queries search results
     - Work backwards in chat history citations to backfill the source window to the workspace's assigned snippet window value.
 
 
 With this, we can now ensure that follow-up questions will be more relevant and contextually pertinent as a conversation continues. 
 
 **"What if the contextual topic changes"?**
 - This is handled as we prioritize sources from search results  first, then backfill in by more recent chats with historical references
 - If you have a relevant contextual change in the chat, those sources will take precedence over the historical ones and will become the new historical reference

### DIFF
--- a/server/utils/chats/index.js
+++ b/server/utils/chats/index.js
@@ -151,16 +151,27 @@ async function chatWithWorkspace(
     };
   }
 
-  contextTexts = [...contextTexts, ...vectorSearchResults.contextTexts];
+  const { fillSourceWindow } = require("../helpers/chat");
+  const filledSources = fillSourceWindow({
+    nDocs: workspace?.topN || 4,
+    searchResults: vectorSearchResults.sources,
+    history: rawHistory,
+    filterIdentifiers: pinnedDocIdentifiers,
+  });
+
+  // Why does contextTexts get all the info, but sources only get current search?
+  // This is to give the ability of the LLM to "comprehend" a contextual response without
+  // populating the Citations under a response with documents the user "thinks" are irrelevant
+  // due to how we manage backfilling of the context to keep chats with the LLM more correct in responses.
+  // If a past citation was used to answer the question - that is visible in the history so it logically makes sense
+  // and does not appear to the user that a new response used information that is otherwise irrelevant for a given prompt.
+  // TLDR; reduces GitHub issues for "LLM citing document that has no answer in it" while keep answers highly accurate.
+  contextTexts = [...contextTexts, ...filledSources.contextTexts];
   sources = [...sources, ...vectorSearchResults.sources];
 
-  // If in query mode and no sources are found from the vector search and no pinned documents, do not
+  // If in query mode and no context chunks are found from search, backfill, or pins -  do not
   // let the LLM try to hallucinate a response or use general knowledge and exit early
-  if (
-    chatMode === "query" &&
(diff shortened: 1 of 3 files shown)


shatfield4:
LGTM

Propheticus:
Pulled latest docker image. Chat in query mode, looks like it works:
image
As designed no citations are shown (because no hits on last query), but still an answer is given based on earlier context in the conversation. 💯 

(also confirmed by checking LM Studio logs)

DangerousBerries:
> Pulled latest docker image. Chat in query mode, looks like it works: image As designed no citations are shown (because no hits on last query), but still an answer is given based on earlier context in the conversation. 💯
> 
> (also confirmed by checking LM Studio logs)

Huh, I thought all this time that only the last message mattered. Why is it not showing the citations for Context snippets?

DangerousBerries:
Is there a way to see what Context snippets are added to the system prompt?

Issue #1494, opened 2024-05-22
[BUG]: Render upload issue


RahSwe:
### How are you running AnythingLLM?

Docker (remote machine)

### What happened?

Starting a fresh Render server, some files are not possible to upload, getting the error report below:

<img width="581" alt="image" src="https://github.com/Mintplex-Labs/anything-llm/assets/163884933/e6064b9e-b76b-45ae-bd3c-93f31e84c52f">

The "unuploadable" files are however possible to upload to another instance running on Render.

RahSwe:
To avoid confusion, the "working instance" has not been redeployed the last 48 hrs

RahSwe:
Edit: It does not seem to be a Render issue. I started a new AWS instance, with the same result, i.e. same files are not possible to upload with the same error message.

timothycarambat:
@RahSwe what exactly are these types of files you are trying to upload. Unable to reproduce otherwise

RahSwe:
@timothycarambat Trying to attach such a file below

[Advokaten och alibivittnet - NJA 2021 s. 576.pdf](https://github.com/Mintplex-Labs/anything-llm/files/15409009/Advokaten.och.alibivittnet.-.NJA.2021.s.576.pdf)

RahSwe:
@timothycarambat I downloaded the file which I uploaded and then tried to upload it to ALLM. It worked. The difference seems to be that punctuations were added in the filename when uploaded to Github. I.e. it seems that it is not possible to upload file witch spaces in the document name.

shatfield4:
@RahSwe if you take the file that is failing to upload (not downloaded from github) and rename it to `test`, does it upload properly?

shatfield4:
Are you able to copy and paste the exact filename here in github? I am thinking there is a hidden character where the spaces are which is causing this error to happen.

RahSwe:
@shatfield4 
This is copy of the file name of an unuploadablie, from "God Advokatsed (Bentelius, 1998).pdf"

It should be just regular spaces. It works fine to upload the file to previous versions of ALLM.

RahSwe:
@timothycarambat @shatfield4 
Further tests:
I have managed to upload most files by replacing spaces with underlines.
However this file still cannot be uploaded, even without spaces.
It however works if I rename it by removering the second part of the document name.

[...]

PR #1516, opened 2024-05-23
repatch path normalization

timothycarambat:
Repatch path normalization to validate pathing within destination folders


timothycarambat:
resolves #1494

PR #1561, opened 2024-05-29
[FIX] Incorrect vectors count with Qdrant

### LINKED ISSUE #1220: [BUG]: qdrant/qdrant:latest (1.9.0) causes embeddings to be ignored

Docker (local)

### What happened?

when using the latest version of qdrant locally I would add embeddings to a workspace, be notified that the embeddings had been added successfully and then check the workspace settings to find the vector count for that workspace to be 0. 

This seems to be caused by the latest (1.9.0) version of [qudrant on docker hub](https://hub.docker.com/r/qdrant/qdrant/tags) no longer providing a `vectors_count`. For some reason this seems to mess with the anythingllm code. 

The more important symptom is that all queries to this index via chat always return an empty sources dict, even with no minimum distance.

This may have been related to what a commenter was saying in #516 

> I don't know if it is just for me but the number of vectors in the Settings tab for all of my workspaces are showing zero even after leaving and reloading the Setting dialog. I have about 13 documents in each workspace and they are successfully referenced in chats but it always says 0 vectors in Settings for all of my workspaces.

1. clone the latest anythingllm
2. follow the local docker set up instructions in `docker/HOW_TO_USE_DOCKER.md`
3. set up qdrant locally by following the [qdrant quickstart documentation](https://qdrant.tech/documentation/quick-start/) i.e. 

```
docker run -p 6333:6333 -p 6334:6334 \
    -v $(pwd)/qdrant_storage:/qdrant/storage:z \
    qdrant/qdrant
```

4. look at the terminal output to confirm that we are now using the 1.9.0 version of the qdrant docker image
5. Set up as usual choosing any llm provider
6. alter the vectordb to the qdrant, passing in the local url (no api key needed unless you set one in step 3)
7. create a new workspace, add some embeddings to id
8. check the settings for that workspace, vector count will be 0
9. try some queries, there will be no sources

Now just change to `qdrant/qdrant:v1.8.4`, create a new index and everything will work as normal.

Anush008:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

Resolves #1220 

### What is in this change?

With the 1.9 release of Qdrant, the `vectors_count` attribute has been removed from the collection info. This PR updates the implementation to use the [count API](https://qdrant.tech/documentation/concepts/points/#counting-points) instead.


timothycarambat:
@Anush008 ping me for contributor badge in discord!

PR #1607, opened 2024-06-04
[FEAT] Rename new threads on thread creation

### LINKED ISSUE #701: [FEAT]:  Can thread names be automatically generated based on key information from the chat, rather than manually renaming the thread names?

 Can thread names be automatically generated based on key information from the chat, rather than manually renaming the thread names?

image

### LINKED ISSUE #1454: [FEAT]: General UI Feedback

Apologies for the novel here. Stick with it though :) Isn’t as big of an ask as it seems. Thanks

wimage

First. 
New threads get the title of New Thread. That’s fine by itself. But not when the button to create a new thread goes by the same call sign on the screen. It’s too repetitive, and too close together. One thing *is* something while another *does* something… in the same spot. Not everybody takes the time to rename the threads. It would serve folks better to have a pretty timestamp as the initial name of the thread. That way, they can tell a bit more context than New Thread. This would take very little time to tweak. Moment to the rescue. This week, I must have clicked the New Thread button 50 times to open an existing thread only to find out it opened another thread. They multiply like rabbits because the eyes can’t tell them apart. 

Second. 
The button named New Thread isn’t so bad by itself. It *is* clunky & redundant verbiage, though, and paired with the plus sign, and being a chat, it is very obvious what it does. In fact, the 2 extra words contribute to a more busy ui. If that is desired, then hey, whatever. It isn’t my product. But after using it a while, honestly, seeing “New Thread” “New Thread” all over the place… it’s a bit tiring to look at and use. I’m sure that is not the feeling you want it to invoke. The New Workspace button’s core color is a perfect starting color for the button to add a new thread. If you’re looking for an alternative or another riff to my suggestion above, maybe wrap the New Thread text up inside a white button with the plus sign (exactly like the New Workspace button - hey theming!) and then extend the workspace “tree” down to the button. 

Third. 
The new workspace button color is great, but for being such a big button in the top of the screen it is a little “loud” and top-heavy feeling. It makes all the other buttons not as interesting, especially when it precedes them . Could you keep the color but mute it a bit? To be completely honest, I think the best thing to do here is swap the color scheme of the new workspace button with the color scheme of the currently selected workspace. That way a user’s eye is drawn to the thing they are working on instead of distracting them. 

Welp, there’s my thoughts after spending a week straight looking at it every day. And hopefully you have some ideas flowing for making things look more polished. 

I have a lot more feelings on the ui in general too, if you’re interested. Dropped some suggestions into discord, focused on the document indexing/upload modal. 

We are about to deploy a customer chat bot using anythingLLM this summer, so looking forward to using it more and getting set up in k8s. 

image

### LINKED ISSUE #1606: [FEAT]: Rename "New thread" when thread is created

- When creating a new thread, the new thread should not be named "New thread" to avoid confusion
- On new thread creation, after first chat message is sent, we should use the LLM provider to generate a title for the thread

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1606 
resolves #1454
resolves #701 

### What is in this change?

- When a new thread is created, we name the thread "Thread" instead of "New thread" to avoid confusion
- Truncate first message sent to thread and set as thread name to differentiate threads

### Developer Validations


shatfield4:
WIP automatic thread renaming

shatfield4:
@timothycarambat

- [ ] Removed LLM used for thread naming
- [ ] Uses custom events/chatHandler event listener on frontend to rename thread

Issue #1402, opened 2024-05-15
[BUG]: Container cashing with docker run command provided in the documentation


clipod:
### How are you running AnythingLLM?

Docker (local)

### What happened?

I am trying to run anythingllm in local docker. My docker is installed on a Debian instance. I am running the command given in the documentation as root(unfortunately I dont have a choice). 
```
export STORAGE_LOCATION=/tmp/anything-llm/docker && mkdir -p $STORAGE_LOCATION && touch "$STORAGE_LOCATION/.env" && docker run -d -p 3001:3001 --cap-add SYS_ADMIN -v ${STORAGE_LOCATION}:/app/server/storage -v ${STORAGE_LOCATION}/.env:/app/server/.env -e STORAGE_DIR="/app/server/storage" mintplexlabs/anythingllm
```

I am getting the below error, 
```
Collector hot directory and tmp storage wiped!
Document processor app listening on port 8888
Environment variables loaded from .env
Prisma schema loaded from prisma/schema.prisma
✔ Generated Prisma Client (v5.3.1) to ./node_modules/@prisma/client in 197ms
Start using Prisma Client in Node.js (See: https://pris.ly/d/client)

import { PrismaClient } from '@prisma/client'
const prisma = new PrismaClient()

or start using Prisma Client at the edge (See: https://pris.ly/d/accelerate)

import { PrismaClient } from '@prisma/client/edge'
const prisma = new PrismaClient()

See other ways of importing Prisma Client: http://pris.ly/d/importing-client
Environment variables loaded from .env
Prisma schema loaded from prisma/schema.prisma
Datasource "db": SQLite database "anythingllm.db" at "file:../storage/anythingllm.db"
Error: Schema engine error:
SQLite database error
unable to open database file: ../storage/anythingllm.db
```

Not sure what I am doing wrong here.

timothycarambat:
`/tmp/` is a non-writable directory by the docker user. This typically works better on *inx system when you either give the docker-user the permission to write at those top level folders or just run the container in the $HOME of the user running the container.

hmazomba:
Ok but bruv that doesn't help because I am using my $HOME

timothycarambat:
Sorry i didn't see the second comment for some reason, clearly you did that

> unable to open database file: ../storage/anythingllm.db

[...]

PR #1611, opened 2024-06-05
add chmod 777 -R $STORAGE_LOCATION && \

JOduMonT:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [x] 📝 docs

### Relevant Issues

`chmod 777 -R $STORAGE_LOCATION && \`
for less comment like this ;)
- https://github.com/Mintplex-Labs/anything-llm/issues/1402

### Developer Validations

### DIFF
--- a/docker/HOW_TO_USE_DOCKER.md
+++ b/docker/HOW_TO_USE_DOCKER.md
@@ -54,6 +54,7 @@ docker pull mintplexlabs/anythingllm
 ```shell
 export STORAGE_LOCATION=$HOME/anythingllm && \
 mkdir -p $STORAGE_LOCATION && \
+chmod 777 -R $STORAGE_LOCATION && \
 touch "$STORAGE_LOCATION/.env" && \
 docker run -d -p 3001:3001 \
 --cap-add SYS_ADMIN \


cyberlink1:
chmod 777 is a bad security practice, especially in a docker container, it would be better to allow the user to set the UID and GID then have the container check the permissions. 

The default container deploy on my docker swarm uses UID 1000 and GID 1000 which seems to be common among docker containers. So you can set the directory with "chown 1000:1000 <directory>" rather than doing a chmod 777.

docker/HOW_TO_USE_DOCKER.md
cyberlink1, line 1: chmod 777 is global read/write and is a bad security practice.

timothycarambat:
I understand why you would want this, because so many people are bitten by it, but I would hesitate to add this because all someone needs to do is mount this to some bad root level folder and accidentally cause a bad day for them in the future.

That command is using `$HOME` which should always be okay, so it should make the `chmod` duplicitous. Honestly, id rather answer the support question constantly then accidentally set someone up in a way that can have bad security implications

PR #1668, opened 2024-06-13
[FEAT] Workspace threads developer API endpoints

### LINKED ISSUE #1552: [FEAT]: API endpoints to create/delete threads within workspace

- Create API endpoints with the ability to create and delete threads via API
- Create API endpoints to chat with workspace threads

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1552 

### What is in this change?

- Allow developers to create and delete threads via the developer API
- Allow developers to chat with workspace threads by user id

### Developer Validations

### DIFF
--- a/server/endpoints/api/index.js
+++ b/server/endpoints/api/index.js
@@ -4,6 +4,7 @@ const { apiAuthEndpoints } = require("./auth");
 const { apiDocumentEndpoints } = require("./document");
 const { apiSystemEndpoints } = require("./system");
 const { apiWorkspaceEndpoints } = require("./workspace");
+const { apiWorkspaceThreadEndpoints } = require("./workspaceThread");
 
 // All endpoints must be documented and pass through the validApiKey Middleware.
 // How to JSDoc an endpoint
@@ -16,6 +17,7 @@ function developerEndpoints(app, router) {
   apiSystemEndpoints(router);
   apiWorkspaceEndpoints(router);
   apiDocumentEndpoints(router);
+  apiWorkspaceThreadEndpoints(router);
 }
 
 module.exports = { developerEndpoints };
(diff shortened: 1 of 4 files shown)


darkburak:
Looks good. Goed dat het via workspaces threads toch mogelijk was.

server/endpoints/api/index.js
```diff
@@ -16,6 +17,7 @@ function developerEndpoints(app, router) {
   apiSystemEndpoints(router);
   apiWorkspaceEndpoints(router);
   apiDocumentEndpoints(router);
+  apiWorkspaceThreadEndpoints(router);
```
timothycarambat, line 20: This should be a sub-module of the REST API, not its own module.

server/endpoints/api/workspaceThread/index.js
```diff
@@ -0,0 +1,594 @@
+const { v4: uuidv4 } = require("uuid");
```
timothycarambat, line 1: While some of these endpoints make sense to be within a `Workspace Thread` model/object/tag. Additionally, How does this API work when not in multi-user mode as well as in multi-user?

In single-user you typically are working as the user when in the API and in multi-user you are not even a user. I see one endpoint has a param for a userID, but there is no User API - so how would one retrieve the id for such a call?

timothycarambat (changes requested):
See comments about scope and implementation.

server/endpoints/api/workspaceThread/index.js
timothycarambat: Think we may need to expand some more API calls to user management as well as right now it would be cumbersome to find/guess user ids as i dont think we show them anywhere

timothycarambat (approved):

PR #1771, opened 2024-06-27
1603 speech to text hotkey

### LINKED ISSUE #1603: [FEAT]: Add speech-to-text prompting possibility activated with a key combination

Adding the possibility of prompting the chat with speach would greatly improve the user experience because of the time optimisation. 

The common use case for the AnythingLLM is to perform RAG on some documents/books/other source materials and then Q&A against them. The process usually involves research performed by the user while augmenting his processing capabilities by asking about some data/calculations the LLM. 

Currently the flow looks like this:

Research document -> question arises -> switch window and your mind focus to the AnythingLLM -> think about precise way of specifying the prompt -> Write the prompt and wait for the answer. 

It doesn't seem like a big trouble, but considering that:

1 - you are doing it very often
2 - it causes you to lose focus on the thing/though that you had before (what is the big distraction as it takes time to re-focus again)

It can and will grow into a significant amount of time.

What I'd love to see is the following:

Research document -> press key combination -> ask the question verbally -> continue to research document without losing focus & eye contact -> eventually see the LLM answer afterwards

Rrojaski:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1603

### What is in this change?

Implemented Ctrl (Cmd on Mac) + 'm' hotkeys to activate the existing React Speech Recognition feature.

### Developer Validations


timothycarambat:
Love the idea. I have some feedback currently from a UI perspective. When I press CMD+Enter the mic turns on but is not super obvious my hotkey did that (the icon just gets white for the mic). Then I can speak and I see the words coming in, but then I don't know how to really "stop" the streaming other than shutting up for the silence interval.

If I click on the mic icon, it'll auto-send, but if I click the send icon itself it will double-send my query as the mic gets the event to stop listening.

Thinking about a UX perspective that makes this less confusing for someone who might find themselves toying with this hot-key combination

Rrojaski:
Thanks for the feedback! It looks like the query was being double-sent by clicking the mic button or by pressing the hotkeys because of a timeout inside the SpeechToText component. I've added an event listener for PROMPT_INPUT_EVENT to check if the user clicks send before the timeout finishes.

I've changed the hotkeys to CMD+SPACE to avoid interfering with what the ENTER key currently does.

I'm still considering a better user experience for stopping the speech recognition. Do you have any recommendations?

Side note: I'm not sure if it's just me, but "SpeechRecognition.stopListening()" doesn't appear to stop the speech recognition.

Issue #1922, opened 2024-07-22
[FEAT]: data connector to batch load repositories


DipFlip:
### What would you like to see?

It would be useful to be able to collect many git repositories at once through a data connector.

Perhaps by inputting multiple repository addresses separated by return and then loading them at once or through uploading a `repos.txt` file containing all repo addresses to load.

Another solution could be to recursively load git submodules present in a repo that is collected. Then a user could create a new repo where they add all the repositories they would like to collect and submit this repo in AnythingLLM's data connector.

PR #1934, opened 2024-07-23
1922 batch repo collection

DipFlip:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves https://github.com/Mintplex-Labs/anything-llm/issues/1922

### What is in this change?

This updates the GitHub and GitLab data connectors to accept multiple repo addresses at once. Repos are entered one per line in the input field and will be collected one by one after the user clicks `Submit`.

### Developer Validations


timothycarambat:
I can see the utility here, but I dont think we can merge this in its current state as one detail is missing and will likely lead to issues being opened. If you have two+ repos but want different branches from each repo there is no way to do that currently.

Going to close for now since to accommodate that we would need to entirely re-work the entire data connector to accommodate this since each repo would have its own independent config 🤷 

It does indeed work,but only if all the repos in the list want to have the exact same branch pulled - which is not often the case and would make more sense to be single-repo input for now even though it is more time consuming...

PR #1942, opened 2024-07-24
Render HTML tags inside table row in Markdown (#1940)

### LINKED ISSUE #1940: [BUG]: I found <br> or HTML tags inside table row in Markdown not rendered

Local development

### What happened?

Markdown content

Result after rendered

_No response_

007900:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1940

### What is in this change?

Change configuration by enable HTML tags in source

### Developer Validations

### DIFF
--- a/frontend/src/utils/chat/markdown.js
+++ b/frontend/src/utils/chat/markdown.js
@@ -5,7 +5,7 @@ import "highlight.js/styles/github-dark-dimmed.min.css";
 import { v4 } from "uuid";
 
 const markdown = markdownIt({
-  html: false,
+  html: true,
   typographer: true,
   highlight: function (code, lang) {
     const uuid = v4();


frontend/src/utils/chat/markdown.js
```diff
@@ -5,7 +5,7 @@ import "highlight.js/styles/github-dark-dimmed.min.css";
 import { v4 } from "uuid";
 
 const markdown = markdownIt({
-  html: false,
+  html: true,
```
timothycarambat, line 8: Unfortunately, we cannot enable this as it is _technically_ unsafe.
https://github.com/markdown-it/markdown-it/blob/master/docs/security.md#security

The reason is was changed a while back was because of an odd issue during streaming where if you are streaming an input and under the proper circumstances the chunks come in with a `<script>alert('XSS');</script>` it would execute that script instead of rendering it. This also can occur with `img` and other tags. 

So while this _does work_ for your specific issue we cannot enable it this way as it will retroactively undo a fix for another bug

timothycarambat (changes requested):
See comment

timothycarambat:
Checkout the proposed fix on https://github.com/Mintplex-Labs/anything-llm/pull/1958 and see if that works for you since you are running locally.

frontend/src/utils/chat/markdown.js
007900: Oh, I apologize for not noticing that such a case would occur with that setting. Thank you very much for resolving that issue for me.

timothycarambat: I didnt think about it either until is happened to someone asking a question about XSS and then it happened right in their browser haha

PR #2037, opened 2024-08-03
New user account validations

### LINKED ISSUE #1315: [FEAT]: New User account - lowercase and spacing

Continuation of #1312

**Apply to _new_ user accounts only**
- Must be lowercase on input
- Must contain no spaces

do not enforce this check on login however to prevent breaking previously existing user accounts.

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1315 

### What is in this change?

- On new user creation we now validate for usernames that are lowercase and have no spaces
- Modified `NewUserModal`, `EditUserModal`, and `AccountModal` components for these validations
- Frontend validation and `User` prisma model are updated for these validations
- We do not validate the update function in the `User` model to prevent forcing users to have to change usernames when updating just their password
- Still allow login with usernames that are not all lowercase and have spaces to prevent current users from getting locked out of their accounts

### Developer Validations


frontend/src/components/UserMenu/AccountModal/index.jsx
```diff
@@ -34,6 +37,11 @@ export default function AccountModal({ user, hideModal }) {
     setPfp(null);
   };
 
+  const handleUsernameChange = (e) => {
+    const value = e.target.value.toLowerCase().replace(/\s/g, '');
```
timothycarambat, line 41: You can use the `pattern` attribute on inputs to do this sanitation on the form-level and it will block submission.

frontend/src/components/UserMenu/AccountModal/index.jsx
```diff
@@ -143,10 +155,14 @@ export default function AccountModal({ user, hideModal }) {
               </label>
               <input
                 name="password"
-                type="password"
+                type="text"
```
timothycarambat, line 158: does it make sense to make this text visible as a password input?

frontend/src/pages/Admin/Users/UserRow/EditUserModal/index.jsx
```diff
@@ -53,10 +59,14 @@ export default function EditUserModal({ currentUser, user, closeModal }) {
                   className="bg-zinc-900 placeholder:text-white/20 border-gray-500 text-white text-sm rounded-lg focus:ring-blue-500 focus:border-blue-500 block w-full p-2.5"
                   placeholder="User's username"
                   minLength={2}
-                  defaultValue={user.username}
+                  value={username}
```
timothycarambat, line 62: If we edit only a users password - does the new validation block a previously working username from being updated?

timothycarambat (changes requested):
Leverage HTML form sanitization where possible, validate this does not block existing usernames from forcing to be changed.

Do these rules also apply to API endpoints for users?

frontend/src/components/UserMenu/AccountModal/index.jsx
shatfield4: For this one, yes. We do not have a verify password field and I changed it to text to match all other password setting fields.

shatfield4:
Updated backend validations to only validate the username with all lowercase and no spaces when calling the `User.create()` db model function.

timothycarambat:
> [!NOTE]
> We are aware you can create a valid username and _edit_ it to an invalid name. This is simply a convenience function around the username creation to enable backward compatibility for now with minimal overhead

PR #2106, opened 2024-08-13
fix(Dockerfile): remove hardcoded exposed port

kabaluyot:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

Exposed hardcoded port are causing "overriding" problems in Docker compose labels. In our use case, we are using traefik labels but the exposed port keeps overriding.

### What is in this change?

Remove exposed port. It might also be improper since we can override th SERVER_PORT making the expose port not usable

### Developer Validations

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -169,9 +169,6 @@ USER anythingllm
 ENV NODE_ENV=production
 ENV ANYTHING_LLM_RUNTIME=docker
 
-# Expose the server port
-EXPOSE 3001
-
 # Setup the healthcheck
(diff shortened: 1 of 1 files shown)


timothycarambat:
Great point, that line has probably been in there forever and there is no need to `EXPOSE` any port when we direct people to boot the container with `-p xxxx:xxxx` anyway 👍

kabaluyot:
Shout out to @shatfield4 :D

PR #2158, opened 2024-08-21
Api session id support

timothycarambat:
Enable the developer API to use `sessionId` in the body to logically partition chats without needing to create a user or pass in the entire chat history like you would for the OpenAi compatible API endpoints.

If not provided, the API chats can appear in the window in single-user mode. In multi-user mode these chats will not appear in the UI. When a sessionId is provided it is basically a pseudo-user ID we can limit search history by when doing API calls.


saroj-pattnaik:
Tim, your Friday morning update was amazing! I’m thrilled to start this now. This will undoubtedly enhance the developer experience, as developers prefer not to leave their IDE or copy code from another environment.

Thank you so much for creating such a superb product, AnythingLLM.

PR #2466, opened 2024-10-14
Update OpenAI TTS config to allow a custom BaseURL allowing for any TTS engine with a compatible API

phazei:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #xxx

### What is in this change?

This adds an input field for a custom BaseURL for the OpenAI TTS endpoint when selecting OpenAI as a TTS provider.

Since OpenAI has become an industry standard many tools have taken to mirroring their API for ease of integration.

With this change it allows for a much richer choice of options to use as a TTS provider.

I tested it using [AllTalk_TTS's OpenAI speech endpoint](https://github.com/erew123/alltalk_tts/wiki/API-%E2%80%90-OpenAI-V1-Speech-Compatible-Endpoint) and it works wonderfully.

### Developer Validations

     couldn't find any relevant docs, but it is notated in the env comments
     
- [ x] I have tested my code functionality

- [ x] Docker build succeeds locally
    In order to debug this, I needed to get the frontend/server/collector all working since the docker takes a VERY long time on the production-build  3/3 chown step.  There has got to be a way to speed that up.  Also, when running the servers independently, it doesn't save the config in any .env file so with every restart I had to set it up again, don't know what I missed there.  I was also running it under WSL ubuntu, so that was it's own pita, my dev machine is a MBP, but only my PC can run the models :/

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -214,6 +214,7 @@ GID='1000'
 # TTS_PROVIDER="openai"
 # TTS_OPEN_AI_KEY=sk-example
 # TTS_OPEN_AI_VOICE_MODEL=nova
+# TTS_OPEN_AI_ENDPOINT="https://api.openai.com/v1" #optional for openAI compatible endpoints
 
 # TTS_PROVIDER="elevenlabs"
 # TTS_ELEVEN_LABS_KEY=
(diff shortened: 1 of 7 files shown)


timothycarambat:
Moved to #2487 (youll still have PR credit)

PR #2578, opened 2024-11-03
Added vertex

### LINKED ISSUE #2575: [FEAT]: Application Default Credentials on Gemini

The vast majority of google libraries support ADC for transparent auth.  That also means they just "work" when deployed as applications on GKE using Workload Identity.   

Not needing to create secrets which then require storage, sharing between devs, and rotation is a huge win.

I did have a quick look at the code but it looks like the app is using `@google/generative-ai` which doesn't support ADC (no clue why - the hopscotch of google libraries always confuses me): https://github.com/google-gemini/generative-ai-js/issues/289.

The [@google-cloud/vertexai](https://www.npmjs.com/package/@google-cloud/vertexai) does, so the app could pivot to that.  That's the library I used to use (although that would then break existing users as that doesn't support api keys).

Maybe it'd need to be a new API integration (`Gemini: ADC`) 🤷

Stono:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2575 

### What is in this change?

This adds support for Vertex, which allows enterprises to run gemini on vertex. 

This is better because:

- It suppose service accounts, transparent auth with application default credentials
- Better privacy, the results aren't used to train the model
- Allows you to specify the region, which is important for those who want to keep their data within certain localities

### Additional Information

None

### Developer Validations


timothycarambat:
Closing as stale!

Issue #2262, opened 2024-09-11
[FEAT]: Name threads with AI generated summary text


glinkot:
### What would you like to see?

Right now I just see conversations shown as 'thread' and 'new thread'.  Other similar tools I've used use a lighter/cheaper LLM to generate a brief name for the chat, like "asking about dog breeds" or whatever.  This request was asking for it I believe (https://github.com/Mintplex-Labs/anything-llm/issues/701) but what was implemented was different to what I'm referring to.

timothycarambat:
Yes, and the primary reason is latency. Not everyone is using a high (or even mid) tier performance device and the latency to a load the model, run inference, and store the text, with a high speed. Most would not be willing to wait an additional 1-2s for inference completion just to name a thread, which is why the implementation was different.

So as it stands now, which is naming a thread based on the first chat, this is both fast for any device and LLM implementation but gets the idea of delineating chat threads without the user needing to explicitly rename them. Add an LLM into that process for what amounts to a very minor UX improvement was considered extraneous and would hurt the UX on lower-end systems using local models.

PR #2701, opened 2024-11-22
Use workspace LLM to create thread name

### LINKED ISSUE #2699: [FEAT]: Thread titles should be descriptive instead of showing the start of the first sentence

Currently when a new thread is created, the title on the left panel is the beginning of the first sentence. It would be useful if the LLM chosen for the workspace is used to generate a more descriptive title, similar to how the major platforms handle this.

A simple method would be to use an `{action} {thing}` approach.

For example, if I was asking how to build a brick wall, I might write "How do I build a brick wall?" in a thread.

If this was converted to `{action} {thing}` it would be:

`{Building} {a brick wall}`

And the prompt to the LLM could be.

`Based on the user's initial input, concisely create a title for this conversation in the format of: {action} {thing}. For example 'building' (action) 'a brick wall' (thing)`

scottybo:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2699

### What is in this change?

Uses the workspace's chosen LLM to rename the thread title to something more descriptive, rather than the first sentence of the conversation.

### Developer Validations

### DIFF
--- a/server/models/workspaceThread.js
+++ b/server/models/workspaceThread.js
@@ -140,7 +140,7 @@ const WorkspaceThread = {
         "You are a helpful AI that generates short, descriptive titles for conversations. Keep titles under 23 characters.";
       const result = await ApiChatHandler.chatSync({
         workspace,
-        message: `Based on the user's initial input, concisely create a title for this conversation in the format of: {action} {thing}. For example 'building' (action) 'a brick wall' (thing). Return only text a-z, 0-9 and spaces. This is the user's initial input: ${newName}`,
+        message: `Based on the user's initial input, concisely create a title for this conversation in the format of: {action} {thing}. For example 'Building' (action) 'a brick wall' (thing). Return only text a-z, 0-9 and spaces. This is the user's initial input: ${newName}`,
         mode: "chat",
         systemPrompt,
       });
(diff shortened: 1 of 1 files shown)


timothycarambat:
https://github.com/Mintplex-Labs/anything-llm/issues/2262#issuecomment-2344463292

The above comment is the primary reason why we dont support this. We offer an application on many various devices and not all of them are capable of concurrent generation or have the resources to do so - which is why the naming is static now since the latency could be high for low-end machines and the UX is not worth it considering that use-case.

Perhaps this could be something that could be done via a system toggle for appearance? This PR in its current form though does not enable that and should not make use of the `APIChatHandler` as while that works, is a different system that shouldn't be used for text generation in the backend

scottybo:
Ah - which handler should I be using? Still getting to grips with the code base!

scottybo:
@timothycarambat I know you're busy, but just wondering if you can give me some insight into which system should be used instead of `APIChatHandler`?

Also, where would you suggest putting this setting? I'm thinking that each it could be at the workspace level, under Chat Settings, something like this:

image

PR #2808, opened 2024-12-11
Redirect to default thread on thread deletion

### LINKED ISSUE #2767: [FEAT]: Redirect to the default Thread after current Thread is deleted
When I create a new Thread and then I delete the created Thread, the chat history remains open

**Proposed feature:**
After a Thread is deleted redirect back to the Default Thread since the Default Thread always will exist

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2767 

### What is in this change?

- Fixes UI/UX issue where on thread deletion, user would still be presented with the page of the deleted thread and sending messages here would throw a 404 error since the thread no longer exists
- Fixed by redirecting user back to default thread on thread deletion

### Developer Validations


timothycarambat (changes requested):
This does not handle the case where the user does a `bulkDelete` and are on a selected threadID which is is going to be removed.

This does work when clicking the more dots and deleting one-by-one

timothycarambat (approved):

andydev404:
The only issue that I see with the fix is that now the toast is not gonna work as expected due that you're immediately redirecting to another page.

timothycarambat:
That's acceptable since it would only be the case when you deleted the page you are on, which would then show all current threads indicating the one you were on was deleted without showing a message.

This also works because of how thread handling is done downstream on desktop

PR #2812, opened 2024-12-12
Add vector search API endpoint

### LINKED ISSUE #2801: [DOCS]: /v1/openai/embeddings has incorrect example body
### Description

In the Swagger docs for the `/v1/openai/embeddings` it gives an example value for the input as:

```json
{
  "input": [
    "This is my first string to embed",
    "This is my second string to embed"
  ],
  "model": null
}
```

However, the `input` should be plural `inputs` otherwise an error gets thrown. I.e.

```json
{
  "inputs": [
    "This is my first string to embed",
    "This is my second string to embed"
  ],
  "model": "text-embedding-3-small"
}
```

In addition, it should be made clear that in order for this endpoint to work with the correct embedding engine (rather than the default NativeEmbedder), the environment variable `EMBEDDING_ENGINE` should be set (e.g. to `openai`) otherwise a 500 error gets thrown.

It would also be useful to give an example of the model, e.g. `text-embedding-3-small`.

Just for reference if anyone else comes across this issue, the options for the `EMBEDDING_ENGINE` are

- openai
- azure
- localai
- ollama
- native
- lmstudio
- cohere
- voyageai
- litellm
- mistral
- generic-openai

And they can be seen in `function getEmbeddingEngineSelection` in `server/utils/helpers/index.js`

scottybo:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2801

### What is in this change?

Adds an API endpoint so that a vector search can be performed.

### Developer Validations

### DIFF
--- a/server/endpoints/api/workspace/index.js
+++ b/server/endpoints/api/workspace/index.js
@@ -841,6 +841,109 @@ function apiWorkspaceEndpoints(app) {
       }
     }
   );
+
+  app.post(
+    "/v1/workspace/:slug/vector-search",
+    [validApiKey],
+    async (request, response) => {
+      /*
+    #swagger.tags = ['Workspaces']
+    #swagger.description = 'Perform a vector similarity search in a workspace'
+    #swagger.parameters['slug'] = {
+        in: 'path',
+        description: 'Unique slug of workspace to search in',
+        required: true,
+        type: 'string'
+    }
+    #swagger.requestBody = {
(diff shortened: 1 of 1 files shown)


timothycarambat:
Moved to https://github.com/Mintplex-Labs/anything-llm/pull/2815#issuecomment-2539699643

PR #2818, opened 2024-12-12
Fix(2753): Replace React Router Navigate with window.location for SSO state refresh

### LINKED ISSUE #2753: [BUG]: when logging in to anythingllm using SSO as default user I see the create workspace button available.

Docker (remote machine)

### What happened?

I login to anythingllm instance using SSO auth token as default user of a workspace and can see create workspace, settings edit buttons enabled . The same are disabled when you login through login screen but visible only when logging in using SSO. The actual functionality is that these options should not be visible to default user.

Screenshot 2024-12-02 at 10 39 46 AM

Go to <your-anythingllm-instance/api/docs>
try the api <v1/users/<userid>/issue-auth-token> (give userid of user with default role)
use the token generated as response to the api call with your anythingllm instance url like : <anythingllm-url>/<sso-token>
you will be logged in to anythingllm instance as userid but can see the create workspace and other things which usually should be disabled for default role.

andydev404:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2753 

### What is in this change?

Fixed an issue where the SSO redirect wasn’t properly refreshing the application state after successful authentication. Replaced the React Router Navigate component with window.location.replace() to ensure a full page reload and proper state initialization.

### Developer Validations


timothycarambat:
How can i replicate this bug? I have not had issues passthrough on successful auth with simple SSO.

Is this a browser behavior, I really just test in chrome and there may be some bug there that is browser related

timothycarambat:
Beautiful! Thank you so much for handling this 🙌

PR #2819, opened 2024-12-12
Purge cached docs and remove docs from all workspaces on vectorDB/embedder changes

### LINKED ISSUE #2745: [BUG/FEAT]: Delete documents under “My Documents” after changing the embedding model
Hi,

I don't understand how the workspaces and the functions provided there are implemented.

**Question #1:**
On the left, under “My Documents”, I can create folders and assign names to them.
In my case, I created the folders “Testfolder” and “Testfolder2” in addition to the already existing “custom-documents” folder. However, despite multiple attempts, I have not been able to upload the documents I uploaded via the “Click to upload or drag and drop” area into these subfolders. They always end up in the “custom-documents” directory and cannot be moved from there to one of the other two folders, “Testfolder” and “Testfolder2”, using drag and drop.
Where can I find more detailed instructions that will help me solve the above problem so that I can create folders with appropriate names under “My Documents” to structure the source documents and upload the desired source documents to them?

**Question #2:**
The actual workspace is on the right. To get one or more of the documents from “My Documents” there, the relevant document from “My Documents” must be selected and then “Move to Workspace” clicked. Vectorization can then be started by clicking on “Save and Embed”.

Before using it as a local RAG system, the appropriate combination of LLM model (LM Studio) and embedding model (Ollama) must be found, which means that a large number of tests have to be carried out. This also includes testing different embedding models (via Ollama).

**Problem:**
If you test a new embedding model, you have to reset the existing vector database with “Reset Vector Database” and repeat the embedding process with the new embedding model. In my opinion, it should be sufficient to delete all documents in the workspace on the right, then reselect the relevant documents on the left under “My Documents”, then click on “Move to Workspace” so that the documents move to the workspace on the right again to be vectorized by the new embedding model (to be tested) (Save and Embed).
Exactly this procedure, which is in line with expectations, does NOT work!

Because afterwards, a glance at “Max Context Snippets” and “Vector Count” in the relevant workspace shows that the values have not changed, although I deliberately always change the value of “Text Chunk Size” beforehand to test a different embedding model.
Only after I delete the relevant documents on the left under “My Documents”, upload them again, then move them to the right using “Move to workspace” and then click on “Save and Embed” is the vector database generated and saved by the new embedding model.

As I understand it, this should not be the case, but should work as follows: as soon as I delete the documents that have already been embedded on the right-hand side of the workspace, then click on these documents again on the left under “My Documents”, move them into the workspace and click on “save and embed”, this process should be carried out using the new embedding model and the data should be vectorized accordingly.

Thanks in advance and kind regards
Joomgallerytestit

anythingllm-workspace

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2745 

### What is in this change?

- When changing embedding models or vector databases, this causes mismatching of vector dimensions
- This change clears the vector db of all vectors and purges all cached documents to allow for re-embedding to improve the UX for users testing multiple vectorDBs and embedding models
- Improved `WarningModal` component to appear more destructive to the user and match styles of other modals around the app

### Developer Validations


frontend/src/models/workspace.js
```diff
@@ -238,6 +238,14 @@ const Workspace = {
...
       .catch(() => false);
   },
+  wipeAllVectorDbs: async function () {
+    return await fetch(`${API_BASE}/workspace/reset-all-vector-dbs`, {
+      method: "POST",
+      headers: baseHeaders(),
+    })
+      .then((res) => res.ok)
+      .catch(() => false);
+  },
```
timothycarambat, line 248: We should do this at the server level because this would not work if the provider is changed via API or if for some reason this request does not send.

Can we use a `postUpdate` hook in `updateENV` on the LLM_PROVIDER or embedding model key and check if the key changed and run this function for cleanup then?

server/endpoints/workspaces.js
```diff
@@ -328,6 +329,58 @@ function workspaceEndpoints(app) {
...
 
+  app.post(
+    "/workspace/reset-all-vector-dbs",
+    [validatedRequest, flexUserRoleValid([ROLES.admin, ROLES.manager])],
+    async (request, response) => {
+      try {
+        const user = await userFromSession(request, response);
+        const VectorDb = getVectorDbClass();
+
+        const workspaces = multiUserMode(response)
```
timothycarambat, line 340: We can make this some generic Util function instead of an endpoint. Even then, managers should not be able to call this since only admins can modify providers

embed
timothycarambat, line 1: cd into embed and `git pull origin main` so this diff stops showing in every PR

timothycarambat (changes requested):

PR #2965, opened 2025-01-12
2964 file type default

### LINKED ISSUE #2964: [FEAT]: Add setting to change default handling for unrecognized file types to '.txt'

Many files having unknown extensions can be correctly processed as text files; I'd like to add a configuration option which would change the default behavior when a file's type is unrecognized from "display an error message" to "treat as text".  (Longer-term, it would be handy if a file's type could be specified/overridden when it was imported, but that would take a lot more effort.)
Conceptually related to issue #2638

wprice-uh:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2964 

### What is in this change?

Added an experimental setting labeled 'File Type Default' which, when turned on, changes the default behavior when a file being imported can't be recognized, from displaying an error message to treating the file as if it were plain text.

### Additional Information

Happy to receive feedback and make changes as needed; I'm not well-versed in javascript or in the frameworks/libraries used here, so most of what I did was based on cloning the live-file-sync functionality.

### Developer Validations


c0mmand3rOpSec:
Thanks for this! Works great on my end, exactly what I was needing.

timothycarambat:
@wprice-uh 
>  I'm not well-versed in javascript or in the frameworks/libraries used here

I dont believe that for a second, this looks great haha.

There is a general improvement that I will make on behalf of the PR, but this is great work - will be reviewing soon

PR #3001, opened 2025-01-22
Display `thinking` in the prompt response - model agnostic

timothycarambat:
This updates the UI to include `thinking` responses in the UI when the model provides it. This targets mostly widely available models like `deepseek-r1` but your model thoughts may not be supported.

- Handles both historical, static, and thought streaming


timothycarambat:
Then that makes sense, the Generic OpenAI connector doesn't have any of the "bells and whistles" yet to support thinking models like we do for other dedicated providers.

This issue is closed #3109 but i think it is relevant to reopen

PR #3005, opened 2025-01-23
2749 ollama client auth token

ssbodapati:
### Pull Request Type

- [ x] ✨ feat

### What is in this change?

- Ollama auth token support - Added UI field in the LLM Preferences
- setting OLLAMA_AUTH_TOKEN through update_env , updated env key_map
- Ollama client constructor uses OLLAMA_AUTH_TOKEN

### Developer Validations


ssbodapati:
@timothycarambat, thank you for reviewing this, the workspace chat fails because there is a dependency conflict with current ollama-js version 5.0, it does not pass headers while creating a completion.

The latest version with header support is 5.10.0, attaching screenshots. Workspace chat  after upgrading works as expected.

Pushed the changes.

Ollama 5.0 not passing headers to constructor

Ollama 5.10 passing headers to constructor

timothycarambat:
Excellent catch, I had only tested on my local Ollama to make sure current integrations work without a header since that is the normal and majority use case. i hadn't done a full QA yet, so thanks for catching that.

PR #3023, opened 2025-01-24
Add Version to AzureOpenAI

### LINKED ISSUE #3021: Azure OpenAI o1 Model with AnythingLLM
I am trying to configure the o1 model with AnythingLLM, but I got this error

```sh
Model o1 is enabled only for api versions 2024-12-01-preview and later
```

I can't find option to configure the api versions/

https://github.com/Mintplex-Labs/anything-llm/blob/604e7c9218ec72fc2b10e946cc3cae8ecb7d97fe/server/utils/AiProviders/azureOpenAi/index.js#L19-L22

We need to modify that section of code to this
```js
this.openai = new OpenAIClient(
      process.env.AZURE_OPENAI_ENDPOINT,
      new AzureKeyCredential(process.env.AZURE_OPENAI_KEY),
      {
        apiVersion: "2024-01-01-preview",
      }
    );
```

timothycarambat:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3021

### Developer Validations

### DIFF
--- a/server/utils/AiProviders/azureOpenAi/index.js
+++ b/server/utils/AiProviders/azureOpenAi/index.js
@@ -16,9 +16,13 @@ class AzureOpenAiLLM {
     if (!process.env.AZURE_OPENAI_KEY)
       throw new Error("No Azure API key was set.");
 
+    this.apiVersion = "2024-12-01-preview";
     this.openai = new OpenAIClient(
       process.env.AZURE_OPENAI_ENDPOINT,
-      new AzureKeyCredential(process.env.AZURE_OPENAI_KEY)
+      new AzureKeyCredential(process.env.AZURE_OPENAI_KEY),
+      {
+        apiVersion: this.apiVersion,
+      }
     );
     this.model = modelPreference ?? process.env.OPEN_MODEL_PREF;
     this.limits = {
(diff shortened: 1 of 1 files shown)


MoustafaAMahmoud:
@timothycarambat o1 doesn't have temperature
```sh
Unsupported parameter: 'temperature' is not supported with this model.
```

image

timothycarambat:
@MoustafaAMahmoud The issue is there is _no way_ for us to know what model your endpoint runs since they can be named anything and Azure provides no transparency to the true model running behind the deployment.

In OpenAI we can check by the model tag, we will need to do something different here for supporting o1 in Azure. We dont have access to o1 on Azure so it makes testing very difficult.

ericthomas1:
Hello,

We're facing the same issue. Unable to use AnythingLLM for Azure `o3-mini` deployments. 

Can a "model type" be added in the LLM Provider area? 
- If an `o*` -type model is selected, the API call omits the `temperature` param?

timothycarambat:
@ericthomas1 The issue with that solution will be assuming someone names their deployment according to the model. We can do it, but I'm sure someone will name their deployment `mymodel` and it will really be o3 and still fail :/

ericthomas1:
Might be able to leave the Deployed Model Name as-is. 

Would something like this work?: 
image

timothycarambat:
@ericthomas1 That is good longer term fix. I will make a PR from that, but in the interim: https://github.com/Mintplex-Labs/anything-llm/commit/a5ee6121e22bb41d61104c03d96e063633b0dbf4 should set you up now assuming the deployment name starts with `o`

Build has already been pushed since so it is live in `latest`

ericthomas1:
Thats so awesome! How does one run the `latest` _build_? I just updated this morning to the latest _release_. Are these synonymous?

timothycarambat:
@ericthomas1 are you on Docker (I was assuming you are). If so, `docker pull mintplexlabs/anythingllm:latest`  would be the way

ericthomas1:
I’m on the standalone install until we get Azure auth figured out.

ericthomas1:
Hi @timothycarambat, is 1.7.4 expected soon?

PR #3045, opened 2025-01-28
Add embedding support for message.content which is string/object array type

hehua2008:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

None

### What is in this change?

Add embedding support for message.content which is string/object array type.

### Additional Information

https://platform.openai.com/docs/api-reference/chat/create
Screenshot2025-01-29 15 15 24
Screenshot2025-01-29 10 39 44

### Developer Validations

### DIFF
--- a/server/models/workspaceChats.js
+++ b/server/models/workspaceChats.js
@@ -11,10 +11,29 @@ const WorkspaceChats = {
     apiSessionId = null,
   }) {
     try {
+      let promptString;
+      if (typeof prompt === "string" || prompt instanceof String) {
+        promptString = prompt;
+      } else if (Array.isArray(prompt)) {
+        promptString = prompt
+          .map((x) => {
+            if (typeof x.text === "string" || x.text instanceof String) {
+              return x.text;
+            } else if (typeof x === "string" || x instanceof String) {
+              return x;
+            } else {
+              return JSON.stringify(x);
+            }
+          })
+          .join("\n\n");
+      } else {
+        promptString = JSON.stringify(prompt);
+      }
+
       const chat = await prisma.workspace_chats.create({
         data: {
           workspaceId,
-          prompt,
+          prompt: promptString,
           response: JSON.stringify(response),
           user_id: user?.id || null,
           thread_id: threadId,
--- a/server/utils/vectorDbProviders/astra/index.js
+++ b/server/utils/vectorDbProviders/astra/index.js
@@ -268,7 +268,26 @@ const AstraDB = {
       };
     }
 
-    const queryVector = await LLMConnector.embedTextInput(input);
+    let textInput = "";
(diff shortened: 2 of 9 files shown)


timothycarambat:
Blocking this PR until some more explanation is provided since there is no issue.

@hehua2008 What is the intended function of this PR since there is no issue and it touches a critical part of the code its pretty tantamount that the root cause is discussed here. How is it you are getting `Array` type data into the chat window and if so - how are you doing so?

Then we can have a larger discussion around why this is necessary or if there is some other pre-processing step we can apply here.

Thanks

hehua2008:
> Blocking this PR until some more explanation is provided since there is no issue.
> 
> @hehua2008 What is the intended function of this PR since there is no issue and it touches a critical part of the code its pretty tantamount that the root cause is discussed here. How is it you are getting `Array` type data into the chat window and if so - how are you doing so?
> 
> Then we can have a larger discussion around why this is necessary or if there is some other pre-processing step we can apply here.
> 
> Thanks

For write code with Cline or Roo Code:
```json
{
  "model": "mlx-python",
  "messages": [
    {
      "role": "system",
      "content": "You are Roo, a highly skilled software engineer with extensive knowledge in many programming languages, frameworks, design patterns, and best practices..."
    },
    {
      "role": "user",
      "content": "<task>\nWrite a MLX example in Python\n</task>"
    },
    {
      "role": "user",
      "content": "<environment_details>\n# VSCode Visible Files\nsrc/api/providers/anythingllm.ts\n\n</environment_details>"
    }
  ],
  "temperature": 0,
  "stream": true
}
```

timothycarambat:
@hehua2008 This seems like something that should be done for pre-proc in your fork though, no?

hehua2008:
> @hehua2008 This seems like something that should be done for pre-proc in your fork though, no? I can see why it is here in the repo, I just wonder if that is basically our responsibility or not is all.
> 
> That function modified is intended to take a single query string to be used for semantic search

Yes, it is AnythingLLM's responsibility.
Please see the OpenAI API documentation:
https://platform.openai.com/docs/api-reference/chat/create
Screenshot2025-01-29 10 39 44

hehua2008:
This commit may not be a good solution, but I want to bring it up here to draw your attention. It would be a great and valuable thing to make AnythingLLM support Cline or Roo-Code, etc. If you are interested in it, please help improve this commit or provide a better solution. Thank you very much!

hehua2008:
@timothycarambat So, what's your plan to support content in array type as described in https://platform.openai.com/docs/api-reference/chat/create ?

PR #3269, opened 2025-02-19
add stt output to input prompt area

### LINKED ISSUE #3268: [FEAT]: Speech to text confirmation before submission

Currently once we end STT session we immediately send the transcript as a prompt to the LLM provider. Sometimes its annoying when - 
1. STT recognises some words incorrectly but still send it as a prompt before giving me a chance to correct it.
2. I need to think while speaking but if the pause is too long (currently more than 3.2s) it submits the incomplete prompt. 

One way to achieve both of the above issues is if we can put the recognised text in prompt input area (already happening) but not send it till user clicks on send button or some equivalent action (min one extra click required from user). This can -
1. give me a chance to correct the incorrectly recognised words.
2. let me add to existing text (in between or at the end)

As far as I have searched, I found no existing feature request similar to this hence opening a new issue.

yashschandra:
### Pull Request Type

- [x] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3268 

### What is in this change?

Stopped sending voice command directly as prompt to LLM provider so that user can get a chance to edit or add to it if there was a long pause while speaking.

### Additional Information

Before - 

https://github.com/user-attachments/assets/f3504b6f-dfb5-49a6-aa88-8a93784a28c7

Now - 

https://github.com/user-attachments/assets/d3260303-7602-4567-b259-b82958244b6a

Additionally fixed a minor typo: `endTTSSession` -> `endSTTSession`

### Developer Validations


timothycarambat:
FWIW, it used to be this way, then people wanted it to autosubmit, now this PR would change it back

therealtimex:
Can we implement two modes:

-  A long press of the microphone icon activates "continuous" mode, which enables autosubmit. The UI should visually indicate that this mode is active. 
- A short press of the microphone icon triggers manual submit.

yashschandra:
> FWIW, it used to be this way, then people wanted it to autosubmit, now this PR would change it back
I get this, but I feel for a non-native english speaking (like myself) it may be useful

> Can we implement two modes:
> 
> A long press of the microphone icon activates "continuous" mode, which enables autosubmit. The UI should visually indicate that this mode is active.
> A short press of the microphone icon triggers manual submit.
I was thinking having this configuration in _Settings_ but this may also work. @timothycarambat any thoughts on this approach?

timothycarambat:
@therealtimex That kind of UX is ambiguous and is bound to be non-discoverable. Will have to make this a setting in the Voice and Speech area or elsewhere so that we can stop flip-flipping PRs every couple months on this.

therealtimex:
Agreed, setting is always a safe bet.

yashschandra:
added a checkbox for **Autosubmit** in _Voice and Speech_ settings section - 

https://github.com/user-attachments/assets/cd957fcc-532c-4f74-aa85-7bb2c5ffad34

yashschandra:
@timothycarambat wondering if this change is good enough to be added? will be very happy to get a feature like this along with all other features people are building here 😄

PR #3354, opened 2025-02-26
Windows development environment variables support

kbourro:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [x] 🔨 chore
- [ ] 📝 docs

### What is in this change?

yarn dev wasn't working on windows machines as we need a way to set cross environment variables. For this reason i have changed the packages with cross-env

### Developer Validations


collector/package.json
```diff
@@ -20,6 +20,7 @@
     "bcrypt": "^5.1.0",
     "body-parser": "^1.20.2",
     "cors": "^2.8.5",
+    "cross-env": "^7.0.3",
```
timothycarambat, line 23: This should be in `devDependencies` since it is not required for production.

frontend/package.json
```diff
@@ -15,16 +15,17 @@
     "@mintplex-labs/piper-tts-web": "^1.0.4",
     "@phosphor-icons/react": "^2.1.7",
     "@tremor/react": "^3.15.1",
+    "cross-env": "^7.0.3",
```
timothycarambat, line 18: Same here

server/package.json
```diff
@@ -47,6 +47,7 @@
     "chromadb": "^1.5.2",
     "cohere-ai": "^7.9.5",
     "cors": "^2.8.5",
+    "cross-env": "^7.0.3",
```
timothycarambat, line 50: and here

timothycarambat (changes requested):
`cross-env` should be moved into `devDependencies`, running in production should not leverage cross-env in this aspect.

PR #3416, opened 2025-03-07
Skip too large files to read

marcin-golebiowski:
### Pull Request Type

- [ ] ✨ feat
- [X] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

It's not possible to view documents if there is a file bigger than 512 MB in documents.

### What is in this change?

Skipping the reading the file when the size is bigger than 512 MB.

### Additional Information

By preemptively filtering out oversized files, we improve reliability and maintain system efficiency.
Oversized files can happen due to big PDFs with images.

### Developer Validations

### DIFF
--- a/server/utils/files/index.js
+++ b/server/utils/files/index.js
@@ -45,10 +45,15 @@ async function viewLocalFiles() {
       };
       const subfiles = fs.readdirSync(folderPath);
       const filenames = {};
+      const maxSize = 0x1fffffe8;
 
       for (const subfile of subfiles) {
         if (path.extname(subfile) !== ".json") continue;
         const filePath = path.join(folderPath, subfile);
+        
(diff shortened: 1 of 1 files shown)


timothycarambat:
So in this case, large files just become invisible? How are you ever supposed to be able to select them to embed afterwards?

marcin-golebiowski:
The other solution  would be to check the size of the file before fs.readFileSync.
In case of huge file the empty content could be returned instead of calling fs.readFileSync which will fail.

timothycarambat:
Moved to https://github.com/Mintplex-Labs/anything-llm/pull/3497

PR #3446, opened 2025-03-12
Add ability to reordering workspaces

### LINKED ISSUE #1438: [FEAT]: Manual ordering of workspace

Allows administrator and / or users to manually order the location of workspaces

MrMarans:
### Pull Request Type

- [X] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

No issue created
resolves #1438

### What is in this change?

This change add the ability to reorder workspaces when logged in as admin or manager. 

### Additional Information

With the usage of the react-beautiful-dnd I created code to make workspaces drag and dropable. This creates changes within the DB for workspaces 

image

And it did work for me with old workspaces. The update made no change to the current ordering and workspaces were usable as much as before.

### Developer Validations

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -29,6 +29,7 @@
     "onnxruntime-web": "^1.18.0",
     "pluralize": "^8.0.0",
     "react": "^18.2.0",
+    "react-beautiful-dnd": "^13.1.1",
     "react-device-detect": "^2.2.2",
     "react-dom": "^18.2.0",
     "react-dropzone": "^14.2.3",
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -29,7 +29,7 @@
     "onnxruntime-web": "^1.18.0",
     "pluralize": "^8.0.0",
     "react": "^18.2.0",
-    "react-beautiful-dnd": "^13.1.1",
+    "react-beautiful-dnd": "13.1.1",
     "react-device-detect": "^2.2.2",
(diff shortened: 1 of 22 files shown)


shatfield4 (approved):
LGTM, tested and works very well. This improves the UX a lot for users with large amounts of workspaces.

My only concern here is the size of the `react-beautiful-dnd` package and all it's dependencies. This may bloat the main application but IMO it is worth the tradeoff here.

timothycarambat:
This is so dope @MrMarans - thank you for handling this

MrMarans:
@timothycarambat thanks. Probably more pull requests coming in the future since I want to use anythingLLM in production at my company and there are a few features still missing for me.

timothycarambat:
@MrMarans Made some changes to move this to the client-side for ordering. The reason for this change is that in a multi-user mode where a workspace is commonly accessed by multiple sub-users the changing of ordering by one user will change the order to all users on reload.

While the tradeoff is portability between order on multiple browsers, this simplifies the logic greatly and works with all user modes. An alternative is to use a `systemSetting` key in single-user and then add a field on like you did on workspaces, but on the `user` record. Then it would be an atomic setting per-user and be cross device.

Given this is a nice QoL change, I figured the local solution is easier to maintain and does the job.

PR #3483, opened 2025-03-17
Added chat mode to only use agent

MrMarans:
### Pull Request Type

- [X] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

No issue created

### What is in this change?

Adds Chat Mode "Agent"
image
Every first message when loading the chat window adds the @agent command. 
Also the agent mode doesnt have a timeout anymore when in this mode. 

### Additional Information

No agent timeout when in agent chat mode of workspace, only agent calls. Added language parts for translations (and added german translation)

I would love to edit the documentation but I currently have no clue where the files for the documentation is. When answered I'll add it to my branch

### Developer Validations


deathwish:
@MrMarans the documentation appears to be in the  [Mintplex-Labs/anythingllm-docs](https://github.com/Mintplex-Labs/anythingllm-docs) repository, looks like you'd need another PR over there which references this one.

@timothycarambat Any progress on getting this reviewed? I would really like to have this feature.

MrMarans:
I just updated the pull request to work with version 1.8.2 that got released yesterday. 
Changes in the past months broke my original modification.

I also made users be able to /reset when the workspace is in agent mode, and removed the /exit command. Wasnt able to use custom slash commands tho.

jezweb:
I still dont really understand why agent mode isnt engaged all the time, is it to stop people from using tools in a chat instance?

MrMarans:
> I still dont really understand why agent mode isnt engaged all the time, is it to stop people from using tools in a chat instance?

Because the agent mode doesn't use rag. It can only load and search files, but not give partial data (as far as I know)

timothycarambat:
> I still dont really understand why agent mode isnt engaged all the time, is it to stop people from using tools in a chat instance?

Because with lots of models for local it leads to a waste of a ton of tokens and smaller models do very very badly here. So you get by far worse results, which is why it is own action (and with this PR you can have always on agents too)

jezweb:
Could there be a way of using rules written into the system prompt that would engage agent mode? Ie if the user says abc-123 then switch to agent mode?

timothycarambat:
> Could there be a way of using rules written into the system prompt that would engage agent mode? Ie if the user says abc-123 then switch to agent mode?

That is what `@agent` does. If you are asking about inferring agent calling based on the semantic meaning of the user query, that is what the LLM _should_ do with tool calling. We have our own universal tool calling so we don't have to wait for the user to update the chat template, but the tradeoff is that the model may try to call a tool anyway even on a simple greeting

csshenouda:
@timothycarambat: you mentioned on #4569 that there are good reasons this isn't merged yet. I understand the limitations if users would accidentally enable agent mode just anywhere. But it seems enabling it per workspace as @MrMarans suggested makes a lot of sense. 
In my case for example I have a workspace dedicated only to legal research which I have an MCP-server for. So whenever I'm in that workspace I only care about agentic mode and nothing else. 

Or did you mean there are technical/architectural reasons for not merging this? Maybe add it as an "experimental feature" like live document sync and mobile?
Glad to hear your thoughts.

tgkap:
> > I still dont really understand why agent mode isnt engaged all the time, is it to stop people from using tools in a chat instance?
> 
> Because the agent mode doesn't use rag. It can only load and search files, but not give partial data (as far as I know)

How would agent mode not be using RAG?

I have done tests with queries on my documents and prepending @agent before a query tends to give me better answers. It has to be getting its answers from the embedded vector database used in the RAG, correct?

PR #3501, opened 2025-03-19
RamaLama works as a server for this

ericcurtin:
RamaLama is a project that makes inferencing boring, RamaLama enables one to use engines like llama-server from llama.cpp or vLLM with just a few steps amongst other features.

Tested AnythingLLM+RamaLama as I wanted to try out RAG, worked great. Ran deepseek-r1:14b on RamaLama for testing.

 ### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [x ] 📝 docs

### Relevant Issues

resolves #xxx

### Developer Validations

- [ x] I ran `yarn lint` from the root of the repo & committed changes
- [x ] I have tested my code functionality
- [ x] Docker build succeeds locally


timothycarambat:
That list in the readme is for explicitly supports providers - you probably were able to use your provider via the Generic OpenAI connector (which is great!) but that README list is reserved for direct integrations. 

We have generic OpenAI compatible services listed on there - which covers most other providers

PR #3599, opened 2025-04-05
Allow User to Change UID/GID Values at Container Runtime

nraboy:
In the current version, the documentation advertises changing the UID and GID values in the `.env` file, however it doesn't appear it actually does anything when inspecting the codebase.

The current codebase allows you to set the UID and GID values when building the image, which isn't particularly useful when deploying containers.

My proposed changes allow for optionally changing the UID and GID values at runtime. To make this possible, the entire `Dockerfile` is ran as the `root` user. In the `docker-entrypoint.sh` file, which is ran at container runtime, if the UID or GID values are set in the environment variables, then they will be changed for the `anythingllm` user. The application is started as the `anythingllm` user through `gosu` rather than continuing as `root` user.

This proposed change is valuable to avoid permission issues on mapped volumes.

 ### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #xxx

There is no matching issue ticket. I had the issue and opted to fix it myself, so I didn't feel an issue ticket was necessary.

### What is in this change?

This change is strictly related to the Docker configuration. It allows users to deploy containers with specified UID and GID values as environment variables. Setting these values is optional because it will default to UID-1000 and GID-1000.

### Developer Validations

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -5,6 +5,11 @@ FROM ubuntu:jammy-20240627.1 AS base
 ARG ARG_UID=1000
 ARG ARG_GID=1000
 
+ENV UID=$ARG_UID
+ENV GID=$ARG_GID
+
+USER root
+
 FROM base AS build-arm64
 RUN echo "Preparing build of AnythingLLM image for arm64 architecture"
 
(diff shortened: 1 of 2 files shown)


timothycarambat:
This is already permitted during build
https://github.com/Mintplex-Labs/anything-llm/blob/8b6083c8044fcea9f3adc8152563399e7fc29645/docker/docker-compose.yml#L13-L15

The property is ARG_UID or ARG_GID, which when unset fallback to `1000`. You can set them in `args` in a `docker-compose` or via `[--build-arg`](https://docs.docker.com/reference/cli/docker/buildx/build/#build-arg) in `docker build` or via ENV provided the ENV is the one in the `docker` folder in the repo.

nraboy:
Yes I get that it is available during build. My changes were for runtime.

If a user wants to do a custom build every time they want to build something, great, but for the user who wants to get the official prebuilt image, this approach does not work.

Example, let's say I pull the image to my Synology NAS. Building on my NAS is likely not feasible. On my NAS I protect my container volumes by user which means the default UID and GID values are going to get me permission denied errors. Having the values set at runtime solves this.

This approach is used for many other popular projects like Plex, Syncthing, etc.

I'm hoping you can reconsider or have a discussion about these two different approaches.

Thanks,

PR #3620, opened 2025-04-09
handling of citations in openRouter provider #3581

### LINKED ISSUE #3581: [FEAT]: showing citations when using Perplexity through OpenRouter

AnythingLLM already shows citations when Perplexity is used through its own api #2827 . When accessing Perplexity through OpenRouter the citations are not shown.

I checked that OpenRouter does return these citations:
```python
import os

from openai import OpenAI


def main():
    client = OpenAI(
        base_url="https://openrouter.ai/api/v1/",
        api_key=os.getenv("OPENROUTER_API_KEY"),
    )
    response = client.chat.completions.create(
        model="perplexity/sonar",
        messages=[
            {
                "role": "user",
                "content": [
                    {"type": "text", "text": "battle of waterloo"},
                ],
            },
        ],
    )

    print(response.citations)


if __name__ == "__main__":
    main()
```

rudimichal:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3581 

### What is in this change?

If OpenRouter returns citations they are used to enrich tokens with urls (just like it's already implemented in Perplexity provider). Additionally, all citations are appended to model response.

Screenshot 2025-04-09 at 11 40 55 AM

### Developer Validations


timothycarambat:
Modified the PR since the code paths were assuming that the OR model would always be perplexity, and we should short-circuit any PPLX logic since most models on OR are not PPLX models.

- Modified the enrichToken to be more generic so we can extend it in the future if any other handling cases arise
- Removed text added to the end of the response of all the links. We don't do this in the native perplexity provider, and they should function the same. That being said, I think this can be extended into `sources` so they can be seen in the response like what you did have previously

Tested:
- PPLX reasoning and non-reasoning in-line citations
- Reasoning only models like Deepseek
- Normal models like Llama 4 Scout with no reasoning or citations

PR #3648, opened 2025-04-14
TTS auto-speak and STT auto-submit

### LINKED ISSUE #2237: [FEAT]: Voice & Speech Autoplay

There should be a flag for "auto playback" to play the audio stream / file when it's completed
More settings for elevenlabs: selectable model

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2237 
closes #3269
depends on #3678

### What is in this change?

- Add option for auto-submit on STT in chat window
- Add option for auto-speaking response messages
- Add toggles under appearance settings
- Add toggles on account modal for multiuser accounts

### Developer Validations


frontend/src/components/UserMenu/AccountModal/index.jsx
```diff
@@ -178,9 +180,11 @@ export default function AccountModal({ user, hideModal }) {
                   defaultValue={user.bio}
                 />
               </div>
-              <div className="flex flex-row gap-x-8">
+              <div className="grid grid-cols-2 gap-6">
                 <ThemePreference />
                 <LanguagePreference />
+                <AutoSubmitPreference />
+                <AutoSpeakPreference />
```
timothycarambat, line 187: Why grid here when we use flex everywhere else in the app?

frontend/src/App.jsx
```diff
@@ -93,174 +98,196 @@ export default function App() {
...
             <PfpProvider>
               <I18nextProvider i18n={i18n}>
-                <Routes>
-                  <Route path="/" element={<PrivateRoute Component={Main} />} />
-                  <Route path="/login" element={<Login />} />
-                  <Route
-                    path="/sso/simple"
-                    element={<SimpleSSOPassthrough />}
-                  />
+                <TTSProvider>
```
timothycarambat, line 101: Why is the entire application wrapped in the TTSProvider when only the chat container layout would be the only time these hooks are utilized? Otherwise we are wrapping every page load in this context and hardly using it

Is there a benefit here to wrapping it over every route as opposed to just the layouts that may require it?

frontend/src/models/appearance.js
```diff
@@ -1,10 +1,10 @@
 import { APPEARANCE_SETTINGS } from "@/utils/constants";
 
 const Appearance = {
-  defaultSettings: { showScrollbar: false },
+  defaultSettings: { showScrollbar: false, autoSubmit: true, autoSpeak: true },
```
timothycarambat, line 4: Defaults should _not_ result in spoken responses. should an existing user have a setup we should not automatically assume they want this, since the current UX is:
- Speak -> autosubmits
- Click speaker icon to render & hear response

This PR should add the functionality but we should not modify the UX for TTS/STT as it is right now so that current behaviors don't result in new outputs unexpectedly.

`autospeak` should be false, at least for now.

frontend/src/components/WorkspaceChat/ChatContainer/index.jsx
```diff
@@ -262,6 +269,23 @@ export default function ChatContainer({ workspace, knownHistory = [] }) {
...
+    if (!lastMessage) return;
+    if (lastMessage.role !== "assistant") return;
+    if (!lastMessage.content || lastMessage.content === "") return;
+    if (!lastMessage.chatId) return;
+    if (lastMessage.uuid === lastMessageRef.current) return;
+
+    lastMessageRef.current = lastMessage.uuid;
+    speak(lastMessage.content, lastMessage.chatId, workspace.slug);
+  }, [chatHistory, autoSpeak]);
+
```
timothycarambat, line 288: This feels like it needs a refactor - lots of logic in this hook that can be passed to a function where the only qualification for the function to then possibly execute speak is checking if autoSpeak is enabled

frontend/src/contexts/TTSProvider.jsx
```diff
@@ -0,0 +1,155 @@
...
+        isLoading,
+        togglePlay,
+        stop,
+        provider,
+        settings,
+        speak,
+      }}
+    >
+      {children}
+      <audio ref={audioRef} hidden={true} />
```
timothycarambat, line 144: In this provider, if we have a list of AI replies as historical messages, is the following what occurs?

- We play AI message 1
   - TTS provider generates TTS and attaches to audio element here
- We play AI message 2
  - TTS provider generates TTS and attaches to audio element here
- We click to replay AI message 1
   - TTS provider generates TTS and attaches to audio element here(?) - duplicating the TTS inference for an already generated message?
   
Currently, already rendered TTS responses are stored in their blobs and audio elements and until the page is reloaded, the TTS reply is not regenerated _per click_. If we overload a single `audio` to do this, there is no current way in this hook to keep track of the blobs already generated?

timothycarambat (changes requested):

Issue #3626, opened 2025-04-10
[FEAT]: Allow custom headers for the upload-link endpoint


agorman:
### What would you like to see?

When using the `/v1/document/upload-link` endpoint I'd like to be able to add my own custom headers. This would allow me to include API keys for links that require authorization.

timothycarambat:
How would we then pass these onto whatever the destination link would be? We run puppeteer on that part - so I presume we would append them before opening the page?

agorman:
I'm not sure I follow. I'm running this in docker aren't you just using an http client in the backend to pass the call?

timothycarambat:
No, we use pupteer and fallback to regular HTTP fetch if that fails, so that JS-enabled pages are rendered properly and can be collected once the content is fully rendered.
https://github.com/Mintplex-Labs/anything-llm/blob/master/collector/processLink/convert/generic.js#L72-L122

agorman:
Thank you for the explanation and the link. That makes sense now. Would something like this work? https://pptr.dev/api/puppeteer.page.setextrahttpheaders

timothycarambat:
@agorman Absolutely, that would be the main idea here. However, I think passing auth could be tricky here since often that can be site-specific. So the issue becomes pretty wide in impact.

For example, consider basic JWT `Authentication Bearer` headers. That same header is used across many sites, but the token changes per site. So there would need to be an additional way to manage this per site. Obviously some sites use multiple headers for other forms of auth or fingerprinting

owlowluio:
I may have found a solution to your problem. You can look here [click](https://javierlopez.eu/captha/) Hope this helped

agorman:
@timothycarambat maybe I'm misunderstanding. If I pass a link and headers to `/v1/document/upload-link` wouldn't that be a single set of headers to a single link? So in that case the headers I send would always be site specific based on the link.

Maybe I'm taking the wrong approach. What I'm trying to do is feed it information from an internal HTTP API. I ran into a couple road blocks right away. Not being able to auth and https://github.com/Mintplex-Labs/anything-llm/issues/3625. Is there a better way to accomplish what I'm trying to do? Or maybe it's just out of scope?

timothycarambat:
> If I pass a link and headers to /v1/document/upload-link wouldn't that be a single set of headers to a single link? So in that case the headers I send would always be site specific based on the link.

[...]

PR #3652, opened 2025-04-15
Enable bypass of ip limitations via ENV in collector processing

### LINKED ISSUE #3625: [BUG]: Can't use upload-link endpoint on local network IPs

Docker (local)

### What happened?

When using the `/v1/document/upload-link` endpoint if I use a `link` with a local IP on my network I get the following error.

~~~
{
	"success": false,
	"error": "Not a valid URL.",
	"documents": []
}
~~~

Links sent to routable IPs or domain names seem to work `https://google.com` or `https://8.8.8.8`. However something like `https://192.168.1.100` doesn't work. The error is also returned right away which seems like there is some validation check failing. I can curl the endpoints from the docker container so I know routing isn't an issue.

_No response_

timothycarambat:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3625
connect #3626

### What is in this change?
- Enable bypassing for simple IP checks on web-scraping via a collector ENV for those who want to do this
- Enabled by default
- Set `COLLECTOR_ALLOW_ANY_IP` to any value in your `env` to enable this bypass.

### Developer Validations

- [x][ Relevant documentation has been updated](https://docs.anythingllm.com/configuration#local-ip-address-scraping)


shatfield4 (approved):
LGTM

DoiDoi888 (commented):
我

PR #3795, opened 2025-05-09
Add Disable Spellcheck Option

### LINKED ISSUE #3605: [BUG]: Disable spell checker by default

AnythingLLM desktop app

### What happened?

Please disable the spell checker, it is very annoying for mixed language inputs. I also has very little utility since, LLMs are good at understanding typos anyway.

Image

_No response_

17ColinMiPerry:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3605 

### What is in this change?

This feature allows the option to disable the spellcheck feature (on by default). This will allow users working in multiple languages (or languages other than their system primary language) to avoid issues with the spellchecker.

image

### Additional Information

Uses the useTranslation hook to support future localization. Locales files have been updated to support future localization.

### Developer Validations


timothycarambat:
The only change here worth noting was reducing the complexity of `Appearance.get` since you can just read it on mount since it only references LocalStorage - so it is fast and there is no need to track its state in the components it effects since the state will never change while in that view. `get` also assumes the fallback default value it not found in settings object in localstorage

PR #3820, opened 2025-05-14
Fix Loop bound injection via `preserveCamelCase`

odaysec:
### Pull Request Type

https://github.com/Mintplex-Labs/anything-llm/blob/adf7e8a9a7153690e0d66981425827c288b741cc/server/utils/helpers/camelcase.js#L25-L25

fix the issue need to validate the `string` parameter in the `preserveCamelCase` function to ensure it is a legitimate string with a reasonable length. This can be achieved by:
1. Adding a check at the beginning of the `preserveCamelCase` function to ensure `string` is a string and its length is within acceptable bounds.
2. If the `string` parameter fails validation, we should throw an error or return a default value to prevent further processing.

---

Using the `.length` property of an untrusted object as a loop bound may cause indefinite looping since a malicious attacker can set the `.length` property to a very large number, when a program that expects an array is passed a JSON object such as `{length: 1e100}`, the loop will be run for 10100 iterations. This may cause the program to hang or run out of memory, which can be used to mount a denial-of-service (DoS) attack.

## Recommendation
Either check that the object is indeed an array or limit the size of the `.length` property.

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Developer Validations

### DIFF
--- a/server/utils/helpers/camelcase.js
+++ b/server/utils/helpers/camelcase.js
@@ -17,6 +17,10 @@ const preserveCamelCase = (
   toUpperCase,
   preserveConsecutiveUppercase
 ) => {
+  if (typeof string !== "string" || string.length > 1000) {
+    throw new Error("Invalid input: `string` must be a string with a length of 1000 characters or less.");
+  }
+
   let isLastCharLower = false;
   let isLastCharUpper = false;
   let isLastLastCharUpper = false;


odaysec:
> [Weaviate vectodb](https://github.com/Mintplex-Labs/anything-llm/blob/076cb844fc0b13e2797dfd309cdb756bdca6dc66/server/utils/vectorDbProviders/weaviate/index.js)

Hi @timothycarambat, sorry for the delayed response, I’ve been tied up with other responsibilities recently.

Thanks for pointing out the origin of the helper in the NPM package and for clarifying the scope of usage in the `Weaviate` vector DB provider.

You're right — based on the current usage, the impact is somewhat limited due to the slug length restrictions imposed by browsers (`around ~2048 characters`). However, the motivation behind the patch was to preemptively address a case of unbounded iteration that could pose a problem in a less constrained context or under specific misuse patterns (internal system-to-system queries or malicious proxies that bypass typical browser constraints).

PR #3821, opened 2025-05-14
Fix Uncontrolled data used in path expression 

odaysec:
### Pull Request Type
https://github.com/Mintplex-Labs/anything-llm/blob/388960e79cfbe6f0726b6e80d8f0542415606f28/server/utils/files/index.js#L180-L190

fix the issue need to ensure that the `filename` input is properly sanitized and validated before constructing the file path. Specifically:
1. Use the `path.resolve` function to normalize the path and remove any `..` segments.
2. Verify that the resolved path starts with the `documentsPath` to ensure it is confined to the intended directory.
3. Throw an error or return early if the validation fails.

The `normalizePath` function should also be updated to provide stricter sanitization, ensuring that the resulting path is safe to use.

---

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Developer Validations

### DIFF
--- a/server/utils/files/index.js
+++ b/server/utils/files/index.js
@@ -177,17 +177,23 @@ async function storeVectorResult(vectorData = [], filename = null) {
 // Purges a file from the documents/ folder.
 async function purgeSourceDocument(filename = null) {
   if (!filename) return;
-  const filePath = path.resolve(documentsPath, normalizePath(filename));
+  try {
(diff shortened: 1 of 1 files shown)


timothycarambat (changes requested):
Looking for feedback on comments

PR #3825, opened 2025-05-14
Ditd 4140

Bert4214:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #xxx

### Developer Validations


timothycarambat:
I see what you are doing here and it all looks functional, but I did not test. However there is no issues associated, no description beyond why this PR is required or the use case it is solving, and otherwise has zero context. We also have reranking in the system as it is - so I dont see the benefit of this beyond using a reranker. That being said this PR has zero context as to what the goal is.

Otherwise, looks like something that would be used in a fork

PR #3832, opened 2025-05-15
Disable message send while content is embedding

timothycarambat:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### What is in this change?
Disable the prompt from being sent while content is actively embedding in the window. This prevents the user from sending a prompt too early where RAG is incomplete, resulting in a bad response since the document cannot be searched.

Will apply:
- Any Drag and Drop doc
- Any pasted doc
- Any paperclip click > select document
- Will be disabled until all objects resolve regardless of status

### Additional Information

- Extracted this disable watch into a hook that we can extend for any other cases we wish to disable sending for whatever reason.

### Developer Validations


shatfield4 (approved):
LGTM

PR #3864, opened 2025-05-21
Add Flowise AI LLM provider integration

### LINKED ISSUE #1785: [FEAT]:  Flowise APi

Do you have a Template to create a Custom Endpoint for your AnythingLLM ? I want to create a way to use Flowise API https://docs.flowiseai.com/using-flowise/api - and have it work with your AnyhtingLLM framework. ?

Rrojaski:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1785

### What is in this change?

Adds Flowise AI as a new LLM provider (environment variable, server settings, and provider class)

Integrates Flowise options into both onboarding's and chat setting's frontend LLM selection UIs

### Developer Validations


server/utils/AiProviders/flowise/index.js
```diff
@@ -0,0 +1,313 @@
...
+          question: lastMessage.content,
+          streaming: true,
+        },
+        {
+          headers: {
+            "Content-Type": "application/json",
+          },
+          responseType: "stream",
+        }
+      );
```
timothycarambat, line 177: According to https://docs.flowiseai.com/api-reference/prediction `streaming` is not a property you can pass in? Also since it can take a history, why not pass in the whole history? In this format it would never recall a conversation message from a previous message.

Likewise, if you are managing the history in flowwise, `/reset` from AnythingLLM would not clear the history from flowise so you would still have issues with history and it retaining information you no longer wish to use

server/utils/AiProviders/flowise/index.js
```diff
@@ -0,0 +1,313 @@
+const { NativeEmbedder } = require("../../EmbeddingEngines/native");
+const {
+  LLMPerformanceMonitor,
+} = require("../../helpers/chat/LLMPerformanceMonitor");
+const {
+  formatChatHistory,
+  writeResponseChunk,
+  clientAbortedHandler,
+} = require("../../helpers/chat/responses");
+const axios = require("axios");
```
timothycarambat, line 10: We don't use `Axios` anywhere in our codebase directly and instead rely on fetch. I'm surprised this import even worked, as it must be a sub-dep of something else we use in this execution path.

server/utils/AiProviders/flowise/index.js
```diff
@@ -0,0 +1,313 @@
...
+      );
+
+    if (!process.env.FLOWISE_LLM_CHATFLOW_ID)
+      throw new Error(
+        "FlowiseLLM must have a valid Chatflow ID to use for the api."
+      );
+
+    this.basePath = process.env.FLOWISE_LLM_BASE_PATH;
+    this.model = "flowise";
+    if (!this.model) throw new Error("FlowiseLLM must have a valid model set.");
```
timothycarambat, line 28: This can never be false since you manually set it and its not user-configurable

server/utils/AiProviders/flowise/index.js
```diff
@@ -0,0 +1,313 @@
...
+    const content = [{ type: "text", text: userPrompt }];
+    for (let attachment of attachments) {
+      content.push({
+        type: "image_url",
+        image_url: {
+          url: attachment.contentString,
+          detail: "high",
+        },
+      });
+    }
```
timothycarambat, line 92: According to the predicition endpoint images should be sent as
```
"uploads": [
    {
      "type": "file",
      "name": "image.png",
      "data": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAABgAAAAYCAYAAADgdz34AAABjElEQVRIS+2Vv0oDQRDG",
      "mime": "image/png"
    }
  ]
  ```

timothycarambat (changes requested):
I dont have Flowise running locally to better understand how this functionality all works, but it has some pretty signifiant code diffs in the LLM provider that would indiciate that flowise is quite different from other model providers and would lack the ability to use historical chats.

If there is an easy flow to create a testing flowise path I can test against I can see more about how this works, but looking at their docs there are a lot of fields and formatting we could use, but dont - or fields that are used that the API says do not exist or are invalid to set.

server/utils/AiProviders/flowise/index.js
Rrojaski: You're right! I’ve updated the format to match the documentation. That said, I noticed Flowise is pretty different from the rest of the Anything LLM app, and these attachments aren’t actually getting passed to the Flowise API yet. Give me a bit to fix that.

server/utils/AiProviders/flowise/index.js
Rrojaski: Removed

server/utils/AiProviders/flowise/index.js
Rrojaski: Haha, yeah, I can use axios. My bad, I didn’t realize. I’ve updated the PR to use fetch instead. Thanks for pointing it out!

PR #3873, opened 2025-05-23
Direct output for agent flows

### LINKED ISSUE #3872: [FEAT]: Direct output to chat from agent flows

- In some cases users may want to directly output the response from an agent flow directly to the chat window
- This is useful for smaller models that are not great at processing lots of data being fed to them
- Prevents the LLM from processing the data before sending it back to the user

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3872 

### What is in this change?

- Allows users to send the response from the last block directly to the UI

### Additional Information

- Downside of direct output is that no future tool calls can be executed since that is the end of the chain. So we need to update docs to reflect this.

### Developer Validations

### DIFF
--- a/.github/workflows/dev-build.yaml
+++ b/.github/workflows/dev-build.yaml
@@ -6,7 +6,7 @@ concurrency:
 
 on:
   push:
-    branches: ['warn-bad-docker-command'] # put your current branch to create a build. Core team only.
+    branches: ['3872-feat-direct-output-to-chat-from-agent-flows'] # put your current branch to create a build. Core team only.
     paths-ignore:
       - '**.md'
       - 'cloud-deployments/*'


shatfield4:
Having issues here with sending direct output to the chat window without killing the aibitat agent session. Everything else works fine in here other than that.

shatfield4 (commented):
LGTM! All is functioning properly and tested multiple flows to ensure the agent session is not being terminated on direct output calls.

scottybo:
This is brilliant!

PR #3924, opened 2025-05-31
Agent skills UI/UX improvements

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [x] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #xxx

### Developer Validations

### DIFF
--- a/frontend/src/pages/Admin/Agents/AgentFlows/index.jsx
+++ b/frontend/src/pages/Admin/Agents/AgentFlows/index.jsx
@@ -13,6 +13,7 @@ export default function AgentFlowsList({
         <a
           href="https://docs.anythingllm.com/agent-flows/getting-started"
           target="_blank"
+          rel="noreferrer"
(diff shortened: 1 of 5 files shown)


shatfield4:
Implemented all new UI elements from designs. Having trouble refactoring backend to get switches working on the SkillList component.

Some notes:
- SQL connector UI is complete as well but backend is also broken since moved from modal to nested component per designs
- Removing the contextual save bar will also improve the UX here a lot but we would have to remove the forms for all the default skills (causes issues with the SQL connector since it's a nested form)
- The only section that was using the save bar was the default skills so it may be confusing to users when they have to click save to save these but for all other sections (custom skills, agent flows, mcp) toggling the switch immediately updates the backend without the extra step for pressing save again

PR #3984, opened 2025-06-11
Flag to disable login UI and endpoints for credentialed auth

### LINKED ISSUE #3982: [FEAT]: Disable username and password sign in for Simple SSO

Here is the workflow I am implementing with AnythingLLM.

- User Access: User visits the app through a reverse proxy.
- Session Check: Proxy checks for a valid session cookie.
- Login Redirect: If no session, proxy redirects user to an external Identity Provider (Keycloak) for authentication.
- Authentication: User logs in at the Identity Provider and is redirected back to the proxy with an authorization code.
- Token Exchange: Proxy exchanges the code for tokens and verifies the ID token.
- User Mapping: Proxy maps the authenticated user to an internal user ID.
- Session Creation: Proxy creates a session cookie for the user.
- Redirect to App: User is redirected to the AnythingLLM with the session active URL.
- Authenticated Requests: Proxy validates sessions on subsequent requests and forwards them to the app.
- Logout: Proxy clears session and redirects user to Identity Provider logout.

Request: Disable or bypass the internal AnythingLLM login page and flow entirely.

In my usecae users must authenticate only through the AnythingLLM SSO via the reverse proxy.
Direct access to AnythingLLM’s native login page should be blocked or redirected.
User creation should not require a password since authentication is handled by the SSO (even if the password is just generated on the backend.)
This ensures centralized auth control and prevents users from bypassing the proxy’s SSO flow.

timothycarambat:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3982

### What is in this change?

- Disables the traditional login pages and backend endpoints the `SIMPLE_SSO_NO_LOGIN` is set in ENV config.

This requires `SIMPLE_SSO_ENABLED` **and** multi-user mode enabled to take effect. It will have no impact on non-authenticated or password-only configurations

When enabled:
- The `/login` page will auto redirect to the token auth page
- The backend `request-token` endpoint will `403` on request
- Creation of new invitations will be blocked on backend (error visible in UI)
- Accepting new invitations where the new user can specify username/password is also blocked.

### Developer Validations

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -329,6 +329,7 @@ GID='1000'
 # Enable simple SSO passthrough to pre-authenticate users from a third party service.
 # See https://docs.anythingllm.com/configuration#simple-sso-passthrough for more information.
 # SIMPLE_SSO_ENABLED=1
+# SIMPLE_SSO_NO_LOGIN=1
 
 # Allow scraping of any IP address in collector - must be string "true" to be enabled
 # See https://docs.anythingllm.com/configuration#local-ip-address-scraping for more information.
--- a/frontend/src/hooks/useSimpleSSO.js
+++ b/frontend/src/hooks/useSimpleSSO.js
@@ -1,0 +1,33 @@
+import { useEffect, useState } from "react";
+import System from "@/models/system";
+
+/**
+ * Checks if Simple SSO is enabled and if the user should be redirected to the SSO login page.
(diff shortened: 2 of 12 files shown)


jessethorn13:
Thank you, excellent work.

Issue #2864, opened 2024-12-17
[FEAT]: Add support for AWS Credentials profile when connecting with Bedrock to avoid frequently updating temporary credentials


SHASHANKB:
### What would you like to see?

Currently Bedrock connector only supports long term IAM User credentials and temporary session credentials. long term IAM User credentials are insecure because they don't rotate automatically. Temporary session credentials need to be update frequently (e.g. every 12 hours)

If you can add support for AWS Credentials profile when connecting with Bedrock, it'll remove the dependency to frequently update the credentials in the Bedrock connector. Since the AWS credentials profile can be setup with IAM assume Role, it doesn't have to depend on using non-secure static IAM User credentials

varunvijamz:
+1 to this feature request. Supporting credentials profiles would make anythingllm much more convenient to use. As an reference, the continue dev plugin for vscode lets you configure bedrock with a region and profile.

heiwais25:
I also had the same ask while using with Bedrock. 
To resolve the issue, I just created a new pull request to add the profile support.
Please let me know if you find anything I missed from PR.

Kornelos:
+1 to this feature - blocks me from using anything-llm currently. @heiwais25 did you get any feedback from team?

timothycarambat:
Close by #3986

PR #3986, opened 2025-06-12
fix to support: feat2864 - using local credentials file with Amazon Bedrock

nonatofabio:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

connect #2864

### What is in this change?

This PR fixes an authentication bug in the AWS Bedrock provider where passing an empty object {} as credentials caused the AWS SDK to reject the configuration with the error:
Resolved credential object is not valid.

The credentials getter in server/utils/AiProviders/bedrock/index.js now returns undefined instead of {} for iam_role and default cases. This enables the AWS SDK to correctly fall back to its default credential provider chain (e.g., local credentials file, environment, or EC2/ECS IAM role), restoring support for local development and environments that use assumed roles.

### Additional Information

Verified that returning undefined allows the AWS SDK to use local credentials as expected.
No other logic or side effects are introduced by this change.

### Developer Validations

- [] Relevant documentation has been updated

### DIFF
--- a/server/utils/AiProviders/bedrock/index.js
+++ b/server/utils/AiProviders/bedrock/index.js
@@ -107,7 +107,7 @@ class AWSBedrockLLM {
       // IAM role is used for long-term credentials implied by system process
       // is filled by the AWS SDK automatically if we pass in no credentials
       case "iam_role":
-        return {};
+        return undefined; // No explicit credentials needed, AWS SDK will use the IAM role of the instance or local credential if returning undefined.
       default:
         return {};
     }


server/utils/AiProviders/bedrock/index.js
```diff
@@ -107,7 +107,7 @@ class AWSBedrockLLM {
       // IAM role is used for long-term credentials implied by system process
       // is filled by the AWS SDK automatically if we pass in no credentials
       case "iam_role":
-        return {};
+        return undefined; // No explicit credentials needed, AWS SDK will use the IAM role of the instance or local credential if returning undefined.
       default:
         return {};
```
timothycarambat: we should return undefined in the default case then as well, yeah?

Issue #2804, opened 2024-12-11
[FEAT]: Disable Streaming via user-defined ENV Config.


timothycarambat:
### What would you like to see?

We should have a global ENV that is defined _not via the UI_ to disable streaming
https://github.com/Mintplex-Labs/anything-llm/blob/69b672b625cc9fc8e4eba3793610b8c8c5cad46f/server/utils/helpers/updateENV.js#L938

connect #1807
connect #2760

PR #4079, opened 2025-07-01
Add option to disable streaming via env for generic openai provider

### LINKED ISSUE #4063: [FEAT]: Disable streaming for LLM Provider(s) with ENV setting

>  is there a way to tell anythingllm that my generic openai provider does not support streaming?

We should have a way to specify or override if a provider is allowed to stream or not. All LLM providers support both chat and stream, so disabling this should be done via a flag.

We can apply this to the Generic OpenAI provider at this time only, since the others are pinned to a service and there is no reason to disable streaming, since that is a better experience.

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4063 

### What is in this change?

- Add env variable for disabling generic OpenAI provider streaming
- Docs here: https://github.com/Mintplex-Labs/anythingllm-docs/pull/177

### Developer Validations

### DIFF
--- a/server/utils/AiProviders/genericOpenAi/index.js
+++ b/server/utils/AiProviders/genericOpenAi/index.js
@@ -35,8 +35,6 @@ class GenericOpenAiLLM {
       user: this.promptWindowLimit() * 0.7,
     };
 
-    this.streamingDisabled =
-      process.env.GENERIC_OPENAI_STREAMING_DISABLED || false;
     this.embedder = embedder ?? new NativeEmbedder();
     this.defaultTemp = 0.7;
     this.log(`Inference API: ${this.basePath} Model: ${this.model}`);
@@ -59,7 +57,7 @@ class GenericOpenAiLLM {
   }
 
   streamingEnabled() {
-    if (this.streamingDisabled) return false;
+    if (process.env.GENERIC_OPENAI_STREAMING_DISABLED === "true") return false;
     return "streamGetChatCompletion" in this;
   }
 
--- a/server/utils/AiProviders/genericOpenAi/index.js
+++ b/server/utils/AiProviders/genericOpenAi/index.js
@@ -35,6 +35,8 @@ class GenericOpenAiLLM {
       user: this.promptWindowLimit() * 0.7,
     };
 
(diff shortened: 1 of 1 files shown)


server/utils/AiProviders/genericOpenAi/index.js
```diff
@@ -57,6 +59,7 @@ class GenericOpenAiLLM {
   }
 
   streamingEnabled() {
+    if (this.streamingDisabled) return false;
     return "streamGetChatCompletion" in this;
   }
```
timothycarambat: Just check for the ENV in this function - there is no need to attach this as a property to the class. You can check if a process key is defined and if it is - return `false`.

Otherwise we can check for `streamGetChatCompletion` property on the class. Since this property is not used elsewhere there is no need to keep in on the class object

timothycarambat (changes requested):

timothycarambat (changes requested):
Is it supposed to be `true` explictly or **any** value? Either the docs are wrong and need ot be updated or this line can be updated
https://github.com/Mintplex-Labs/anythingllm-docs/pull/177/files#diff-f77de0c7d4ad5c26479f2df4c3e6a6ac69170e83dd0665bc0cc0ed1698dbcd6eR211-R217

```
// This would be any value
if("GENERIC_OPENAI_STREAMING_DISABLED" in process.env) return false;
```

shatfield4:
> Is it supposed to be `true` explictly or **any** value? Either the docs are wrong and need ot be updated or this line can be updated https://github.com/Mintplex-Labs/anythingllm-docs/pull/177/files#diff-f77de0c7d4ad5c26479f2df4c3e6a6ac69170e83dd0665bc0cc0ed1698dbcd6eR211-R217
> 
> ```
> // This would be any value
> if("GENERIC_OPENAI_STREAMING_DISABLED" in process.env) return false;
> ```

Fixed typo in docs.

timothycarambat:
resolved #2804

PR #3799, opened 2025-05-10
Custom Roles & Perms - RBAC

timothycarambat:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #1787

### Developer Validations

- [ ] I ran `yarn lint` from the root of the repo & committed changes
- [ ] Relevant documentation has been updated
- [ ] I have tested my code functionality
- [ ] Docker build succeeds locally

PR #4104, opened 2025-07-08
Feat: creator permission level

filipwronskilc:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### What is in this change?
New permission level "creator". User with this permission level can:
- create workpaces
- open file manager + upload files (no data connectors access)
- open workspace settings -> access to General Settings, Chat Settings (without choosing the model), Vector Database
- perform changes to their workspaces

### Developer Validations

### DIFF
--- a/frontend/src/components/PrivateRoute/index.jsx
+++ b/frontend/src/components/PrivateRoute/index.jsx
@@ -123,7 +123,8 @@ export function ManagerRoute({ Component }) {
   }
 
   const user = userFromStorage();
-  return isAuthd && (!["default", "creator"].includes(user?.role) || !multiUserMode) ? (
+  return isAuthd &&
+    (!["default", "creator"].includes(user?.role) || !multiUserMode) ? (
     <KeyboardShortcutWrapper>
       <UserMenu>
         <Component />
--- a/frontend/src/pages/WorkspaceSettings/ChatSettings/index.jsx
+++ b/frontend/src/pages/WorkspaceSettings/ChatSettings/index.jsx
@@ -18,7 +18,7 @@ export default function ChatSettings({ workspace }) {
   const [saving, setSaving] = useState(false);
 
   const formEl = useRef(null);
-  
+
   const user = userFromStorage();
   const userRole = user?.role;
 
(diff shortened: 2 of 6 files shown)


pawelrosada:
@timothycarambat hello ;)

Could you explain why this PR was closed? That was a really helpful feature for this level of permissions.

timothycarambat:
@pawelrosada Sorry for the lack of explanation. This PR has no associated issue so it has not even been validated or triaged for its utility into the core app. When something is merged ultimately **we** become maintainers of the changes for perpetuity - so when it comes to things that are critical to the core functions - like permissioning, we usually tag those issues with "core team only". 

Adding a new role is certainly lower lift in effort; however, it makes more sense for the overhaul of permissions to be controlled by system admins instead of creating every combination of roles possible, since every person has preferences on how their specific installation of AnythingLLM works.

While it is substantially more work to implement RBAC (#3799) this is likely a more suitable long-term solution rather than adding more hardcoded roles. This PR seems like it wants to address "user-owned workspaces", but I cannot be sure.

It is on us to update our Contributor guidelines to prevent situations like this and I will do that now since situations like this seem to be occurring much more frequently now. In the interim, changes like this PR, while functional for your case, are best on a fork

pawelrosada:
@timothycarambat 

Everything is clear, it seems that our organization would like to develop this tool further, we wouldn't want to make a separate fork. Do you have any plan or way how one can join you as core developers discussing the vision etc? We have at least several ideas for interesting development and perhaps instead of doing it separately we could join forces?

timothycarambat:
> we wouldn't want to make a separate fork.

Can I ask why? typically, merging upstream code is not a massive lift. It seems like this particular issue is a top priority for you and your organization and we are happy to work with people to get things to a suitable state.

What is the main lacking feature or blocker that you need at this time and we can go from there? Is it the ability to allow any subuser to own their own workspaces and manage them?

pawelrosada:
@timothycarambat 

The main gaps for us in terms of permissions are several things - one of them is the ability to share a Workspace with all users - such a "basic Workspace that everyone would have". Another case is the creation of a Workspace by a regular user, which they can then share with other people - by adding them to a list or just use own Workspace

Regarding not creating our own fork - we would like to contribute to the software development and add new functionalities, such as we currently have planned to submit a PR for adding Workspace search functionality, work with image creation and recently one of our people added Polish language support, etc. - it's simply easier to maintain this together in one place than to create separate forks that then need to be synchronized, and a large number of changes could simply make synchronization with new releases difficult.

PR #4114, opened 2025-07-09
Configurable message limit for embed chat widget

### LINKED ISSUE #4043: [FEAT] Embed chat history limit setting
> It is currently a fixed limit 
> https://github.com/Mintplex-Labs/anything-llm/blob/c4f49ff6b445f8f1b15d3bd42f8a2d24317b745c/server/utils/chats/embed.js#L37
> 
> However, it may make sense to have this be a property server-side on the embed config 

 _Originally posted by @timothycarambat in [#4027](https://github.com/Mintplex-Labs/anything-llm/issues/4027#issuecomment-3000950911)_

shatfield4:
r embed widget

 ### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4043

### What is in this change?

- Allows users to set and modify the message history limit to include as chat context for embed chat widgets

### Developer Validations


server/prisma/schema.prisma
```diff
@@ -240,6 +240,7 @@ model embed_configs {
   allow_prompt_override      Boolean       @default(false)
   max_chats_per_day          Int?
   max_chats_per_session      Int?
+  message_limit              Int           @default(20)
```
timothycarambat: Make `?` optional - this change makes a new migration that drops the whole table and recreates. This is usually something we want to avoid at all costs.

Migration should look like a single column addition

server/utils/chats/embed.js
```diff
@@ -34,7 +34,7 @@ async function streamChatWithForEmbed(
   });
   const VectorDb = getVectorDbClass();
 
-  const messageLimit = 20;
+  const messageLimit = embed.message_limit;
```
timothycarambat: Add fallback

server/models/embedConfig.js
```diff
@@ -230,6 +236,9 @@ function validatedCreationData(value, field) {
   }
 
   if (NUMBER_KEYS.includes(field)) {
+    if (field === "message_limit") {
+      return !isNaN(value) && Number(value) >= 0 ? Number(value) : 20;
+    }
```
timothycarambat: Instead of a nested conditional, can we just have a guard clause about the Number `if` for the field?

timothycarambat (changes requested):

Issue #4272, opened 2025-08-11
[BUG]: Update Security UI to match all other Settings


AshikaAnand12:
### How are you running AnythingLLM?

Docker (local)

### What happened?

The UI for the Security tab in the General Settings does not match the rest of the settings Tab.

Suggesting a change to:

<img width="1169" height="756" alt="Image" src="https://github.com/user-attachments/assets/36168f95-535f-4c3c-bebc-c9663710a547" />

PR #4274, opened 2025-08-11
Update Security UI to match all other Settings

AshikaAnand12:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [X] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves [#4272](https://github.com/Mintplex-Labs/anything-llm/issues/4272)

### What is in this change?

Suggesting an UI update to the Security tab in the General Settings to ensure all the tabs are consistent

### Additional Information

Suggested change is added a screenshot in the Issue

### Developer Validations

- [X] Relevant documentation has been updated - no update to documentation required


AshikaAnand12:
Hello,
I normalized the translations and also verified translations. 
For the css designs, I copied the same styles from other settings like `settings/text-splitter-preference`

AshikaAnand12:
Done. The translations are normalized, so there won't be any issue in another language but the values are set to null after I ran normalizeEn.mjs

shatfield4:
> Done. The translations are normalized, so there won't be any issue in another language but the values are set to null after I ran normalizeEn.mjs

Okay so this will take a little more work to get this complete now. But as you can see after normalizing and linting all the files it makes it much easier to see what is being changed.

You are correct that all these keys are null in other languages and this is the problem with moving the existing keys around. The reason is because the normalize script is just checking all other translation files to ensure they contain the same structure as the English translations. If the keys are different, the script will automatically create matching keys by default and set them to null. In your PR it makes complete sense for moving these keys around and makes the translations structured better. 

What you will need to do in order to ensure we do not lose our existing translations from these files is to manually fill each one of the new null keys in with the previous translations. You can do this by opening the files changed tab in this PR and just copy manually each key to the new keys that were created. Once this is done we can get it merged in. Great work so far on your first contribution!

AshikaAnand12:
I added back the values for everything that was replaced to null. 
After making the change, I reran normalizeTranslations, verifyTranslations and yarn lint to ensure there is no discrepancy

PR #4317, opened 2025-08-20
API request delay for Generic OpenAI embedding engine

### LINKED ISSUE #3570: [FEAT]: Configurable Delay Between Embedding Requests

I'm currently using the [experimental Gemini Embeddings model](https://developers.googleblog.com/en/gemini-embedding-text-model-now-available-gemini-api/) via the generic OpenAI API in AnythingLLM Desktop. The model has a rate limit of 5/10 RPM. Even after setting the max concurrent chunks to 1, I sometimes still encounter error 429 when uploading some PDF files (sent one by one every few minutes, with each file being less than 1MB). The same files can be embedded smoothly using Gemini API text-embedding-004.

[429 RESOURCE_EXHAUSTED	You've exceeded the rate limit.	You are sending too many requests per minute with the free tier Gemini API.	Ensure you're within the model's rate limit. Request a quota increase if needed]

Would it be possible to add a configuration option that allows users to set a delay between embedding requests?

chaserhkj:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3570 

### What is in this change?

This PR adds a new configurable environment variable to add a small delay for dispatching each API request calls when using the generic OpenAI embedding engine.

This is meant to be used together with `GENERIC_OPEN_AI_EMBEDDING_MAX_CONCURRENT_CHUNKS=1` and is mostly relevant to the use of local LLM backends like `llama.cpp`, where their API server implementation is naive and will just reject API requests with a 429 error if the underlying worker is busy.

### Developer Validations

I am skipping documentation for now since there is no change to front end UI, but I could add it to this PR if requested.


timothycarambat:
The previous change here would execute all the batches at once, and they would all just wait at the same time instead of doing each batch sequentially with a wait in between. Refactored code solves this and was tested.

PR #4402, opened 2025-09-18
Introduce keyboard (Up/Down/Enter/Escape) selection for slash commands

### LINKED ISSUE #2019: [FEAT]: Enhanced Keyboard Navigation for Preset Prompts Menu

Firstly, I don't think I've really ever said thanks for your work on Anything LLM - huge thanks to the Anything LLM team for creating such an amazing open-source product, licensed under MIT. The separate workspaces with RAG segregation, and prompt menu (hence this feature) and different models per workspace. Love it. Thanks so much.

**Feature Request:**

When clicking into the prompt window, users can type `/` to open the preset prompts menu. Currently, navigation within this menu requires using the mouse. To streamline the interaction, I propose adding keyboard navigation to this menu.

**Specifically:**

1. After typing `/` to open the menu, allow users to navigate the menu items using the **Up** and **Down** key.
2. Provide two options to interact with the selected menu item:
	* Press **Escape** to close the menu and return to the prompt window.
	* Press **Enter** to add the selected preset prompt to the chat.

**Benefits:**

This feature would significantly reduce the need to switch between keyboard and mouse, making the interaction more efficient and user-friendly. By keeping the user on the keyboard, we can reduce the back-and-forth between input methods and improve the overall experience.

**Current Behavior:**

Currently, users need to:

1. Click into the prompt window with the mouse.
2. Type `/` with the keyboard to open the menu.
3. Switch back to the mouse to select a menu item.
4. Click back into the prompt window to return to keyboard input.

**Example Use Case:**

A user wants to explore different preset prompts to generate creative content. With the proposed feature, they can type `/` to open the menu, navigate using the **Up** and **Down** arrow key, and select a prompt using **Enter**. This workflow would be more seamless and efficient.

Hopefully it's something you also think would improve the user experience and might not be too much work (I would attempt to even implement this, but I can't seem to get github-codespaces (and not locally) dev set up working (it sits there typing to build some dependencies forever - I might look at this tomorrow).

angelplusultra:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #2019

### What is in this change?

This PR introduces keyboard navigation for the slash command selection popup.

Users can now:
- Navigate options using the `Up` and `Down` arrow keys.
- Select a command by pressing `Enter`.
- Close the popup by pressing `Escape`.

### Developer Validations

### DIFF
--- a/frontend/src/hooks/useSlashCommands.js
+++ b/frontend/src/hooks/useSlashCommands.js
@@ -6,7 +6,6 @@ export function useSlashCommands({ sendCommand }) {
   const [showSlashCommand, setShowSlashCommand] = useState(false);
   const [highlightedSlashCommand, setHighlightedSlashCommand] = useState(1);
   const [slashCommands, setSlashCommands] = useState([]);
-  //   const [qtyOfSlashCommands, setQtyOfSlashCommands] = useState(0);
   // Ref to the textarea element, this is needed to apply focus to the textarea when slash commands are triggered by clicking the slash command button
   const textareaRef = useRef(null);
 
@@ -24,7 +23,6 @@ export function useSlashCommands({ sendCommand }) {
   const fetchSlashCommands = async () => {
     const presets = await System.getSlashCommandPresets();
     setSlashCommands(presets);
-    // setQtyOfSlashCommands(presets.length + 1);
   };
 
   useEffect(() => {
@@ -95,8 +93,6 @@ export function useSlashCommands({ sendCommand }) {
     setShowSlashCommand,
     highlightedSlashCommand,
     setHighlightedSlashCommand,
-    // qtyOfSlashCommands,
-    // setQtyOfSlashCommands,
     watchForSlash,
     handleKeyDown,
     textareaRef,


frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/SlashCommands/SlashPresets/index.jsx
```diff
@@ -113,10 +149,11 @@ export default function SlashPresets({ setShowing, sendCommand, promptRef }) {
 
   return (
     <>
-      {presets.map((preset) => (
+      {presets.map((preset, index) => (
```
timothycarambat, line 123: Can we not do highlight operation with `preset.id`?

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/SlashCommands/SlashPresets/index.jsx
```diff
@@ -187,7 +224,7 @@ function PresetItem({ preset, onUse, onEdit, onPublish }) {
   return (
     <button
       onClick={onUse}
-      className="border-none w-full hover:cursor-pointer hover:bg-theme-action-menu-item-hover px-2 py-2 rounded-xl flex flex-row justify-start items-center relative"
+      className={`border-none w-full hover:cursor-pointer hover:bg-theme-action-menu-item-hover px-2 py-2 rounded-xl flex flex-row justify-start items-center relative ${highlighted ? "bg-theme-action-menu-item-hover" : ""}`}
```
timothycarambat: 👍

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/SlashCommands/index.jsx
```diff
@@ -32,7 +32,14 @@ export default function SlashCommandsButton({ showing, setShowSlashCommand }) {
...
 }
 
-export function SlashCommands({ showing, setShowing, sendCommand, promptRef }) {
+export function SlashCommands({
+  showing,
+  setShowing,
+  sendCommand,
+  promptRef,
+  highlightedSlashCommand,
+  setHighlightedSlashCommand,
```
timothycarambat: If we are passing the state getter and setter, it _could_ be scoped too high and might make more sense inside of this component.

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/SlashCommands/reset.jsx
```diff
@@ -1,9 +1,38 @@
+import React, { useEffect } from "react";
 import { useIsAgentSessionActive } from "@/utils/chat/agent";
 import { useTranslation } from "react-i18next";
 
-export default function ResetCommand({ setShowing, sendCommand }) {
+export default function ResetCommand({
+  setShowing,
+  sendCommand,
+  highlightedSlashCommand,
```
timothycarambat: would make more sense as a prop like `isHighlighted`, no?

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/SlashCommands/reset.jsx
```diff
@@ -1,9 +1,38 @@
...
+        // Reset command is selected
+        setShowing(false);
+        sendCommand({ text: "/reset", autoSubmit: true });
+      }
+    };
+
+    window.addEventListener(
+      "selectHighlightedSlashCommand",
+      handleSelectHighlighted
+    );
```
timothycarambat: For events, we often export them as consts so we can update the string of an event name without a lot of LOC. [example with attachment event handling](https://github.com/Mintplex-Labs/anything-llm/blob/7549f51dbca5c87ab01e7a83bdb9a912ba6fdea0/frontend/src/components/WorkspaceChat/ChatContainer/DnDWrapper/index.jsx#L12)

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/SlashCommands/reset.jsx
```diff
@@ -1,9 +1,38 @@
...
+
+    return () => {
+      window.removeEventListener(
+        "selectHighlightedSlashCommand",
+        handleSelectHighlighted
+      );
+    };
```
timothycarambat: Good to cleanup 👍

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
```diff
@@ -41,11 +42,12 @@ export default function PromptInput({
   const [promptInput, setPromptInput] = useState("");
   const { showAgents, setShowAgents } = useAvailableAgents();
   const { showSlashCommand, setShowSlashCommand } = useSlashCommands();
+  const [highlightedSlashCommand, setHighlightedSlashCommand] = useState(1);
   const formRef = useRef(null);
   const textareaRef = useRef(null);
-  const [_, setFocused] = useState(false);
```
timothycarambat, line 46: Why was this removed?

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
```diff
@@ -60,9 +62,17 @@ export default function PromptInput({
...
+    const presets = await System.getSlashCommandPresets();
+    setQtyOfSlashCommands(presets.length + 1);
+  };
+
+  useEffect(() => {
+    fetchPresets();
+  }, []);
+
   useEffect(() => {
-    if (!!window)
```
timothycarambat, line 64: Functionally, while `!!` and just eval'ing the var in `if` are the same, the idea is to communicate this var should exist as any truthy value. Stylistically, We use this nomenclature in a lot of places since we are not TS to communicate truthiness, there is no need to remove them as we will wind up two styles throughout the app

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
```diff
@@ -99,7 +108,10 @@ export default function PromptInput({
 
   function checkForSlash(e) {
     const input = e.target.value;
-    if (input === "/") setShowSlashCommand(true);
+    if (input === "/") {
+      setShowSlashCommand(true);
+      setHighlightedSlashCommand(0); // Start with reset command highlighted
```
timothycarambat: If this state was managed at the component level for slash commands - you would not need to do this and it would be handled on mount of SlashCommand modal

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
```diff
@@ -121,6 +133,8 @@ export default function PromptInput({
     // Is simple enter key press w/o shift key
     if (event.keyCode === 13 && !event.shiftKey) {
       event.preventDefault();
+      // Don't submit if slash commands are showing - let the slash command handler take care of it
+      if (showSlashCommand) return;
```
timothycarambat, line 123: good catch

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
```diff
@@ -274,9 +329,7 @@ export default function PromptInput({
                   handlePasteEvent(e);
                 }}
                 required={true}
-                onFocus={() => setFocused(true)}
```
timothycarambat, line 277: Why was this removed?

timothycarambat (changes requested):
Not a heavy code review atm, but current feedback:
- If i click on the slash command button `[/]` in the prompt box, keyboard nav does not work
- Keyboard nav does work if I type `/`
- If I type `/` nav down one or more, click "Add new preset" and add it. Then i type `/` again, my new preset is unreachable via keyboard nav (because qty state is out of sync probably). Does work on page reload.

Ideally, either way the modal pops up, this functionality for navigation becomes usable and is cleaned up and unmounted when the modal is dismissed. 

Instead of using JS state, you could probably more easily do this with DOM manipulation while still using events. This would make counting results on the fly easy, work on all cases, and when pressing enter all the logic we have around clicking the item would take over - simplifying things a lot (maybe?) ;)

angelplusultra:
> Not a heavy code review atm, but current feedback:
> 
>     * If i click on the slash command button `[/]` in the prompt box, keyboard nav does not work
> 
>     * Keyboard nav does work if I type `/`
> 
>     * If I type `/` nav down one or more, click "Add new preset" and add it. Then i type `/` again, my new preset is unreachable via keyboard nav (because qty state is out of sync probably). Does work on page reload.
> 
> 
> Ideally, either way the modal pops up, this functionality for navigation becomes usable and is cleaned up and unmounted when the modal is dismissed.
> 
> Instead of using JS state, you could probably more easily do this with DOM manipulation while still using events. This would make counting results on the fly easy, work on all cases, and when pressing enter all the logic we have around clicking the item would take over - simplifying things a lot (maybe?) ;)

Fixes in [3e85053](https://github.com/Mintplex-Labs/anything-llm/pull/4402/commits/3e85053755a50e7aecb3e54a2aef49c15deb4a93)

1. I abstracted most of the logic related to slash commands into the custom hook `useSlashCommands` to keep the `PromptInput` comp as light as possible.
2. I fixed the bug when the slash command popup would be triggered via the slash command button and wouldn't respond keyboard navigation/selection
3. I also found a bug in the process where if a user had enough slash commands to cause vertical overflow, out of view slash commands would be highlighted but not visible, I added scroll-into-view logic when a slash command is highlighted.
4. The bug where newly created slash commands would not be highlightable/selectable is now fixed.

PR #4425, opened 2025-09-22
Add HTTP request/response logging middleware for development mode

### LINKED ISSUE #4401: [FEAT]: Dev HTTP Logger

Need an HTTP logger middleware on `/server` and `/collector` to log the path (ex. /workspaces/27) and method (ex. GET) of each request. Should only be enabled in development.

angelplusultra:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4401 

### What is in this change?

This PR introduces an HTTP logging middleware layer in both express servers (`/server` and `/collector`) that logs:
- The Request Method ( ex. GET)
- The Request Path (ex. /workspace/my-workspace/stream-chat)
- The Response Status Code (ex. 200)
- An optional time log of the moment the response is sent.

image

With time log:

image

### Developer Validations

### DIFF
--- a/collector/__tests__/utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.test.js
+++ b/collector/__tests__/utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.test.js
@@ -1,16 +1,32 @@
 const { YoutubeTranscript } = require("../../../../../utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.js");
 
 describe("YoutubeTranscript", () => {
-  it("should fetch transcript from YouTube video", async () => {
-    const videoId = "BJjsfNO5JTo";
-    const transcript = await YoutubeTranscript.fetchTranscript(videoId, {
-      lang: "en",
-    });
+  if (process.env.GITHUB_ACTIONS) {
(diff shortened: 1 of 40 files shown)


collector/index.js
```diff
@@ -15,9 +15,13 @@ const { wipeCollectorStorage } = require("./utils/files");
 const extensions = require("./extensions");
 const { processRawText } = require("./processRawText");
 const { verifyPayloadIntegrity } = require("./middleware/verifyIntegrity");
+const { httpLogger } = require("./middleware/httpLogger");
 const app = express();
 const FILE_LIMIT = "3GB";
 
+if (process.env.NODE_ENV === "development") {
```
timothycarambat: This should be optional to _enable_ in development. So there should be another process.env to check here that is documented in the .env.example as well and disabled by default.

timothycarambat (changes requested):
Make optional, even in development. I would turn this on as a debug tool, but not have it on all the time so I can get clean logs when doing other items.

collector/index.js
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4425/commits/da4ecf6ec3fff10b68854c2c1629ee46631e80f0

PR #4441, opened 2025-09-26
Validation of PG Vector Connection on Message Sent

### LINKED ISSUE #4431: [FEAT]: Validate PG Vector Database Connection

Currently, ALLM does not validate the connection details provided for any vector database.
If the connection string is incorrect or the database is unavailable, initiating a chat with an LLM results in a prolonged loading state (in my case, 2 minutes and 35 seconds with a PGVector instance) before the LLM finally responds.

Once the timeout lifts and the LLM proceeds to respond, **there’s no indication that the application failed to connect to the user’s vector database.** This can easily lead to confusion, as the user might assume the delay is due to the AI provider rather than a database connection issue.

⸻

### Proposed Solutions

1. Validate Connection on Setup

When users provide their vector database connection details in Settings > Vector Database and click Save, the system should attempt to connect immediately:
	•	 If successful: Save the details as usual.
	•	 If unsuccessful: Alert the user that the connection could not be established.

This would catch misconfigurations early. However, it won’t address cases where the database becomes unreachable later, such as during runtime.

⸻

2. Validate Connection Before Enabling Chat

When a user navigates to a chat session, attempt to connect to their vector database before enabling chat functionality:
	•	Display a loading state while checking the connection.
	•	 If the connection is successful: Proceed and allow the user to chat.
	•	 If unsuccessful: Alert the user that the database is unreachable instead of leaving them stuck in a long, unexplained loading state.

⸻

3. Both of the above

This approach would fully resolve the current issue and improve the overall user experience.

angelplusultra:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4431 

### What is in this change?

This PR introduces an explicit heartbeat check when PGVector is selected as the user’s vector database. The check runs whenever a chat message is sent. If the heartbeat takes longer than 3 seconds, it triggers a timeout and returns a clear chat error to the user, indicating a PostgreSQL connection issue.

This check is non-blocking, once the error is returned, the process continues with the rest of the streaming workflow, allowing the LLM to respond to the prompt while skipping the similarity search step.

## Previous Behavior

The connection to Postgres would timeout after a couple minutes with no indication of this connection problem to the user. LLM responds at 2:35.

https://github.com/user-attachments/assets/4b5bb956-e554-42c3-ada7-1d21de425a75

## Current Behavior

The connection times out after 3 seconds, sending a chat error to the user notifying them of the connection issue and then proceeding with the LLM response.

https://github.com/user-attachments/assets/30437c4c-8319-49ac-9ff0-28520b7d1465

### Developer Validations

### DIFF
--- a/server/utils/chats/stream.js
+++ b/server/utils/chats/stream.js
@@ -56,9 +56,34 @@ async function streamChatWithWorkspace(
   });
   const VectorDb = getVectorDbClass();
 
+  let vectorDatabaseIsAlive = true;
+  // If the vector database is PostgreSQL, we need to check if it is alive. The reason for this explicit heartbeat check is because Postgres will hang on connection if the database is not running without any error.
+  if (VectorDb.name === "PGVector") {
+    console.log("Checking if PostgreSQL is alive...");
+    const heartbeat = await VectorDb.heartbeat();
+    if (heartbeat.error) {
+      writeResponseChunk(response, {
+        id: uuid,
+        type: "textResponse",
+        sources: [],
+        close: true,
+        error: `Postgres::${heartbeat.error}. Failed to connect to the PostgreSQL database. Please verify your connection string and confirm the database server is running.`,
+      });
+      vectorDatabaseIsAlive = false;
+    }
+  }
+
   const messageLimit = workspace?.openAiHistory || 20;
-  const hasVectorizedSpace = await VectorDb.hasNamespace(workspace.slug);
-  const embeddingsCount = await VectorDb.namespaceCount(workspace.slug);
+
+  let hasVectorizedSpace = false;
+  let embeddingsCount = 0;
+  // Only query the vector database if it is alive. This will prevent the chat from hanging if the database is not running if using PostgreSQL.
+  if (vectorDatabaseIsAlive) {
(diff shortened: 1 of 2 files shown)


timothycarambat:
Closing this for a few reasons:
- I think the idea of being able to report a RAG search failure is good, but the UI/UX for this should be in the message as a hint below the response if there was a failure. Not a big red failure in chat.
- The idea of doing a pre-check is actually very useful, since all of the providers have a `heartbeat` we should just use that instead - some providers have it stubbed though.
- In specific, I cannot replicate this issue using Postgres as a brew service. even when totally offline, the timeout is basically zero.

**However** the biggest thing that we can do to allow people to override the timeout with no code change on our end is to have the user just add it in [their connection string](https://www.npgsql.org/doc/connection-string-parameters.html).

Adding `Timeout=30;` or `Command Timeout=30;` as connection string vars would suffice without us having to implement code we need to maintain. The `pg` code would handle this as is currently.

PR #4459, opened 2025-09-30
Fix TPS Miscalculation for Gemini

### LINKED ISSUE #4449: [BUG/CHORE]: Gemini models token metrics incorrect

All versions

### What happened?

The TPS measurement of Gemini models is incorrect. This is likely due to manual counting of the tokens on request, which has always been a rough estimation. If possible, this value should come directly from the API chunks.

Use Google Gemini as the LLM provider (any model)
In a chat send a simple prompt like `Tell me a short story`
Observe that the metrics are clearly wrong (~3-8 tps, which should be 100+)

More Context:
https://github.com/Mintplex-Labs/anything-llm/pull/4459#pullrequestreview-3287426632

angelplusultra:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4449 

### What is in this change?

This PR fixes under-reported Tokens Per Second (TPS) and completion token counts, especially for Gemini, by estimating tokens from the full generated text when provider usage is missing or unavailable during streaming.

`server/utils/helpers/chat/responses.js`
- Estimate `completion_tokens` from the accumulated full streamed text using `TokenManager` on finish, abort, and error paths.
- Keep existing prompt token estimation.
- Maintain SSE behavior and metrics emission format.

`server/utils/AiProviders/gemini/index.js`
- Add non-stream fallback: when `usage.completion_tokens` is missing, tokenize `textResponse` to estimate completion tokens.
- Recompute `total_tokens` and `outputTps` using the fallback value.

### Developer Validations

### DIFF
--- a/server/utils/AiProviders/openAi/index.js
+++ b/server/utils/AiProviders/openAi/index.js
@@ -168,11 +168,11 @@ class OpenAiLLM {
     return {
       textResponse: result.output.output_text,
       metrics: {
-        prompt_tokens: usage.prompt_tokens || 0,
-        completion_tokens: usage.completion_tokens || 0,
+        prompt_tokens: usage.input_tokens || 0,
+        completion_tokens: usage.output_tokens || 0,
         total_tokens: usage.total_tokens || 0,
-        outputTps: usage.completion_tokens
-          ? usage.completion_tokens / result.duration
+        outputTps: usage.output_tokens
+          ? usage.output_tokens / result.duration
           : 0,
         duration: result.duration,
       },
@@ -224,6 +224,7 @@ class OpenAiLLM {
             if (token) {
               fullText += token;
               if (!hasUsageMetrics) usage.completion_tokens++;
+
               writeResponseChunk(response, {
                 uuid,
                 sources: [],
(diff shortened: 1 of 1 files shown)


timothycarambat (changes requested):
This seems like a lot of LoC for something that should be normalized as output from the OpenAI SDK.

Two things to confirm first.
**Non-streaming**
The code as is should be OK. You can test this by running a prompt in a simple chat and then changing 
```js
streamingEnabled() {
    return false;
   // return "streamGetChatCompletion" in this;
  }
```
 
This will cause gemini to also be non-streaming and you can inspect `result` from `getChatCompletion` to ensure the result.usage object has the proper values (`prompt_tokens`, completion tokens, total, etc)

 -----
 **For streaming**
 This should be easily doable via the OpenAI SDK. By default the streaming chunk events do not report usage ever - so that is why we have to manually calculate them.
 
 I would check if we can simply add this parameter on [Chat Completions With Streaming](https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream_options) in `streamGetChatCompletion`
 
 ```js
async streamGetChatCompletion(messages = null, { temperature = 0.7 }) {
    const measuredStreamRequest = await LLMPerformanceMonitor.measureStream(
      this.openai.chat.completions.create({
        model: this.model,
        stream: true,
        messages,
        temperature: temperature,
        stream_options: {
           include_usage: true // now we can get `usage` in final chunk of stream
         }
      }),
      messages,
      false // Disables manual token counting (default: true)
    );
    return measuredStreamRequest;
  } 
  ```
  
  After this, our stream handler should detect `usage` in a chunk and apply it properly.
  
  Ref: https://github.com/Mintplex-Labs/anything-llm/blob/master/server/utils/helpers/chat/LLMPerformanceMonitor.js#L65

PR #4466, opened 2025-10-01
Add stream options to Gemini LLM for usage tracking

### LINKED ISSUE #4449: [BUG/CHORE]: Gemini models token metrics incorrect

angelplusultra:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4449 

### What is in this change?

This PR enhances the Gemini provider by enabling token metrics on response chunks, leading to accurate token measurement.

### Developer Validations

### DIFF
--- a/server/utils/AiProviders/gemini/index.js
+++ b/server/utils/AiProviders/gemini/index.js
@@ -416,6 +416,9 @@ class GeminiLLM {
         stream: true,
         messages,
         temperature: temperature,
+        stream_options: {
+          include_usage: true,
+        },
       }),
       messages,
       true
--- a/server/utils/AiProviders/gemini/index.js
+++ b/server/utils/AiProviders/gemini/index.js
@@ -421,7 +421,7 @@ class GeminiLLM {
         },
       }),
       messages,
-      true
+      false
     );
 
     return measuredStreamRequest;


server/utils/AiProviders/gemini/index.js
```diff
@@ -416,6 +416,9 @@ class GeminiLLM {
         stream: true,
         messages,
         temperature: temperature,
+        stream_options: {
+          include_usage: true,
+        },
       }),
       messages,
       true
```
timothycarambat: This should be false, the [`measureStream`](https://github.com/Mintplex-Labs/anything-llm/blob/d6f0d305ab5e221911ba9141547176b8975aa52f/server/utils/helpers/chat/LLMPerformanceMonitor.js#L62) takes a 3rd positional value for `runPromptTokenCalculation` since more often than not, we cannot get usage metrics from a provider.

This will override the output of the usage and use our token counter - which is less accurate.

timothycarambat (changes requested):
Review comment.

server/utils/AiProviders/gemini/index.js
angelplusultra: I'm getting an accurate token calculation now after providing that options object to enable token metrics for streaming (100+ tps). 

All `runPromptTokenCalculation` seems to do is enable or disable counting tokens in the prompt:
https://github.com/Mintplex-Labs/anything-llm/blob/d6f0d305ab5e221911ba9141547176b8975aa52f/server/utils/helpers/chat/LLMPerformanceMonitor.js#L72

Which is then overwritten by the `endMeasurement()` method once called, which doesn't seem right:

https://github.com/Mintplex-Labs/anything-llm/blob/d6f0d305ab5e221911ba9141547176b8975aa52f/server/utils/helpers/chat/LLMPerformanceMonitor.js#L90

angelplusultra: fix in https://github.com/Mintplex-Labs/anything-llm/pull/4466/commits/2775657311eaf878c8d0f18f1b391eee2db80625

PR #4467, opened 2025-10-01
Migrate OpenAI Agent to use ResponsesAPI

### LINKED ISSUE #4465: [CHORE]: Migrate OpenAI `agent` provider to Responses API

The [Responses API](https://platform.openai.com/docs/api-reference/responses/create) for OpenAI is useful since it paves over the nuances of model differences without us having to create cases for each model - which changes so often.

The Agent provider for OpenAI is still using completions - breaking GPT-5, O-* outputs and generally causing incompatibilities with tools. Migrating to responses should resolve all of this.

This can have a ton of side-effects and should have **no impact** on other providers who override the baseURL or the `OpenAI` SDK we use where we leverage `completions` instead.

timothycarambat:
### Pull Request Type

- [x] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4465

### What is in this change?

Migrate OpenAI `aibitat` provider to use the `Responses` API from the OpenAI SDK. This allows us to easily support both past and future models through a unified API so now you can use agents for `gpt-5` and the `o-*` models as well as the previous models that have already been launched.

### Additional Information

Due to the unique flow by the responses API:
- OpenAI provider now has its own stream and complete functions on its class to override the inherited `Provider` which still will use completions (for unification of interface among all other LLM providers)
- Required us to add `originalFunctionCall` to the stream and non-stream handler for the main aibitat class so that we can consistently rebuild the function call post-execution since the message history must have a matching `function_call` and `function_call_output` in the messages array in order to be valid.

### Developer Validations

### Additional QA

- [ ] Test new models (gpt-5, o-*, etc)
- [ ] Test older gen models like 3.5
- [ ] Test with and without tools available
- [ ] Test with streaming disabled (manually) to force `complete` to run
- [ ] Test with followup messages, multiple tool calls, etc.
- [ ] Test existing providers for spot check for changes introduced to the aibitat handlers.


timothycarambat:
Know bad models:
- o4 deep research ones
- o1-* (dont use responses API??)

Everything else works

Issue #4411, opened 2025-09-19
[BUG]: unable to connect to a Streamable mcp server


omerbrandis:
### How are you running AnythingLLM?

Docker (local)

### What happened?

(these are my first steps with mcp servers and anyting-llm... )

using anything-llm desktop version v1.8.5-r2

my anythingllm_mcp_servers.json : 
{
  "mcpServers": {
	"MyServer" : { "type": "streamable",  
      		      "url": "http://www....."
		    }
	}
}

the ui shows that MyServer is stopped,
when I try to start it , I get the following message : 
Failed to start MCP server: Upfront [NO_CODE] Error POSTing to endpoint (HTTP 405)

I can see that a GET request was sent to the url. 
but as far as I can tell , for streamable http, there need not be support for GET on the url,
Only for post.
( i've managed to connect to Myserver with another ai desktop client ....).

can someone confirm that this is an erorr in anything llm , 
or else add a link to the relevant mcp documentation/spec that addresses this issue?

PR #4479, opened 2025-10-03
Issue solved: In Anything-LLM Desktop, streamable MCP servers fail to…

thakare18:
resolves issue/Bug id #4411

What is in this change?

This PR addresses an issue where streamable MCP servers fail to start because the client sends a `GET` request first. These servers only accept `POST` requests, which leads to an HTTP 405 (Method Not Allowed) error.

* **Fix:** The client is updated to start directly with a `POST` request, bypassing the unnecessary `GET` request that was causing the error.


thakare18:
Thanks for the review The fix ensures that the client now initiates with a POST request as required by the MCP server. 
Please let me know if any additional testing or documentation updates are needed before approval..

timothycarambat:
I dont know why this was not closed a while ago - this appears to only add redundant comments to a file?

Issue #4491, opened 2025-10-04
[UX]: Message for users without workspace access


angelplusultra:
## Problem

<img width="1536" height="949" alt="Image" src="https://github.com/user-attachments/assets/83d13333-8227-471e-9a97-8a9fbc7a5f76" />

When a default user logs in but is not added to any workspaces, they only see the onboarding text. This creates confusion because:
	•	There’s no clear indication that they lack workspace access.
	•	Users don’t know what steps to take next.
	•	It may appear as though the app is broken or incomplete.

## Expected Behavior

If a user has no workspace access, the interface should clearly communicate this and provide guidance.

### Suggested Solution

• Display a dedicated message (e.g., “You are not currently added to any workspaces. Please contact your admin or manager to be granted access.”)
• Optionally provide a button or link for users to request access (if applicable).

### Why This Matters

Improves the onboarding and user experience for new or default users by:
	•	Reducing confusion and frustration.
	•	Preventing unnecessary support requests.
	•	Making the system’s state and next steps clear.

nehaaprasad:
Hello! @angelplusultra 
 fixed it by adding a clear message that tells users they need workspace access and how to contact their admin - no more confusing onboarding text for users without permissions
let me know

PR #4495, opened 2025-10-05
fix: Show clear message for users without workspace access

naaa760:
- when default users log in but aren't added to any workspaces, they previously saw confusing onboarding text. 
- This PR adds a dedicated component that clearly explains they need workspace access and provides guidance on contacting their admin/manager.

fixes: #4491

**Changes:**
- New `NoWorkspaceAccess` component with helpful messaging
- `useWorkspaceAccess` hook to check user permissions  
- Updated Main component routing logic
- Added translation keys for the new messages

**Impact:** Reduces user confusion and support requests by providing clear guidance instead of misleading onboarding text.

### DIFF
--- a/frontend/src/components/NoWorkspaceAccess/index.jsx
+++ b/frontend/src/components/NoWorkspaceAccess/index.jsx
@@ -1,0 +1,60 @@
+import React from "react";
+import { useTranslation } from "react-i18next";
+import { User, EnvelopeSimple } from "@phosphor-icons/react";
+import { userFromStorage } from "@/utils/request";
+import paths from "@/utils/paths";
+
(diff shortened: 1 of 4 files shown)


timothycarambat:
We are implementing a whole new design for this issue. I appreciate you taking this work on, but ultimately we cannot use this for now as we are building this internally. 

If you do spot an issue you would like to attack, do please ping me as there could be work occurring for this internally within the core team or has other requirements that are not explictly listed in the issue. I dont want you to ever have to do any work that we cannot get merged because of some external factor you were not aware of.

nehaaprasad:
@timothycarambat 
sure, I will make sure for the next time!

PR #4197, opened 2025-07-24
Use Workspace System Prompt for agent invocations as well

yaronyam:
### Pull Request Type
- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues
resolves #1979
connect #4160

### What is in this change?
Add Workspace Agent Settings toggle - `Use Workspace System Prompt` - for using workspace chat configured system prompt for agent invocation as well

### Additional Information
image

### Developer Validations

- [x] I ran `yarn lint` from the root of the repo & committed changes
- [x] Relevant documentation has been updated
- [x] I have tested my code functionality
- [x] Docker build succeeds locally

PR #4526, opened 2025-10-11
Agent workspace system prompt with variable expansion

### LINKED ISSUE #4160: [FEAT]: Ability to turn off @agent directive or inject system prompt into @agent workflow
## Context
I've been using AnythingLLM for about 6 months now and have been superbly impressed. 

I limit the topic of discussion using the workspace system prompt so that discussion remains relevant and the endpoint isn't used for anything other than its purpose.

In testing, we have discovered the following problem...

## Problem:
When a user enters the `@agent` directive, the system prompt is no longer obeyed and so the endpoint can go rogue and discuss anything it wants. 

## Requests
It would be ace if either: 
- We could disable the `@agent` directive from specific workspaces, **or**
- We could inject the workspace system prompt into the agent system prompt to maintain the system-prompt level guardrails.

### LINKED ISSUE #4521: [FEAT]: Use system prompt for @agent

If a user sets a custom system prompt, it should propagate to the agent prompt as well and also do variable expansion with any custom variables used.

Relevant issues:
#1979
#4160

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4521 
resolves #4160 
closes #4197 

### What is in this change?

- Uses workspace system prompt in @agent mode
- Compatible with system prompt variables and user custom variables

### Developer Validations

### DIFF
--- a/collector/__tests__/utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.test.js
+++ b/collector/__tests__/utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.test.js
@@ -1,3 +1,4 @@
+process.env.STORAGE_DIR = "test-storage"; // needed for tests to run
 const { YoutubeTranscript } = require("../../../../../utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.js");
 
 describe("YoutubeTranscript", () => {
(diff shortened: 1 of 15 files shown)


antoniohof:
That doesnt seem to work. The Agent still doesnt use the system prompt

timothycarambat:
@antoniohof you will need to provide any kind of reproducible steps, or something to trace how/why it seems not to work for you

antoniohof:
it's only with the first message - which is weird. I'm using Ollama - maybe that's why? I will follow up on that.

timothycarambat:
Using ollama I am manually logging the full payload to Ollama and I see it set as the `system` prompt - that being said it may be lost in tool calling? What test case are you using to see if I can repro

PR #4540, opened 2025-10-14
Update Dockerfile

funktionslust:
Ensure everything in /app dir is owned and traversable by the runtime user


timothycarambat:
Is this required? The app builds work already with no changes so I dont see why this is required at this time?

PR #4545, opened 2025-10-15
fix(server): correct Qdrant batching logic for large uploads

### LINKED ISSUE #4544: [BUG]: Qdrant integration fails on document upload with more than 500 chunks

All versions

### What happened?

When uploading a large document that gets split into more than 500 chunks (the default Qdrant batch limit), the embedding process fails. The `addDocumentToNamespace` function in `server/utils/vectorDbProviders/qdrant/index.js` correctly splits the vectors into chunks but then incorrectly attempts to upsert all vectors at once, rather than iterating through the prepared chunks.

**To Reproduce**
1. Set up AnythingLLM with Qdrant as the vector database.
2. Use an embedding model with a small chunk size (e.g., 1024).
3. Upload a large document that will generate more than 500 vector chunks.
4. Observe the server logs for an error during the `client.upsert` operation.

**Expected behavior**
The document should be successfully embedded by sending the vector chunks to Qdrant in multiple batches, each no larger than the batch limit.

**Additional context**
The issue is in the loop that processes `toChunks(vectors, 500)`. The fix is to use the `chunk` variable inside the loop to create the batch for `client.upsert`, rather than using the full `submission` object.

aeromech-1:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4544

### What is in this change?

This change corrects the batching logic in the Qdrant vector provider (index.js).

Previously, the code would attempt to upsert all generated vector chunks in a single request. This caused an error when the number of chunks exceeded Qdrant's batch limit (e.g., >500).

The fix ensures the code properly iterates through the pre-defined chunks and sends them in smaller, valid batches, allowing for the successful embedding of documents with a large number of chunks.

### Additional Information

This issue was discovered when a less aggressive chunking strategy was used on a document, increasing the chunk count from ~200 to ~2000, which triggered the bug.

### Developer Validations

### DIFF
--- a/collector/__tests__/utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.test.js
+++ b/collector/__tests__/utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.test.js
@@ -1,3 +1,4 @@
+process.env.STORAGE_DIR = "test-storage"; // needed for tests to run
 const { YoutubeTranscript } = require("../../../../../utils/extensions/YoutubeTranscript/YoutubeLoader/youtube-transcript.js");
 
 describe("YoutubeTranscript", () => {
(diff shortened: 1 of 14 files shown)


timothycarambat:
Changed the original `O(3N)` complexity of the batch creation to be `O(N)`. Otherwise LGTM

PR #4553, opened 2025-10-16
Add batch embedding support for Ollama embedding provider

### LINKED ISSUE #4529: [FEAT]: OllamaEmbedder Support Batched or Parallel Embeddings to Improve Performance
Hi there, thank you for building such a powerful tool with anythinglmm!

While using the OllamaEmbedder for document embedding, I noticed that the current implementation processes text chunks sequentially (using for...of + await). This results in very poor performance when handling large documents, even when the backend Ollama instance is running on high-end hardware with multiple GPUs.

📌 Problem Description
Currently, in OllamaEmbedder.embedChunks():

Each chunk is sent in a separate /api/embeddings request
Only one chunk is processed per request
It does not leverage Ollama’s built-in support for batched input via the input: string[] field
This leads to extremely long processing times (e.g., 30+ minutes for 10k+ chunks)
✨ Suggested Improvement
Please consider adding support for one of the following optimizations:

✅ Option 1: Use Ollama’s Batched Embeddings API (Recommended)
Ollama’s /api/embeddings endpoint supports passing an array of strings:

```
await client.embeddings({
  model: "qwen3-embedding:8b",
  input: ["text1", "text2",  #...],
})
```
This drastically reduces network overhead and fully utilizes GPU parallelism.

✅ Option 2: Concurrent Processing (Fallback)
If batched input is not compatible with certain models, please support concurrent requests using Promise.all() with a configurable concurrency limit (e.g., maxConcurrentChunks) to prevent OOM issues.

🔧 Example Code (for reference)
```
async embedChunks(textChunks) {
  const response = await this.client.embeddings({
    model: this.model,
    input: textChunks, // Batched input
    options: { num_ctx: this.embeddingMaxChunkLength }
  });
  return response.embeddings;
}
```

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4529 

### What is in this change?

- Switched `embed()` call to use `input` param vs the old `prompt` param to support batch processing
- Adds new advanced options item to allow for adjusting batch size in Ollama provider (default is 1)
- Updated logging to show batch sizes and progress

### Additional Information

- Ran into this issue that is on Ollama's end: https://github.com/ollama/ollama/issues/6094
- Reduce `Max embedding chunk length` in Ollama settings to get past this error
- Had to change `nomic-embed-text` which usually has context of 2048 to 1024

### Developer Validations


server/utils/EmbeddingEngines/ollama/index.js
```diff
@@ -64,29 +65,45 @@ class OllamaEmbedder {
...
+      const batch = textChunks.slice(i, i + this.maxConcurrentChunks);
+      currentBatch++;
+
       try {
-        const res = await this.client.embeddings({
+        // Use input param instead of prompt param to support batch processing
+        const res = await this.client.embed({
           model: this.model,
-          prompt: chunk,
+          input: batch.length === 1 ? batch[0] : batch,
```
timothycarambat: If `input` can take an array or a single string, why do we check if it is an array first? Additionally, if batch is a string length can be incorrect and embed just a single char - since there is not Array.isArray check prior this could happen?

If it works with both, I say we just pass it as is since it can function with a string or an array of string.s

timothycarambat (changes requested):
See question/feedback on `input` param

server/utils/EmbeddingEngines/ollama/index.js
shatfield4: Removed extra complexity and just return batch as input since Ollama can take in a string or string[] in the input field. Tested thoroughly to make sure no difference with this change.

PR #4568, opened 2025-10-21
fix: Use eval_duration for output TPS calculations in Ollama LLM provider

### LINKED ISSUE #4567: [BUG]: Inaccurate duration metric from Ollama

Docker (local)

### What happened?

I'm setting up analytics in my fork and noticed tps/duration not lining up with expectations (especially when  not streaming). In my tests with Phi4:14b streaming was about 60tps, but non-streaming was about 15tps. In either case, metrics are capturing duration and using that (with network latency + time to first token) to calculate outputTps, which is misleading.

Instead of using duration for the calculation, it should be done with eval_duration, as this is how output speed is generally calculated.

jonathanortega2023:
### Pull Request Type
- [x] 🐛 fix

### Relevant Issues

resolves #4567

## What is in this change?

- Updated `LLMPerformanceMonitor` to support provider duration timing for more accurate output TPS calculations
- `measureAsyncFunction` accepts `duration` from `output.usage.duration`
- Modified `measureStream` to accept `duration` from `reportedUsage.duration` if provided by the LLM provider  
- Updated Ollama provider to use `eval_duration` for more accurate TPS calculation

This is a generic enhancement that any provider can leverage to supply more accurate timing information, while maintaining backward compatibility with providers that don't supply duration metrics.

### Developer Validations

### DIFF
--- a/server/utils/AiProviders/ollama/index.js
+++ b/server/utils/AiProviders/ollama/index.js
@@ -263,7 +263,6 @@ class OllamaAILLM {
               prompt_tokens: res.prompt_eval_count,
               completion_tokens: res.eval_count,
               total_tokens: res.prompt_eval_count + res.eval_count,
-              // Override duration with Ollama's eval_duration for more accurate outputTps
               duration: res.eval_duration / 1e9,
             },
           };
@@ -352,7 +351,6 @@ class OllamaAILLM {
           if (chunk.done) {
             usage.prompt_tokens = chunk.prompt_eval_count;
             usage.completion_tokens = chunk.eval_count;
-            // Override duration with Ollama's eval_duration for more accurate outputTps
             usage.duration = chunk.eval_duration / 1e9;
             writeResponseChunk(response, {
               uuid,
--- a/server/utils/helpers/chat/LLMPerformanceMonitor.js
+++ b/server/utils/helpers/chat/LLMPerformanceMonitor.js
@@ -49,10 +49,7 @@ class LLMPerformanceMonitor {
       const start = Date.now();
       const output = await func; // is a promise
       const end = Date.now();
-      const calculatedDuration = (end - start) / 1000;
-      // Use duration from output.usage if provided (for providers that have more accurate timing)
-      // otherwise use the calculated duration from function start/end times
-      const duration = output?.usage?.duration ?? calculatedDuration;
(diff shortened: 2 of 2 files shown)


server/utils/helpers/chat/LLMPerformanceMonitor.js
```diff
@@ -88,7 +89,12 @@ class LLMPerformanceMonitor {
 
       stream.metrics.total_tokens =
         stream.metrics.prompt_tokens + (stream.metrics.completion_tokens || 0);
-      stream.metrics.outputTps = stream.metrics.completion_tokens / duration;
+
+      // Use eval_duration if provided (for providers like Ollama that report it)
+      // otherwise fall back to total request duration
+      stream.metrics.outputTps =
+        stream.metrics.completion_tokens /
+        (stream.metrics.eval_duration || duration);
```
timothycarambat: eval duration is specific to ollama and we should not be modifying the core utility to accommodate this unless there is no way around this.

server/utils/AiProviders/ollama/index.js
```diff
@@ -282,7 +283,10 @@ class OllamaAILLM {
         prompt_tokens: result.output.usage.prompt_tokens,
         completion_tokens: result.output.usage.completion_tokens,
         total_tokens: result.output.usage.total_tokens,
-        outputTps: result.output.usage.completion_tokens / result.duration,
+        outputTps:
```
timothycarambat, line 286: above this, we get a response object with a custom property `eval_duration` which is not actually eval_duration since we converted it to seconds

When the stream ends we do
```
 stream.endMeasurement = (reportedUsage = {}) => {
      const end = Date.now();
      const duration = (end - stream.start) / 1000;

      // Merge the reported usage with the existing metrics
      // so the math in the metrics object is correct when calculating
      stream.metrics = {
        ...stream.metrics,
        ...reportedUsage,
      };

      stream.metrics.total_tokens =
        stream.metrics.prompt_tokens + (stream.metrics.completion_tokens || 0);
      stream.metrics.outputTps = stream.metrics.completion_tokens / duration;
      stream.metrics.duration = duration;
      return stream.metrics;
    };
```
So in theory, to limit this just to ollama, we could overwrite `results.output.usage.duration` if `eval_duration` is set?

PR #4569, opened 2025-10-21
feat: add Agent Mode toggle in WorkspaceChat

Bilal2005seven:
This PR adds a new Agent Mode toggle feature in the chat interface to improve the usability of tool-calling interactions.

Currently, users must manually prepend @agent to each message to trigger Agent Mode. This update introduces a toggle switch that automates that process.

Key Highlights

🧠 New Toggle Switch near the chat input bar.

🔁 When enabled, all outgoing user messages automatically prepend @agent.

💬 When disabled, messages are sent as normal.

💾 Toggle state persists for the current session (can be extended later via localStorage).

⚙️ Integrated seamlessly into WorkspaceChat/index.jsx.

This enhancement makes interacting with Agents faster and more intuitive, reducing typing friction and improving UX consistency.

### DIFF
--- a/frontend/src/components/WorkspaceChat/index.jsx
+++ b/frontend/src/components/WorkspaceChat/index.jsx
@@ -18,6 +18,16 @@ export default function WorkspaceChat({ loading, workspace }) {
   const [history, setHistory] = useState([]);
   const [loadingHistory, setLoadingHistory] = useState(true);
 
+  // --- Agent Mode Toggle State ---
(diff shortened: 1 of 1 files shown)


timothycarambat:
This is the same as #3483, and there is a reason it is not merged; there is a lot of considerations around doing this. Also, your PR would not work

tgkap:
What are the considerations around doing this? I understand it can use more tokens.

PR #4621, opened 2025-11-06
fix: unhelpful error message for invite link user creation

### LINKED ISSUE #4618: [FEAT]: Unhelpful hint when creating a user via invite link

When sending an invite link and trying to register with an email you get `Error: Could not create user.` instead of a helpful hint. This leaves the user unsure on what to do. It should instead display, what is displayed in the logs

`Username must only contain lowercase letters, periods, numbers, underscores, and hyphens with no spaces`

nehaaprasad:
### Pull Request Type
- [x] 🐛 fix

### Relevant Issues
resolves #4618

### What is in this change?
Fixed invite endpoint to show actual username validation error instead of generic "Could not create user" message. Users now see: "Username must only contain lowercase letters, periods, numbers, underscores, and hyphens with no spaces."

### Developer Validations

### DIFF
--- a/collector/package.json
+++ b/collector/package.json
@@ -12,7 +12,8 @@
   "scripts": {
     "dev": "cross-env NODE_ENV=development nodemon --ignore hotdir --ignore storage --trace-warnings index.js",
     "start": "cross-env NODE_ENV=production node index.js",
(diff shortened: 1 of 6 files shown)


shatfield4 (approved):
LGTM

PR #4626, opened 2025-11-06
Check for invalid VecDB value, log/toast error

### LINKED ISSUE #4606: [BUG]: VecDB page doesn't catch erroneous env selections

Docker (local)

### What happened?

When an invalid VecDB selection is written in env, the page fails to render and has an ambiguous error:
```
Uncaught TypeError: Cannot read properties of undefined (reading 'logo')
    at GeneralVectorDatabase (index.jsx:296:48)
```

Might be ideal behavior to fallback on LanceDB in the selector.

In server/.env, set VECTOR_DB to anything erroneous.

jonathanortega2023:
### Pull Request Type
- [x] 🐛 fix

### Relevant Issues
resolves #4606

### What is in this change?
Checks if vecDB selection from ENV is valid, logs and toasts when invalid and falls back onto LanceDB as the selection with unsaved changes.

### Developer Validations

### DIFF
--- a/frontend/src/pages/GeneralSettings/VectorDatabase/index.jsx
+++ b/frontend/src/pages/GeneralSettings/VectorDatabase/index.jsx
@@ -188,7 +188,17 @@ export default function GeneralVectorDatabase() {
     },
   ];
 
-  const selectedVDBObject = VECTOR_DBS.find((vdb) => vdb.value === selectedVDB);
+  const selectedVDBObject =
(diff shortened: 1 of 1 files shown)


timothycarambat:
This "bug" is only possible by someone opening the ENV file directly, typing in an incorrect value, and then opening the app. It is not possible to select an invalid provider from the UI. Additionally, this only patches the UI crash and not the underlying issue since even if the UI is right we should failover to LanceDB on the backend as well - not just show that on the frontend.

The user can then wind up seeing lanceDB in the frontend due to an invalid value, but the backend upserts still fail!

PR #4667, opened 2025-11-20
[FEAT]: Allow Multiple Connections to the Same Type of LLM Provider 

### LINKED ISSUE #4493: [FEAT]: Allow Multiple Connections to the Same Type of LLM Provider

If I'm not mistaken, AnythingLLM currently seems to allow only one instance of a given LLM Provider to be connected at a time. If I want to connect multiple instances, I have to change it in the settings. One use case for allowing multiple connections to the same type of LLM Provider is: if I have two Ollama servers, I might want one workspace or thread to connect to one of them, while other workspaces or threads use the other. This way, they can perform different tasks simultaneously.

socialviolation:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4493

### What is in this change?

Firstly, I apologise for the size of this one. I appreciate that it is a bit of a paradigm shift.

This PR implements support for multiple LLM provider connections per workspace, resolving issue #4493.

Problem: AnythingLLM currently restricts users to a single instance per LLM provider type (e.g., one Ollama server, one LiteLLM proxy). Users must manually reconfigure settings to switch between different instances, creating friction for teams managing
multiple deployments.

Solution: Introduces a connection-based architecture that allows:
  - Admins to pre-configure multiple LLM connections (e.g., multiple Ollama servers with different API keys/URLs)
  - Workspace managers to select specific connections for chat and agent models
  - Different workspaces to simultaneously use different instances of the same provider type
screenshot-2025-11-20_21-30-07

This is especially useful for enterprise use cases, as I use LiteLLM to:
- abstract api keys away from teams
- assign cost controls to teams
- assign/hide MCP/tools to specific teams
- Have the option of using multiple accounts from the same provider, meaning rate limits from one team don't affect all 'open ai' teams.

### Developer Validations


timothycarambat:
I understand the want here, we know it is a desired feature but as you noted, its a pretty massive paradigm shift in how we manage the selected LLM. From a cursory review, there are a ton of unrelated changes to the core of what this PR should be:
- Testing setup config as opposed to just isolated unit tests
- LiteLLM unsigned certificate changes

Some other notes:
- This does not touch the onboarding screens, but I presume there must be some logic for migration or backward compatibility. This might work out of the box for a new install, but we have about 5M installs or more that might wind up with a broken instance if this feature is not backward compatible to where they can pull in and be instantly up to date.
- UI looks OK from what I can currently see

Either way, this is a really large PR because it has to be, but in its current state, this is such a massive change it is going to need some love for sure. Ill dive into this in the future and see what we can do

socialviolation:
Big thanks @timothycarambat! Have addressed a few of the easy comments you left.

For migration testing - I'm happy to assist. If you had a checklist of scenarios I could follow that would be super handy.

timothycarambat:
closing for same reason as https://github.com/Mintplex-Labs/anything-llm/pull/6178#issuecomment-5591464502 but if we do happen to use any code from this we would credit via an authorship tag.

PR #4680, opened 2025-11-24
Store Chat Prompt Input Value in Local Storage

### LINKED ISSUE #4604: [FEAT]: Chat box should store to localstorage/similar the current message

If your typing something, then jump to settings or even another thread I think, the current message isnt retained. This should be remembered. Its a very bad UX atm.

angelplusultra:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4604 

### What is in this change?

This PR implements a small feature to save the state of the workspace chat input value in local storage and will be applied as the initial state for all workspace chats if truthy. It also clears this value when in multi-user mode and the current users signs out. It also clears this value in every other auth clearing situation.

### Developer Validations


angelplusultra:
@timothycarambat Let me know if you would like each chat thread to have it's own stored prompt input state, I can refactor to do that instead if you prefer.

angelplusultra:
@timothycarambat 

• “Since this is a convenience feature…”
 I reverted most of the localStorage removal calls that were happening in the more aggressive auth flows. The value now clears only when `actions.unsetUser()` is invoked or when a user explicitly logs out from the avatar dropdown.

• “This input should also be workspace/thread specific.”
I’ll refactor it so the stored input is scoped properly. Right now it uses a single global state, but I can switch it to a workspace-level or thread-level key—whichever you prefer. My plan is to store a simple JSON map keyed by conversation/thread ID with the prompt value as the entry. Reading/writing stays O(1).

• “How do we clear the session variable when the input is finally sent?”
This is already handled automatically. The effect listening to the input state clears the value whenever a message is sent (since the input itself resets), so the localStorage entry resets accordingly.

• “It may make sense to move all of this to a util…”
No strong opinion either way. The actual code footprint is small (just two lightweight useEffect hooks), but I understand wanting to keep the prompt input area as clean as possible given how complex that part of the app already is.

If you want me to pull it out, I’d recommend creating a small custom hook (`usePromptStorage`) rather than a context provider. Since we’re not trying to solve prop-drilling here, a hook keeps it simpler and avoids unnecessary abstraction.

Let me know which direction you’d prefer and I’ll implement it.

timothycarambat:
> I’ll refactor it so the stored input is scoped properly. Right now, it uses a single global state, but I can switch it to a workspace-level or thread-level key—whichever you prefer. My plan is to store a simple JSON map keyed by conversation/thread ID with the prompt value as the entry. Reading/writing stays O(1).

Let's do this for sure, as long as we can manage and look up the dict, we should be fine.

> • “It may make sense to move all of this to a util…”
No strong opinion either way. The actual code footprint is small (just two lightweight useEffect hooks), but I understand wanting to keep the prompt input area as clean as possible given how complex that part of the app already is.

Once the first point is implemented, I think the footprint may be larger, since now we are CRUD'ing records in a local storage item where we have to stringify/parse JSON over and over. Once the above is implemented, it is worth looking into if this is breaking out into a util - just so management of this functionality is easier to parse. The prompt input is already so complex so making these kinds of things self-contained at least makes the code more readable.

angelplusultra:
@timothycarambat 

Let me know what you think of this approach. 

- Prompt input values are now saved on the thread level (https://github.com/Mintplex-Labs/anything-llm/pull/4680/commits/1da80037fa95c89620889acb0063a8fff180ad39).
- Logic related to this has been mostly abstracted to a custom hook (https://github.com/Mintplex-Labs/anything-llm/pull/4680/commits/1da80037fa95c89620889acb0063a8fff180ad39).
- Writing to local storage has been debounced 500ms (https://github.com/Mintplex-Labs/anything-llm/pull/4680/commits/4228c6b3b0c73ada76459775743b0dce3e30d3fc)

frontend/src/hooks/usePromptInputStorage.js
```diff
@@ -0,0 +1,85 @@
...
+  useEffect(() => {
+    // Get the user prompt input map from localStorage
+    let promptInputMap = localStorage.getItem(USER_PROMPT_INPUT_MAP);
+
+    // Attempt to deserialize the prompt input map
+    try {
+      promptInputMap = promptInputMap ? JSON.parse(promptInputMap) : {};
+    } catch (error) {
+      promptInputMap = {};
+    }
```
timothycarambat: Just use `safeJSONParse(promptInputMap, {})` - does the same thing. Should be in `request` in utils.

frontend/src/hooks/usePromptInputStorage.js
```diff
@@ -0,0 +1,85 @@
...
+    const userPromptInputValue = promptInputMap[threadSlug ?? workspaceSlug];
+    if (userPromptInputValue) {
+      setPromptInput(userPromptInputValue);
+      // Notify parent component so message state is synchronized
+      onChange({ target: { value: userPromptInputValue } });
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, []);
+
+  // Set the user prompt input map to localStorage (debounced by 300ms)
```
timothycarambat: You dont need to be so verbose with the comments, lots of these things are self-explanatory. Only if some logic is unclear or is _really_ important in the code that is not instantly readable should be we comment. Like ln:44 is a reasonable comment. This line is pretty obvious about what is going on here.

_actually i just noticed the comment says 300ms but debounce is 500ms - so its not even correct!_

frontend/src/utils/constants.js
```diff
@@ -8,6 +8,7 @@ export const COMPLETE_QUESTIONNAIRE = "anythingllm_completed_questionnaire";
 export const SEEN_DOC_PIN_ALERT = "anythingllm_pinned_document_alert";
 export const SEEN_WATCH_ALERT = "anythingllm_watched_document_alert";
 export const LAST_VISITED_WORKSPACE = "anythingllm_last_visited_workspace";
+export const USER_PROMPT_INPUT_MAP = "anythingllm_user_prompt_input_map";
```
timothycarambat, line 11: +1

frontend/src/hooks/usePromptInputStorage.js
```diff
@@ -0,0 +1,85 @@
...
+ * }
+ * ```
+ *
+ * @param {Object} props
+ * @param {Function} props.onChange - Callback invoked when restoring saved value, receives `{ target: { value: string } }`
+ * @param {string} props.promptInput - Current prompt input value to sync
+ * @param {Function} props.setPromptInput - State setter function for prompt input
+ * @returns {void}
+ */
+export default function usePromptInputStorage({
```
timothycarambat, line 28: 👍

timothycarambat (changes requested):
Looks good, just some minor changes to use some existing widely used utils for JSON parsing. Additionally, lets remove some verbosity in the comments so its code-readable for the easy stuff

frontend/src/hooks/usePromptInputStorage.js
angelplusultra: Yeah i changed it last minute to 500ms my bad I'll take note of how comments should be used sparingly.

frontend/src/hooks/usePromptInputStorage.js
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4680/commits/ecf7c542779ce2c1d867b9474a472dffa3b33524

frontend/src/hooks/usePromptInputStorage.js
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4680/commits/ecf7c542779ce2c1d867b9474a472dffa3b33524

PR #4703, opened 2025-12-02
Feat/cohere agent implementation

### LINKED ISSUE #4490: [FEAT]: Cohere `@agent` implemention

Cohere, while available for basic chat, cannot be used in agents. We should allow cohere models to be used for `@agent` since they are plenty capable now.

- Will need to use `Untooled`
- Has its own SDK that will make its implementation of `complete` and `stream` have to have provider class overrides to function.
- Has a specific messaging format as well.

17ColinMiPerry:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4490

### What is in this change?

Adds `@agent` support for Cohere. Uses v1 SDK to be consistent with the LLM provider implementation.

`#convertChatHistoryCohere`: converts standard messages to the format expected by Cohere.
`#handleFunctionCallStream/Chat`: Complies to Cohere API by passing message + chat_history
`streamingFunctionCall`: Overridden from `UnTooled` to handle Cohere's stream event format
`stream/complete`: Overrides to handle Cohere's SDK response structure

### Additional Information

Currently, we don't have the `@langchain/cohere` package, so there is no LangChain integration.
https://docs.cohere.com/v1/reference/about

### Developer Validations

### DIFF
--- a/server/utils/agents/aibitat/providers/cohere.js
+++ b/server/utils/agents/aibitat/providers/cohere.js
@@ -9,7 +9,8 @@ class CohereProvider extends InheritMultiple([Provider, UnTooled]) {
   model;
 
   constructor(config = {}) {
-    const { model = process.env.COHERE_MODEL_PREF || "command-r-08-2024" } = config;
+    const { model = process.env.COHERE_MODEL_PREF || "command-r-08-2024" } =
+      config;
     super();
     const client = new CohereClient({
       token: process.env.COHERE_API_KEY,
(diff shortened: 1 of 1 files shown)


17ColinMiPerry:
The v1 SDK was used to be consistent with the rest of the repository. However, it seems that v1 is largely being deprecated, and v2 is OpenAI-like. Would it be meaningful to convert this part of the codebase (and the Cohere LLM provider) to the v2 SDK API at this time?

timothycarambat:
This was mostly there. Not sure how you tested it but it did not work at all for calling tools. Tried many models and many prompts, and it wouldn't call a basic tool. It was streaming in agent mode, but it refused any tool call.

Had to [rework the flow](https://github.com/Mintplex-Labs/anything-llm/pull/4703/commits/c94c657b68c56fdb044a0aa3ef4719a7d0c728f4) for it to be able to actually see the tools and call them. Now seems to work fine.

17ColinMiPerry:
Thank you for the changes, I looked through the reworked flow. I tested a handful of models, and most refused to call tools, but at least one worked (c4ai-aya-expanse-32b), and so I chalked it up to poor model performance. Clearly there were other underlying issues with my pr, so thanks for the help!

PR #4744, opened 2025-12-09
Migrate Azure OpenAI Integration To v1 API | Enable Streaming for Reasoning Models in Azure OpenAI Basic Inference Provider

### LINKED ISSUE #3340: [FEAT]: Add support for 2024-12-01-preview and 2025-01-01-preview API versions for AOAI
Added support for the latest API versions for Azure OpenAI, 2024-12-01-preview and 2025-01-01-preview.  The specifics for which are detailed here: https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/reasoning?tabs=python-secure. 

### Rationale: 
The latest reasoning models are only supported under the two previous API specifications, namely the o3 and o1 variations.

angelplusultra:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #3340 

### What is in this change?

This PR includes the following updates:
	1.	Replaces the `AzureOpenAI` class with the `OpenAI` class.
	2.	Migrates our Azure OpenAI integration to the new v1 API. This resolves the issue where the older static API version (`2024-12-01-preview`) failed to expose reasoning models.
	3.	Removes a guard in the Azure OpenAI basic inference provider that previously prevented reasoning models from supporting streamable completions.
	4.	Adds a code comment in `server/utils/AiProviders/azureOpenAi/index.js` clarifying the purpose of the `AZURE_OPENAI_MODEL_TYPE` environment variable.
https://github.com/Mintplex-Labs/anything-llm/blob/da55ca1c8fd4cdd418f9f84303ef22781c03150c/server/utils/AiProviders/azureOpenAi/index.js#L23-L28
	5.	Adds a frontend tooltip in the Azure OpenAI LLM provider settings explaining the importance of the Model Type field.

image

### Additional Information

Since we have no way to determine the underlying model behind a user’s Azure OpenAI deployment name, it is crucial that they set the `AZURE_OPENAI_MODEL_TYPE` environment variable to "reasoning". Without this, we may pass invalid parameters for reasoning models (e.g., `temperature`) into the completion request, resulting in chat errors.

### Developer Validations

### DIFF
--- a/.github/workflows/dev-build.yaml
+++ b/.github/workflows/dev-build.yaml
@@ -6,7 +6,7 @@ concurrency:
 
 on:
   push:
-    branches: ['maint/upgrade-docker-base-image'] # put your current branch to create a build. Core team only.
+    branches: ['chore/bump-express-js'] # put your current branch to create a build. Core team only.
     paths-ignore:
       - '**.md'
       - 'cloud-deployments/*'
--- a/collector/package.json
+++ b/collector/package.json
@@ -21,7 +21,7 @@
     "cors": "^2.8.5",
     "dotenv": "^16.0.3",
     "epub2": "git+https://github.com/Mintplex-Labs/epub2-static.git#main",
-    "express": "^4.18.2",
+    "express": "^4.21.2",
     "fluent-ffmpeg": "^2.1.2",
     "html-to-text": "^9.0.5",
     "ignore": "^5.3.0",
--- a/collector/yarn.lock
+++ b/collector/yarn.lock
@@ -697,25 +697,7 @@ bmp-js@^0.1.0:
   resolved "https://registry.yarnpkg.com/bmp-js/-/bmp-js-0.1.0.tgz#e05a63f796a6c1ff25f4771ec7adadc148c07233"
   integrity sha512-vHdS19CnY3hwiNdkaqk93DvjVLfbEcI8mys4UjuWrlX1haDmroo8o4xCzh4wD6DGV6HxRCyauwhHRqMTfERtjw==
 
-body-parser@1.20.3:
-  version "1.20.3"
-  resolved "https://registry.yarnpkg.com/body-parser/-/body-parser-1.20.3.tgz#1953431221c6fb5cd63c4b36d53fab0928e548c6"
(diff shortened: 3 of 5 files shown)


server/utils/AiProviders/azureOpenAi/index.js
```diff
@@ -9,19 +9,23 @@ const {
...
     if (!process.env.AZURE_OPENAI_KEY)
       throw new Error("No Azure API key was set.");
 
-    this.apiVersion = "2024-12-01-preview";
-    this.openai = new AzureOpenAI({
+    this.openai = new OpenAI({
       apiKey: process.env.AZURE_OPENAI_KEY,
-      apiVersion: this.apiVersion,
-      endpoint: process.env.AZURE_OPENAI_ENDPOINT,
+      baseURL: `${process.env.AZURE_OPENAI_ENDPOINT}/openai/v1`,
```
timothycarambat: Given the user could paste in _anything_ for the URL we should do them a favor incase they decide to paste in `http://my.auzre.deployment/profile/account/model/gpt-4o-deployment/openai/v1/models....` since the Azure portal does not always show such a base URL.

Since we know this should _just_ be the hostname create a `static` method on this class to validate and return the complete URL

```nodejs
try {
const url = new URL(process.env.AZURE_OPENAI_ENDPOINT)
url.pathname = /openai/v1
return url.toString()
} catch ....
```

This way, no matter what they give us, we will just use the part of the input needed to resolve the URL the way it needs to be resolved to reach the v1 endpoint.

Then use this util static method in the Agent so you can re-validate there as well before building the client.

timothycarambat (dismissed):
See minor change for validation method on main provider.

server/utils/AiProviders/azureOpenAi/index.js
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4744/commits/bf6aacbd4715954370e69cf001ba512baa7be921

Added validator methods in both the basic inference provider and the agent provider.

image

PR #4757, opened 2025-12-10
Fix pagination bug in paperless-ngx data connector

### LINKED ISSUE #4753: [BUG]: 1.9.1 pull request 4121 - addition of paperless-ngx connector.  Paperless-ngx connector only fetches first page of documents (pagination not implemented)

Docker (local)

### What happened?

Paperless-ngx connector only fetches first page of documents (pagination not implemented)

Description

The new Paperless-ngx data connector introduced in v1.9.1 (https://github.com/Mintplex-Labs/anything-llm/pull/4121) only retrieves approximately 25 documents, even when the Paperless-ngx instance contains thousands of documents. This appears to be because the connector fetches /api/documents/ without handling pagination.

Steps to Reproduce

Configure the Paperless-ngx data connector with a valid instance URL and API token
Connect to a Paperless-ngx instance that contains more than 25 documents
Observe that only ~20-25 documents are available for import
Expected Behavior

All documents from the Paperless-ngx instance should be available for import, regardless of the total count.

Actual Behavior

Only the first page of results (~25 documents) is returned. Documents beyond the first page are not fetched.

Root Cause

The Paperless-ngx API returns paginated results by default. The response includes:

count: total number of documents
next: URL to fetch the next page
previous: URL to fetch the previous page
results: array of documents for the current page
The connector appears to only process results from the initial request without following the next URL to retrieve subsequent pages.

Suggested Fix

The fetchAllDocuments() method should iterate through all pages until next is null. Alternatively, use the page_size parameter to request more documents per page: /api/documents/?page_size=1000

Environment

AnythingLLM version: 1.9.1
Paperless-ngx version: (any)
Deployment: Docker

_No response_

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4753

### What is in this change?

- Fixes a bug where importing documents via the paperless-ngx data connector would only import up to 25 documents (1 page of data)
- Fixed by using the `next` url that is returned by the paperless-ngx API to fetch next page of data until it is `null` indicating there are no more pages to iterate over

### Developer Validations

### DIFF
--- a/collector/utils/extensions/PaperlessNgx/PaperlessNgxLoader/index.js
+++ b/collector/utils/extensions/PaperlessNgx/PaperlessNgxLoader/index.js
@@ -50,6 +50,8 @@ class PaperlessNgxLoader {
           if (!validResults.length) break;
 
           documents.push(...validResults);
+
+          if (data.next === nextUrl) break;
           nextUrl = data.next || null;
           page++;
         } catch (error) {


collector/utils/extensions/PaperlessNgx/PaperlessNgxLoader/index.js
```diff
@@ -26,19 +26,36 @@ class PaperlessNgxLoader {
...
+        documents.push(...(data.results || []));
+        nextUrl = data.next;
+        page++;
+      }
+
+      console.log(
+        `Fetched ${documents.length} documents from Paperless-ngx (Pages: ${
+          page - 1
+        })`
+      );
```
timothycarambat, line 70: This needs try/catch handler for bad requests like 429s or 404s or anything else that could go wrong so we can still break out the while loop

jhsmith409:
Thank you for the quick response on this.

collector/utils/extensions/PaperlessNgx/PaperlessNgxLoader/index.js
```diff
@@ -26,19 +26,48 @@ class PaperlessNgxLoader {
...
+
+          if (!response.ok) {
+            throw new Error(
+              `Failed to fetch documents from Paperless-ngx: ${response.status}`
+            );
+          }
+
+          const data = await response.json();
+
+          // Filter valid docs (throws if doc is invalid)
```
timothycarambat: This is not true. Filter will not throw here and the option-chain will not cause an error either unless the array object is null or non-object.

PR #4759, opened 2025-12-11
Refactor frontend legacy JSON.parse with safeJsonParse

### LINKED ISSUE #4758: [FEAT]: Remove all JSON.parse and replace with safeJsonParse util

In the frontend we have several legacy JSON.parse calls that can potentially throw error boundaries.

We need to refactor every instance of JSON.parse in the frontend to use safeJSONParse in utils/requests that also includes a fallback to the specific conditions' sensible failover value.

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4758 

### What is in this change?

- Refactors every instance of JSON.parse in the frontend to use safeJSONParse
- Includes a fallback to the specific conditions sensible failover value
- Prevents failed JSON.parse calls from throwing errors and causing the entire frontend to crash

### Developer Validations

### DIFF
--- a/frontend/src/pages/GeneralSettings/ChatEmbedWidgets/EmbedConfigs/EmbedRow/index.jsx
+++ b/frontend/src/pages/GeneralSettings/ChatEmbedWidgets/EmbedConfigs/EmbedRow/index.jsx
@@ -141,7 +141,7 @@ export default function EmbedRow({ embed }) {
 }
 
 function ActiveDomains({ domainList }) {
-  const domains = safeJsonParse(domainList, null) || [];
+  const domains = safeJsonParse(domainList, []);
   if (domains.length === 0) return <p>all</p>;
   return (
     <div className="flex flex-col gap-y-2">


frontend/src/components/SettingsSidebar/MenuOption/index.jsx
```diff
@@ -130,7 +131,7 @@ function useIsExpanded({
     if (hasVisibleChildren) {
       const storedValue = localStorage.getItem(storageKey);
       if (storedValue !== null) {
-        return JSON.parse(storedValue);
+        return safeJsonParse(storedValue, true);
```
timothycarambat: I think the default value for menu is collapsed, not expanded. Does this expand them all by default? it should be whatever it does now on an empty key.

frontend/src/pages/GeneralSettings/ChatEmbedWidgets/EmbedConfigs/EmbedRow/EditEmbedModal/index.jsx
```diff
@@ -53,9 +54,7 @@ export default function EditEmbedModal({ embed, closeModal }) {
               <ChatModeSelection defaultValue={embed.chat_mode} />
               <PermittedDomains
                 defaultValue={
-                  embed.allowlist_domains
-                    ? JSON.parse(embed.allowlist_domains)
-                    : []
+                  safeJsonParse(embed.allowlist_domains, null) || []
```
timothycarambat, line 57: Shouldnt this just be `safeJsonParse(embed.allowlist_domains, [])`

frontend/src/pages/GeneralSettings/ChatEmbedWidgets/EmbedConfigs/EmbedRow/index.jsx
```diff
@@ -141,20 +142,17 @@ export default function EmbedRow({ embed }) {
...
+    <div className="flex flex-col gap-y-2">
+      {domains.map((domain, index) => {
+        return (
+          <p key={index} className="font-mono !font-normal">
+            {domain}
+          </p>
+        );
+      })}
+    </div>
+  );
```
timothycarambat, line 156:
```
  const domains = safeJsonParse(domainList, []);
  if (domains.length === 0) return <p>all</p>;
  return (
    <div className="flex flex-col gap-y-2">
      {domains.map((domain, index) => {
        return (
          <p key={index} className="font-mono !font-normal">
            {domain}
          </p>
        );
      })}
    </div>
```
?

frontend/src/pages/GeneralSettings/ChatEmbedWidgets/EmbedConfigs/EmbedRow/EditEmbedModal/index.jsx
shatfield4: > Shouldnt this just be `safeJsonParse(embed.allowlist_domains, [])`

I originally tried doing this but the way safeJsonParse is currently implemented, `null` does not cause it to use the fallback.


```
export function safeJsonParse(jsonString, fallback = null) {
  try {
    return JSON.parse(jsonString);
  } catch {}
  return fallback;
}
```

If you try to `JSON.parse(null)` it actually does not throw an error which is why it won't use the fallback. It returns `null`.

Should I refactor the safeJsonParse util to add a guard clause like this?

```
export function safeJsonParse(jsonString, fallback = null) {
 if (jsonString === null || jsonString === undefined) return fallback;
  try {
    return JSON.parse(jsonString);
  } catch {}
  return fallback;
}
```


I originally hesitated since we use this in many places and was not sure what the impact on other areas of the app would be.

frontend/src/components/SettingsSidebar/MenuOption/index.jsx
shatfield4: Fixed to default to collapsed on failed json parse.

PR #4775, opened 2025-12-13
Refactor localWhisper to use custom FFMPEGWrapper class

### LINKED ISSUE #4774: [FEAT]: Refactor collector to remove fluent-ffmpeg package

In the collector we use `fluent-ffmpeg` - this package is deprecated and we should build our own ffmpeg wrapper for this.

- Remove this dependency from the repo
- Build a custom FFMPEG class that is a sub-util specifically for LocalWhisper
- This module should look for ffmpeg cli command 
- Should work in container - which has ffmpeg installer(Linux), but also work when running unbundled on Mac, Linux, and Windows.
- FFMPEG_PATH is the default path used by the existing library
- We should then use execSync or spawnSync to run the ffmpeg binary wherever it might be
- convert audio to .wav that is single channel @ 16k hz
- Output the file to a known destination in the collector tmp so that wavefile can load it in as a buffer.
- Add tests that are disabled in github actions (process.env.GITHUB_ACTIONS)
- wavefile should not perform any resampling. eg:
```
wavFile.toBitDepth("32f");
wavFile.toSampleRate(16000);
```

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4774 

### What is in this change?

- Adds new FFMPEGWrapper class in order to remove deprecated `fluent-ffmpeg` package
- FFMPEGWrapper class allows for simpler processing of audio files since ffmpeg can convert from any audio file type to WAV @ 16k hz, mono, 32bit float in a single CLI command
- Remove check for if `.wav` file in localWhisper class
- Add tests for FFMPEGWrapper class (disabled in github actions)
- Remove waveFile package from doing any resampling since this is done in the single conversion CLI command from ffmpeg

### Additional Information

- NOTE: could not remove resampling done by the `wavefile` library since this breaks the docker build. The same ffmpeg CLI command does not produce consistent outputs so we keep the `wavefile` resampling to ensure consistency across different environments.

### Developer Validations

### DIFF
--- a/collector/__tests__/utils/WhisperProviders/ffmpeg/index.test.js
+++ b/collector/__tests__/utils/WhisperProviders/ffmpeg/index.test.js
@@ -10,6 +10,7 @@ const { FFMPEGWrapper } = require("../../../../utils/WhisperProviders/ffmpeg");
 const describeRunner = process.env.GITHUB_ACTIONS ? describe.skip : describe;
 
 describeRunner("FFMPEGWrapper", () => {
+  /** @type { import("../../../../utils/WhisperProviders/ffmpeg/index").FFMPEGWrapper } */
   let ffmpeg;
   const testDir = path.resolve(__dirname, "../../../../storage/tmp");
   const inputPath = path.resolve(testDir, "test-input.wav");
@@ -25,15 +26,15 @@ describeRunner("FFMPEGWrapper", () => {
   });
 
   it("should find ffmpeg executable", async () => {
-    const ffmpegPath = await ffmpeg.getFFMPEGPath();
-    expect(ffmpegPath).toBeDefined();
-    expect(typeof ffmpegPath).toBe("string");
-    expect(ffmpegPath.length).toBeGreaterThan(0);
+    const knownPath = ffmpeg.ffmpegPath;
+    expect(knownPath).toBeDefined();
+    expect(typeof knownPath).toBe("string");
+    expect(knownPath.length).toBeGreaterThan(0);
   });
 
   it("should validate ffmpeg executable", async () => {
-    const ffmpegPath = await ffmpeg.getFFMPEGPath();
-    expect(ffmpeg.isValidFFMPEG(ffmpegPath)).toBe(true);
+    const knownPath = ffmpeg.ffmpegPath;
+    expect(ffmpeg.isValidFFMPEG(knownPath)).toBe(true);
   });
 
   it("should return false for invalid ffmpeg path", () => {
@@ -55,7 +56,7 @@ describeRunner("FFMPEGWrapper", () => {
     const buffer = await response.arrayBuffer();
     fs.writeFileSync(inputPath, Buffer.from(buffer));
 
-    const result = await ffmpeg.convertAudioToWav(inputPath, outputPath);
+    const result = ffmpeg.convertAudioToWav(inputPath, outputPath);
 
     expect(result).toBe(true);
     expect(fs.existsSync(outputPath)).toBe(true);
@@ -64,12 +65,12 @@ describeRunner("FFMPEGWrapper", () => {
     expect(stats.size).toBeGreaterThan(0);
   }, 30000);
 
-  it("should throw error when conversion fails", async () => {
+  it("should throw error when conversion fails", () => {
     const nonExistentFile = path.resolve(testDir, "non-existent-file.wav");
     const outputPath = path.resolve(testDir, "test-output-fail.wav");
 
-    await expect(
+    expect(() => {
       ffmpeg.convertAudioToWav(nonExistentFile, outputPath)
-    ).rejects.toThrow("FFMPEG conversion failed");
+    }).toThrow(`Input file ${nonExistentFile} does not exist.`);
   });
 });
--- a/collector/utils/WhisperProviders/ffmpeg/index.js
+++ b/collector/utils/WhisperProviders/ffmpeg/index.js
@@ -11,8 +11,12 @@ const { execSync, spawnSync } = require("child_process");
  * @class FFMPEGWrapper
  */
 class FFMPEGWrapper {
+  static _instance;
+
   constructor() {
-    this.ffmpegPath = null;
+    if (FFMPEGWrapper._instance) return FFMPEGWrapper._instance;
+    FFMPEGWrapper._instance = this;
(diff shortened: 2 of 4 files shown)


collector/__tests__/utils/WhisperProviders/ffmpeg/index.test.js
```diff
@@ -0,0 +1,60 @@
...
+    const result = await ffmpeg.convertAudioToWav(inputPath, outputPath);
+
+    expect(result).toBe(true);
+    expect(fs.existsSync(outputPath)).toBe(true);
+
+    const stats = fs.statSync(outputPath);
+    expect(stats.size).toBeGreaterThan(0);
+
+    fs.rmSync(inputPath);
+    fs.rmSync(outputPath);
```
timothycarambat: Should always clean up this on pass/fail/crash
https://jestjs.io/docs/setup-teardown

collector/utils/WhisperProviders/ffmpeg/index.js
```diff
@@ -0,0 +1,114 @@
...
+   *
+   * @async
+   * @returns {Promise<string>} Path to ffmpeg binary
+   * @throws {Error}
+   */
+  async getFFMPEGPath() {
+    if (this.ffmpegPath) return this.ffmpegPath;
+
+    if (
+      process.env.FFMPEG_PATH &&
```
timothycarambat: Is this ENV ever set? We do not set it in the container and I dont know if it resolves in the container automatically. all we do is `apt-get ffmpeg` which will put it in `/usr/bin` IIRC.

If this ENV is not explictly set then lets not bother checking for it at all.

timothycarambat: It is also called before fix-path anyway, so in Linux/Mac it wouldnt resolve anyway in a non-container production runtime since the shell ENV is not shared by default and this wont work downstream in desktop anyway

collector/utils/WhisperProviders/ffmpeg/index.js
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    if (process.platform !== "win32" && !process.env.JEST_WORKER_ID) {
+      try {
+        const fixPath = await import("fix-path").then((m) => m.default);
+        fixPath();
+      } catch (error) {
+        this.log("Could not load fix-path, using system PATH");
+      }
+    }
```
timothycarambat: Can `fix-path` not run in Jest or does it cause issues?

If so, instead of a code-level test patch, try to stub the import as a noop in the test that will call it. That way it does not resolve to anything that has an impact

collector/utils/WhisperProviders/ffmpeg/index.js
```diff
@@ -0,0 +1,114 @@
...
+   */
+  async convertAudioToWav(inputPath, outputPath) {
+    const ffmpegPath = await this.getFFMPEGPath();
+    const outputDir = path.dirname(outputPath);
+    if (!fs.existsSync(outputDir)) fs.mkdirSync(outputDir, { recursive: true });
+
+    this.log(`Converting ${path.basename(inputPath)} to WAV format...`);
+    // Convert to 16k hz mono 32f
+    const command = `"${ffmpegPath}" -i "${inputPath}" -ar 16000 -ac 1 -acodec pcm_f32le -y "${outputPath}"`;
+    execSync(command, { stdio: "pipe", maxBuffer: 1024 * 1024 * 10 });
```
timothycarambat: Pipe the conversion to the class logger so it is easy to see what is going on.
What is the reasoning around `maxBuffer`? Is there an issue there/why 10mb?

collector/utils/WhisperProviders/localWhisper.js
```diff
@@ -64,61 +64,31 @@ class LocalWhisper {
...
+      fs.rmSync(outputFile);
 
       const wavFile = new wavefile.WaveFile(buffer);
-      try {
-        this.#validateAudioFile(wavFile);
-      } catch (error) {
-        this.#log(`Audio validation failed: ${error.message}`);
-        throw new Error(`Invalid audio file: ${error.message}`);
-      }
+      this.#validateAudioFile(wavFile);
```
timothycarambat: This used to be try/catch'd, but no longer - Does a higher level catch handle this error now? It was mostly just to report a conversion failure or error

collector/utils/WhisperProviders/localWhisper.js
```diff
@@ -64,61 +64,31 @@ class LocalWhisper {
...
-        const stream = fs.createReadStream(sourcePath);
-        for await (let chunk of stream) chunks.push(chunk);
-        buffer = Buffer.concat(chunks);
-      }
+
+      const chunks = [];
+      const stream = fs.createReadStream(outputFile);
+      for await (let chunk of stream) chunks.push(chunk);
+      buffer = Buffer.concat(chunks);
+      fs.rmSync(outputFile);
```
timothycarambat: Taking a second look at this code (i know it was in code previously) it might just make more sense to read the whole file into a buffer without chunking and looping it. Less overhead.

Just readFile it directly with no encoding and we should have a buffer.

collector/utils/WhisperProviders/ffmpeg/index.js
shatfield4: Converted to use spawnSync instead of execSync since ffmpeg writes all progress to stderr (both success and fail) and removed the 10mb maxBuffer.

collector/utils/WhisperProviders/localWhisper.js
shatfield4: Reverted. Previously thought it was redundant and the parent would catch the errors thrown here. I now see the error logs are much more helpful when keeping it wrapped in the try catch for the specific error messages.

collector/utils/WhisperProviders/ffmpeg/index.js
shatfield4: Causes issues in Jest throwing seg faults if trying to use `import`. Switched to `require`, removed check for JEST_WORKER_ID, and now mocks the `fix-path` package as a noop to prevent this from happening.

PR #4780, opened 2025-12-14
feat: add AI Badgr (OpenAI-compatible) LLM provider

ghost:
### Pull Request Type

- [x ] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

connect #xxx

### What is in this change?

Adds AI Badgr as an optional OpenAI-compatible LLM provider.

Includes:

New provider implementation (server/utils/AiProviders/aibadgr) using OpenAI SDK with baseURL=https://aibadgr.com/api/v1

Tier-first default model preference (premium) with support for basic / normal / premium (and any OpenAI-compatible model name)

New onboarding/settings UI options for AI Badgr (API key + model preference)

Environment variable examples added to server/.env.example

Provider wiring in server/utils/helpers/index.js for both instance + class resolution

### Additional Information

AI Badgr is labeled as Budget/Utility to set expectations, and this change is opt-in only (no default behavior changed for existing providers).

### Developer Validations

- [ x] I ran `yarn lint` from the root of the repo & committed changes
- [x ] I have tested my code functionality
- [x ] Docker build succeeds locally

### DIFF
--- a/frontend/src/components/LLMSelection/AiBadgrOptions/index.jsx
+++ b/frontend/src/components/LLMSelection/AiBadgrOptions/index.jsx
@@ -1,0 +1,44 @@
+export default function AiBadgrOptions({ settings }) {
+  return (
+    <div className="flex gap-[36px] mt-1.5">
+      <div className="flex flex-col w-60">
+        <label className="text-white text-sm font-semibold block mb-3">
+          AI Badgr API Key
+        </label>
+        <input
+          type="password"
+          name="AiBadgrApiKey"
+          className="border-none bg-theme-settings-input-bg text-white placeholder:text-theme-settings-input-placeholder text-sm rounded-lg focus:outline-primary-button active:outline-primary-button outline-none block w-full p-2.5"
+          placeholder="AI Badgr API Key"
+          defaultValue={settings?.AiBadgrApiKey ? "*".repeat(20) : ""}
+          required={true}
+          autoComplete="off"
+          spellCheck={false}
+        />
+      </div>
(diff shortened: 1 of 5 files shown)


ghost:
@timothycarambat  can you give me feedback please

timothycarambat:
The main issue is that this PR is lacking an issue. From a code perspective, there are also several gaps:
- no agent functionality
- Missing entries for Data Handling/Onboarding selection
- Missing `customModels` entry. Since the API supports `/models` that should be supported
- Missing several entries for everywhere we do `switch` for other provider tags.
- The [custom logo is not set](https://github.com/Mintplex-Labs/anything-llm/pull/4780/files#diff-9e338906036e791563ecd67cabc932ba2e61f5f9c99640d74f0ac45c3776c374R306)

and that is just from a glance at the current state of the PR.

That being said, when we merge a provider, we basically have to assume we will be the one that has to maintain it forever - since most people will not come back to maintain their original work. So to that end without an easy way for us to test the service at least to confirm an integration, it does not incentivize us to make a bespoke provider for it.

This is especially the case in the crowded "openrouter-like" LLM inference market, where it just makes sense for the thousands of services like this to just use the Generic OpenAI Connector, which would work in this instance. When that is the case, this then becomes more of a marketing exercise than an integration request.

So all in all, we have the discretion to determine if a provider gets its own class or can just roll up under Generic OpenAI

PR #4785, opened 2025-12-15
Add Eslint Config in `/frontend`

angelplusultra:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### What is in this change?

This PR introduces an ESLint configuration scoped to the `/frontend` repository. It integrates ESLint with Prettier via the `eslint-plugin-prettier` package, allowing `eslint --fix` to both lint the code for issues and apply Prettier-based formatting automatically.

All existing rules that currently produce linting errors in `/frontend` have been temporarily disabled. These will be addressed and re-enabled in a follow-up PR.

### Developer Validations

### DIFF
--- a/.github/workflows/dev-build.yaml
+++ b/.github/workflows/dev-build.yaml
@@ -6,7 +6,7 @@ concurrency:
 
 on:
   push:
-    branches: ['enhancement-error-boundary'] # put your current branch to create a build. Core team only.
+    branches: ['refactor-eslint-frontend'] # put your current branch to create a build. Core team only.
     paths-ignore:
       - '**.md'
       - 'cloud-deployments/*'


frontend/package.json
```diff
@@ -8,6 +8,7 @@
     "dev": "cross-env NODE_ENV=development vite --debug --host=0.0.0.0",
     "build": "vite build && node scripts/postbuild.js",
     "lint": "yarn prettier --ignore-path ../.prettierignore --write ./src",
+    "lintv2": "eslint --fix",
```
timothycarambat: Lets not autofix so that we dont accidentally make changes we did not intend to make.

When i run this command in `frontend` I also get changes in files outside of `src` - this should be scoped to `src` only

angelplusultra: 1. I’ll go ahead and scope ESLint to `/src`.
2. Regarding the `--fix` flag:



The flag is intentionally designed to apply only deterministic, safe fixes that should not change runtime behavior. Any issues that could introduce legitimate bugs are not auto-fixed and must be addressed manually.
When combined with the ESLint Prettier plugin, --fix will also apply Prettier formatting to the targeted code.
If there’s still concern about relying on --fix, we can separate linting and formatting into two explicit steps. 

For example:
```json
{
  "scripts": {
    "lint:check": "eslint",
    "format": "prettier --write",
    "lint": "yarn lint:check && yarn format"
  }
}
```
In this setup, `yarn lint` would first run ESLint in **check-only mode** and report any issues without modifying files. If no errors are found, it would then proceed to apply Prettier formatting.


If this is what you prefer I can go ahead and implement.

timothycarambat: Lets do that then, I know what fix is supposed to do, but I also dont want people making PRs commiting changes they did not intend or are conciously aware of.

angelplusultra: Sounds good 👍

PR #4787, opened 2025-12-15
VectorDB class migration

### LINKED ISSUE #4687: [FEAT]: Refactor vector db providers

Refactor all vector dbs into Class objects that all inherit a base VectorDatabase base class.

The base class contains a method for every common method on all vector database providers. In some instances, it may make sense for the base class method to be defined if all or most providers run the same code for a method.

The vector database-specific providers will then have provider-specific overrides for each.

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4687 

### What is in this change?

- Refactors all vector databases to use classes instead of the object structure that was previously used
- Creates `VectorDatabase` base class that is extended onto all vector database providers to ensure all required functions are declared
- Refactors Zilliz Cloud provider to extend the Milvus class helping reduce code complexity since they have identical functions for interacting with the vector database

### Developer Validations

### DIFF
--- a/.github/workflows/dev-build.yaml
+++ b/.github/workflows/dev-build.yaml
@@ -6,7 +6,7 @@ concurrency:
 
 on:
   push:
-    branches: ['4841-aws-bedrock-api-key'] # put your current branch to create a build. Core team only.
+    branches: ['vectordb-class-migration'] # put your current branch to create a build. Core team only.
     paths-ignore:
       - '**.md'
       - 'cloud-deployments/*'


JOduMonT:
@shatfield4 **qdrant** works fine with **http**
image

PR #4793, opened 2025-12-17
Refactor workspace/pfp images

### LINKED ISSUE #4792: [FEAT]: Refactor workspace/pfp image

- Streamable chats requests the workspace and user pfp images on page load, every chunk received, and on stream completion
- We should refactor this to prevent the multiple rerenders that are happening and/or figure out a way to cache these images to limit the amount of network requests for the images
- Inlining SVGs using the vite-plugin-svgr package may help reduce these requests
- Fix rerendering since this prevents citations from being opened during streaming (rerenders cause it to be at its default closed state)

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [ ] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4792 

### What is in this change?

- Uses `vite-plugin-svgr` vite plugin to inline SVGs for workspace and user image
- Prevent entire message history rerendering when typing into the prompt input box by removing `regenerateAssistantMessage`, `saveEditedMessage`, and `forkThread` from `compiledHistory` function and adding `getMessageAlignment`
- Previously any time a user entered any key into the prompt input, it would cause the entire message history to rerender and rebuild the messages. This is because the parent component would rerender on input change, which would recreate the functions for `saveEditedMessage` and `forkThread` (defined in ChatHistory), causing useMemo to trigger the rebuild of messages

### Additional Information

- Discovered that any time you render an SVG in an `` it will cause a network request to be made regardless of if the component is properly memoized or not
- Also discovered that base64 encoding the image and inlining it in the react component prevents the network requests from happening since they are included in the js bundle (very messy)
- `vite-plugin-svgr` allows us to import SVGs as react components using `?react` in the import which inlines them in the js bundle without the base64 encoding and prevents the network requests
- Added logging to all chat/streaming related components when testing to check for unnecessary rerenders (now removed)

### Developer Validations


timothycarambat:
@shatfield4 Was this tested in a local production docker build and also with custom uploaded icons/user PFPs?

shatfield4:
> @shatfield4 Was this tested in a local production docker build and also with custom uploaded icons/user PFPs?

Yes, this was tested in docker and local. I just rebuilt docker image again with this to test and works to get rid of the network requests each time a message is sent for default user and workspace PFPs.

When using custom user/workspace PFPs this PR does not reduce the requests coming from the custom set images just due to the nature of `` (each time an `` is rendered, it will make a network request regardless of if it is memoized or not).

timothycarambat:
Closing this in favor of chat history UI redesign TBD

PR #4847, opened 2026-01-09
Make XLSX spreadsheets visible in chat by combining sheets

### LINKED ISSUE #4840: [BUG]: Excel Spreadsheet not visible

All versions

### What happened?

As a user, if I have an XLSX spreadsheet, we parse the spreadsheet into a sub-file per sheet. This results in dragged and dropped XLSX files not being "seen" by the LLM.

### Root cause
- When we parse XLSX files, we parse them into a folder where each sheet is a file. This then makes the `direct-upload` not work when appending the data to the prompt.

We should be able to recurse the `direct-uploads` for a workspace/thread to append to chat. The token count after dragging and dropping works so this should append properly.

Current fix: Convert XLSX to CSV and drop in chat. 

---

We should take note that any other filetypes we parse into subfolders that can be uploaded will _also_ have this issue.

It is unknown what the side effects of having direct-uploads having a folder when it comes to running the `cleanup-orphaned-docs` background job. We should also ensure this job does not wipe out a folder when it runs if attached to a workspace or chat.

Drag and drop an **xslx** file like [Untitled spreadsheet.xlsx](https://github.com/user-attachments/files/24511246/Untitled.spreadsheet.xlsx) into any chat.

Send any query like `What customers are on this?`
Get refusal from LLM.

shatfield4:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4840

### What is in this change?

- Fixes a bug where uploading an XLSX file directly to chat would not append the file as context to the chat
- Modified the `asXlsx` file converter util to append all sheets into a single file so that it can be injected directly into the context window when in `parseOnly` mode

### Additional Information

- Attempted to modify the `getContextFiles` util to accept folders but modifying the asXlsx converter was much a much simpler solution
- `asXlsx` converter is the only converter that has this problem
- `cleanup-orphan-documents` bg process works with the new fix 

### Developer Validations


collector/processSingleFile/convert/asXlsx.js
```diff
@@ -34,21 +34,16 @@ async function asXlsx({
...
-    for (const sheet of workSheetsFromFile) {
-      try {
+    if (options.parseOnly) {
+      const allSheetContents = [];
+      let totalWordCount = 0;
+      const sheetNames = [];
+
+      for (const sheet of workSheetsFromFile) {
         const { name, data } = sheet;
         const content = convertToCSV(data);
```
angelplusultra: Shouldn't the csv conversion be wrapped in a try/catch like how it is in the `else` block?

Line 112
```ts
for (const sheet of workSheetsFromFile) {
  try { 
    const { name, data } = sheet;
    const content = convertToCSV(data);
    
    if (!content?.length) {
      console.log(`Sheet "${name}" is empty. Skipping.`);
      continue;
    }
    // ... rest of processing
  } catch (err) {
    console.error(`Error processing sheet "${name}":`, err);
    continue; 
  }
}
```

angelplusultra (commented):
Code looks solid to me! Just a single comment.

collector/processSingleFile/convert/asXlsx.js
```diff
@@ -58,34 +53,103 @@ async function asXlsx({
...
+        }
+      );
+      const outFolderPath = path.resolve(documentsFolder, folderName);
+      if (!fs.existsSync(outFolderPath))
+        fs.mkdirSync(outFolderPath, { recursive: true });
+
+      for (const sheet of workSheetsFromFile) {
+        try {
+          const { name, data } = sheet;
+          const content = convertToCSV(data);
```
angelplusultra: Maybe abstract to a function to keep code DRY?

Example:
```ts
/**
 * Processes a single sheet and returns its content and metadata
 * @param {Object} sheet - The sheet object from xlsx.parse()
 * @returns {Object|null} - Object with { name, content, wordCount } or null if sheet is empty
 */
function processSheet(sheet) {
  try {
    const { name, data } = sheet;
    const content = convertToCSV(data);

    if (!content?.length) {
      console.log(`Sheet "${name}" is empty. Skipping.`);
      return null;
    }

    console.log(`-- Processing sheet: ${name} --`);
    return {
      name,
      content,
      wordCount: content.split(/\s+/).length,
    };
  } catch (err) {
    console.error(`Error processing sheet "${sheet.name}":`, err);
    return null;
  }
}
```

collector/processSingleFile/convert/asXlsx.js
shatfield4: Fixed by abstracting logic to a function and using the try/catch inside the new function.

shatfield4:
@angelplusultra Assigned back to you just to look over and approve before I assign to Tim.

collector/processSingleFile/convert/asXlsx.js
```diff
@@ -114,4 +161,31 @@ async function asXlsx({
   return { success: true, reason: null, documents };
 }
 
+/**
+ * Processes a single sheet and returns its content and metadata
+ * @param {Object} sheet - The sheet object
+ * @returns {Object|null} - Object with { name, content, wordCount } or null if sheet is empty
```
timothycarambat: nit: JSDoc here should have the object defined since it is known and only 3 keys

PR #4848, opened 2026-01-09
feat: Add ability to edit existing SQL agent connections

### LINKED ISSUE #4843: [FEAT]: Edit existing SQL connections

_Originally from #4834_

As a user, when I establish a valid SQL connection, I should be able to click on an existing connection and edit its specific connection with the same validations and config.

Clicking on a connection should show it's modal with prefilled fields that can then be edited. This connection will then be validated and updated in-place on save.

angelplusultra:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4843

### What is in this change?

https://github.com/user-attachments/assets/29d876c8-93ec-43eb-8faa-b30fc76eff83

This PR adds the ability to edit existing SQL connections in the agent configuration UI. Previously, users could only add new SQL connections or delete existing ones, but couldn't modify connection details.

**Frontend Changes:**
- Renamed and enhanced `NewConnectionModal.jsx` to `SQLConnectionModal.jsx` to support both create and edit modes
- Added edit button (pencil icon) to each SQL connection card in `DBConnection.jsx`
- Added new `getSQLConnectionDetails()` method in the System model to fetch connection details from the backend
- Implemented duplicate connection name detection on the frontend using slugification that matches backend behavior
- Added connection state refresh mechanism that reloads data from backend after save completes to ensure clean state
- Fixed input defocus issue in edit modal by using `defaultValue` instead of `value` for form inputs

**Backend Changes:**
- Added new `GET /system/sql-connection/:database_id` endpoint to fetch individual connection details
- Implemented connection string parsing to extract individual fields (username, password, host, port, database, etc.)
- Enhanced `mergeConnections()` function to handle `action: "update"` operations
- Improved duplicate connection prevention logic to handle edge cases and concurrent updates
- Fixed connection validation to skip already-processed updates and prevent auto-renaming conflicts

**Key Features:**
- Edit existing SQL connections without recreating them
- Prevent duplicate connection names with proper slugification
- Maintain connection details between frontend and backend consistently
- Proper error handling and user feedback via toast notifications

### Additional Information

The implementation uses the existing connection validation flow, so all connections are validated before being saved. The edit functionality reuses most of the existing modal UI components, just switching between "create" and "edit" modes based on whether an existing connection is provided.

### Developer Validations


frontend/src/models/system.js
```diff
@@ -841,6 +841,21 @@ const System = {
       });
   },
 
+  getSQLConnectionDetails: async function (database_id) {
```
shatfield4: This model method should be removed since it adds unneeded extra complexity.

You can modify the admin endpoint in `server/endpoints/admin.js` so that when we call `Admin.systemPreferencesByFields(["agent_sql_connections"])` it can return the SQL connection string here to prevent another call to our backend.

frontend/src/pages/Admin/Agents/SQLConnectorSelection/index.jsx
```diff
@@ -12,15 +12,32 @@ export default function AgentSQLConnectorSelection({
...
   setHasChanges,
+  hasChanges = false,
 }) {
   const { isOpen, openModal, closeModal } = useModal();
   const [connections, setConnections] = useState([]);
+  const prevHasChanges = useRef(hasChanges);
+
+  // Load connections on mount
   useEffect(() => {
     Admin.systemPreferencesByFields(["agent_sql_connections"])
```
shatfield4, line 27: Modify backend endpoint here to return the connection string so that we do this in a single call and pass as prop to `DBConnection` and parse the connection string there.

frontend/src/pages/Admin/Agents/SQLConnectorSelection/index.jsx
```diff
@@ -32,6 +49,23 @@ export default function AgentSQLConnectorSelection({
...
+    setHasChanges(true);
+    setConnections((prev) => {
+      // Remove the old connection with the original database_id
+      const filtered = prev.filter(
+        (conn) => conn.database_id !== updatedConnection.originalDatabaseId
+      );
+      // Add the updated connection
+      return [...filtered, updatedConnection];
+    });
+  }
```
shatfield4, line 83: This logic should be simplified.

```
function handleUpdateConnection(updatedConnection) {
  setHasChanges(true);
  setConnections((prev) =>
    prev.map((conn) =>
      conn.database_id === updatedConnection.originalDatabaseId
        ? updatedConnection
        : conn
    )
  );
}
```

This is clearer and doesn't require the extra filtering step.

frontend/src/pages/Admin/Agents/SQLConnectorSelection/SQLConnectionModal.jsx
```diff
@@ -0,0 +1,485 @@
...
+          clear: true,
+        });
+        closeModal();
+      } finally {
+        setIsLoadingDetails(false);
+      }
+    }
+
+    fetchConnectionDetails();
+  }, [isEditMode, isOpen, existingConnection?.database_id]);
```
shatfield4: This entire useEffect should be removed. The fetching should not happen inside of the modal but should be passed as props from the parent components to avoid unnecessary fetching.

server/models/systemSettings.js
```diff
@@ -722,25 +721,72 @@ function mergeConnections(existingConnections = [], updates = []) {
...
+      );
+
+      // Add the updated connection
+      updatedConnections.push({
+        engine: update.engine,
+        database_id: newId,
+        connectionString: update.connectionString,
+      });
+    });
+
```
shatfield4: This logic here is pretty hard to follow. Is there any way we can simplify this? It seems overly complex for what the code is trying to accomplish. The code should be self explanatory without being too verbose with comments here.

angelplusultra: I added better documentation of how the `mergeConnections` fn works in https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/c72ee3288fb32fefa4668da571720e6748b1da3d

frontend/src/pages/Admin/Agents/SQLConnectorSelection/SQLConnectionModal.jsx
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/c72ee3288fb32fefa4668da571720e6748b1da3d

frontend/src/models/system.js
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/c72ee3288fb32fefa4668da571720e6748b1da3d

frontend/src/pages/Admin/Agents/SQLConnectorSelection/index.jsx
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/c72ee3288fb32fefa4668da571720e6748b1da3d

frontend/src/pages/Admin/Agents/SQLConnectorSelection/index.jsx
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/07b736b468e0f857571d783791710b90b45f680a

server/models/systemSettings.js
```diff
@@ -710,42 +726,88 @@ const SystemSettings = {
...
 
+/**
+ * Merges SQL connection updates from the frontend with existing backend connections.
+ * Processes three types of actions: "remove", "update", and "add".
+ *
+ * @param {Array<Object>} existingConnections - Current connections stored in the database
+ * @param {Array<Object>} updates - Connection updates from frontend, each with an action property
+ * @returns {Array<Object>} - The merged connections array
+ */
 function mergeConnections(existingConnections = [], updates = []) {
```
shatfield4, line 769: I refactored this `mergeConnections` util to now use a map to help simplify the implementation. From my testing it works exactly the same as the previous implementation but be sure to test this again on your end.

frontend/src/pages/Admin/Agents/SQLConnectorSelection/SQLConnectionModal.jsx
```diff
@@ -40,15 +68,51 @@ const DEFAULT_CONFIG = {
...
+ * @param {Object} props - Component props
+ * @param {boolean} props.isOpen - Whether the modal is currently open
+ * @param {Function} props.closeModal - Callback to close the modal
+ * @param {Function} props.onSubmit - Callback when connection is successfully validated and saved
+ * @param {Function} props.setHasChanges - Callback to mark that changes have been made
+ * @param {Object|null} [props.existingConnection=null] - Existing connection data for edit mode (contains database_id, engine, username, password, host, port, database, scheme, encrypt)
+ * @param {Array} [props.connections=[]] - List of all existing connections for duplicate detection
+ * @returns {React.ReactPortal|null} - Portal containing the modal UI, or null if not open
+ */
+export default function SQLConnectionModal({
```
shatfield4, line 86: I ran into this UI bug when editing connections:

- If you have 2 connections and press edit icon and close modal, then edit again the modal is no longer filled in with previous info
- I notice that the save changes bar pops up when editing multiple times when the modal is cleared so it may have to do with state management of hasChanges

frontend/src/pages/Admin/Agents/SQLConnectorSelection/index.jsx
shatfield4: Need to add a loading state here while connections are being fetched. Currently in the UI it shows there are no connections then they pop up when they are finally fetched.

This can take a few seconds if the user has multiple connections setup and confused me when I was setting up my db connections.

server/endpoints/admin.js
```diff
@@ -376,7 +376,7 @@ function adminEndpoints(app) {
               break;
             case "agent_sql_connections":
               requestedSettings[label] =
-                await SystemSettings.brief.agent_sql_connections();
+                await SystemSettings.agent_sql_connections();
```
shatfield4, line 379: 👍

frontend/src/pages/Admin/Agents/SQLConnectorSelection/SQLConnectionModal.jsx
angelplusultra: It had to do with the modal not recomputing the `existingConnection` values. When the modal closes it doesnt unmount t just returns null so it never recomputed. 

 Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/89f95df011b3f01cda73d3e01a3b03c5161abc0a

frontend/src/pages/Admin/Agents/SQLConnectorSelection/index.jsx
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/b42e2fa3fbb9cb453503ad7200e771d08e913a14

server/models/systemSettings.js
angelplusultra: Everything seems to be working fine on my end with this refactor.

shatfield4 (approved):
LGTM

server/endpoints/system.js
```diff
@@ -11,6 +11,7 @@ const {
   userFromSession,
   multiUserMode,
   queryParams,
+  safeJsonParse,
```
timothycarambat: Unused import?

server/endpoints/admin.js
timothycarambat: what even was `brief`? i know this was way prior to your PR but is it used anywhere or even doing anything? might be legacy code ref

timothycarambat, line 379: There was a reason we are using brief here for sending back the connection back to the frontend, however ill follow that up in another comment

server/models/systemSettings.js
```diff
@@ -677,18 +680,31 @@ const SystemSettings = {
     };
   },
 
-  // For special retrieval of a key setting that does not expose any credential information
-  brief: {
```
timothycarambat, line 685: Ah, I see, brief was to prevent sending the raw credentials back to the frontend since it is the direct SQL connection information - even if read-only.

I suppose all this did prior was send back all information in JSON form with the exception of `connectionString` - which at the time would just be
```json
[
    {
        "database_id": "myuniquedb",
        "engine": "postgresql"
    }
]
```

Thus, not expose any information to the UI via this call. I did go ahead and check and non-admins cannot even hit this endpoint to get the information. However, now we are passing back basically the entire credentials in plaintext back to the UI.

That being said, since this is only for admins, I believe doing this is okay as otherwise there is no way to verify or validate the values for a DB connection, which can be very tedious for those not used to using connection strings or DB's in general.

Take some time to review this carefully to ensure there is no accidental leakage of this information to the frontend (non-admins) and even in logging/tools or anything like that.

server/models/systemSettings.js
```diff
@@ -738,42 +754,88 @@ const SystemSettings = {
...
+      action,
+      database_id,
+      originalDatabaseId,
+      connectionString,
+      engine,
+    } = update;
+
+    // Skip invalid connections
+    if ((action === "update" || action === "add") && !connectionString)
+      continue;
```
timothycarambat: move this into the case statement for update/add?

server/models/systemSettings.js
```diff
@@ -738,42 +754,88 @@ const SystemSettings = {
...
+        }
+
+        connectionsMap.set(slugifiedId, {
+          engine,
+          database_id: slugifiedId,
+          connectionString,
+        });
+        break;
+      }
+    }
```
timothycarambat, line 841: if the action is not valid and default case is hit we should likely throw.

angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/06e7f1e937087cc9ee8005ba7611717797a48b13

server/models/systemSettings.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/7d3b0114db171259a382e4fdf9d016e682d59a52

server/endpoints/system.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/4848/commits/ea8e64eda599ce2c14b05b923797bcf3014196a9

server/models/systemSettings.js
angelplusultra: Only Admins and Managers can hit the endpoint to call this handler. Verified no logging or weird stuff with credentials.

PR #4879, opened 2026-01-18
feat: add n1n LLM provider

n1n-api:
This PR adds n1n as a new LLM provider. n1n is an OpenAI-compatible provider that offers various powerful models. The integration includes backend support, environment variable management, and frontend configuration UI with custom model fetching support.


timothycarambat:
All PRs require an issue.
PR is missing DataHandling entry
PR is missing any `@agent` provider support + entries to other provider case/switch locations to be fully supported. As it stands, any `@agent` request would fail.

You should not need to add any models to [legacy](https://github.com/Mintplex-Labs/anything-llm/pull/4879/files#diff-05be3fc882616141c00bbb5aae85a1695abc7d92da753cdb2eadb1e81e39f925R145) if this service has a working `/models` endpoint

Ultimately, we wind up maintaining every provider integration, and so if you can use the [OpenAI SDK](https://github.com/Mintplex-Labs/anything-llm/pull/4879/files#diff-766d07e63d9e20fe53e13985728f98593fe5a50ba2cebec876a170edd6f2cd9fR14) we recommend your sub-users to simply use our generic provider so there is no major code change and all compatible features are fully supported.

PR #4914, opened 2026-01-27
Fix: Clean username already exists error

### LINKED ISSUE #4912: [BUG]: Clean Username Already Exists Error

All versions

### What happened?

In multi-user mode as an admin:
If i try to create a user with the same name as another user I get a nasty prisma error message
If i try to update an existing user to a username that is taken i get a nasty prisma error.

This should instead return a clear error message like "username is taken".

This error likely also occurs on
Invitation acceptance
Any other places users can put in their own username (like account modal in top right of UI)

When possible, we should show a neat error message instead of this overly verbose error message, which can instead of backend logged.

_No response_

angelplusultra:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4912

### What is in this change?

Handles Prisma unique constraint violation errors (P2002) in the User model to return user-friendly error messages instead of raw Prisma error strings.

**Before:** Users see verbose Prisma errors like `Unique constraint failed on the fields: ('username')`

**After:** Users see clean messages like `A user with that username already exists`

Changes:
- Added `Prisma` import from `@prisma/client` in `server/models/user.js`
- Added P2002 error handling in `User.create()` method
- Added P2002 error handling in `User.update()` method

### Additional Information

This fix covers all user creation and update endpoints across the codebase (admin panel, API, invite acceptance, profile updates, etc.)

### Developer Validations

### DIFF
--- a/server/models/user.js
+++ b/server/models/user.js
@@ -90,6 +90,16 @@ const User = {
     const { password, ...rest } = user;
     return { ...rest };
   },
+  _identifyErrorAndFormatMessage: function (error) {
+    if (error instanceof Prisma.PrismaClientKnownRequestError) {
+      // P2002 is the unique constraint violation error code
+      if (error.code === "P2002") {
+        const target = error.meta?.target;
+        return `A user with that ${target?.join(", ")} already exists`;
+      }
+    }
+    return error.message;
+  },
 
   create: async function ({
     username,
@@ -122,17 +132,7 @@ const User = {
       return { user: this.filterFields(user), error: null };
     } catch (error) {
       console.error("FAILED TO CREATE USER.", error.message);
-      if (error instanceof Prisma.PrismaClientKnownRequestError) {
-        // P2002 is the unique constraint violation error code
-        if (error.code === "P2002") {
-          const target = error.meta?.target;
-          return {
-            user: null,
-            error: `A user with that ${target?.join(", ")} already exists`,
-          };
-        }
-      }
-      return { user: null, error: error.message };
+      return { user: null, error: this._identifyErrorAndFormatMessage(error) };
     }
   },
   // Log the changes to a user object, but omit sensitive fields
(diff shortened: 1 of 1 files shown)


server/models/user.js
```diff
@@ -199,6 +210,17 @@ const User = {
...
+          const target = error.meta?.target;
+          return {
+            user: null,
+            error: `A user with that ${target?.join(", ")} already exists`,
+          };
+        }
+      }
       return { success: false, error: error.message };
     }
   },
```
shatfield4, line 218: Let's extract the error code handling to a private function in the User object so that we don't have 2 return statements in the error handling blocks of `create` and `update` methods.

```
  _handlePrismaError: function (error) {                                                                                        
    if (error instanceof Prisma.PrismaClientKnownRequestError && error.code === "P2002") {                                                                                                                         
      const target = error.meta?.target;                                                                                        
      return `A user with that ${target?.join(", ")} already exists`;                                                           
    }                                                                                                                           
    return null;                                                                                                                
  }, 
```

server/models/user.js
```diff
@@ -121,6 +122,16 @@ const User = {
...
+        // P2002 is the unique constraint violation error code
+        if (error.code === "P2002") {
+          const target = error.meta?.target;
+          return {
+            user: null,
+            error: `A user with that ${target?.join(", ")} already exists`,
+          };
+        }
+      }
       return { user: null, error: error.message };
```
shatfield4: Can be replaced with util function like this

```
} catch (error) {                                                                                                             
    console.error("FAILED TO CREATE USER.", error.message);                                                                     
    return { user: null, error: this._handlePrismaError(error) || error.message };                                              
  } 
```

server/models/user.js
```diff
@@ -199,6 +210,17 @@ const User = {
...
+        // P2002 is the unique constraint violation error code
+        if (error.code === "P2002") {
+          const target = error.meta?.target;
+          return {
+            user: null,
+            error: `A user with that ${target?.join(", ")} already exists`,
+          };
+        }
+      }
       return { success: false, error: error.message };
```
shatfield4: Return values are different here. Early return is returning an object with `user` and `error` and second return is returning `success` and `error`. These should be the same.

But if abstracting to a function it can just be replaced with this:


```
  } catch (error) {                                                                                                             
    console.error(error.message);                                                                                               
    return { success: false, error: this._handlePrismaError(error) || error.message };                                          
  }
```

server/models/user.js
angelplusultra: How does https://github.com/Mintplex-Labs/anything-llm/pull/4914/commits/709c875854521fe014030b32ded97fbfb1ea2c65 look to you?

shatfield4: LGTM

shatfield4 (approved):
LGTM

PR #4931, opened 2026-01-29
Implement new home page redesign

### LINKED ISSUE #4911: [FEAT]: Home page redesign implementation

Implements the redesign of the home page per internal designs.

shatfield4:
### Pull Request Type

- [x] ✨ feat
- [ ] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4911

### What is in this change?

- Implements new designs for the home page and empty state workspace chats
- Adds quick action buttons that are only visible to admin/manager roles
- New UI for suggested chat messages
- New UI for PromptInput component inside workspace chat and home page

### Additional Information

- #4986 - translations PR
### Developer Validations

### DIFF
--- a/.github/workflows/cleanup-qa-tag.yaml
+++ b/.github/workflows/cleanup-qa-tag.yaml
@@ -20,7 +20,7 @@ jobs:
         continue-on-error: true  # Package may not exist if build never ran
         with:
           owner: ${{ github.repository_owner }}
-          package-name: ${{ github.repository }}
+          package-name: ${{ github.event.repository.name }}
           package-type: container
           min-versions-to-keep: 0
           ignore-versions: '^(?!pr-${{ github.event.pull_request.number }}$).*$'
--- a/frontend/src/LogoContext.jsx
+++ b/frontend/src/LogoContext.jsx
@@ -6,18 +6,21 @@ import DefaultLoginLogoDark from "./media/illustrations/login-logo-light.svg";
 import System from "./models/system";
 
 export const REFETCH_LOGO_EVENT = "refetch-logo";
+
+function isLightMode() {
+  return document.documentElement.getAttribute("data-theme") === "light";
+}
 export const LogoContext = createContext();
 
 export function LogoProvider({ children }) {
   const [logo, setLogo] = useState("");
   const [loginLogo, setLoginLogo] = useState("");
   const [isCustomLogo, setIsCustomLogo] = useState(false);
-  const DefaultLoginLogo =
-    localStorage.getItem("theme") !== "default"
-      ? DefaultLoginLogoDark
-      : DefaultLoginLogoLight;
 
   async function fetchInstanceLogo() {
+    const DefaultLoginLogo = isLightMode()
+      ? DefaultLoginLogoDark
+      : DefaultLoginLogoLight;
     try {
       const { isCustomLogo, logoURL } = await System.fetchLogo();
       if (logoURL) {
@@ -25,16 +28,12 @@ export function LogoProvider({ children }) {
         setLoginLogo(isCustomLogo ? logoURL : DefaultLoginLogo);
         setIsCustomLogo(isCustomLogo);
       } else {
-        localStorage.getItem("theme") !== "default"
-          ? setLogo(AnythingLLMDark)
-          : setLogo(AnythingLLM);
+        isLightMode() ? setLogo(AnythingLLMDark) : setLogo(AnythingLLM);
         setLoginLogo(DefaultLoginLogo);
         setIsCustomLogo(false);
       }
     } catch (err) {
-      localStorage.getItem("theme") !== "default"
-        ? setLogo(AnythingLLMDark)
-        : setLogo(AnythingLLM);
+      isLightMode() ? setLogo(AnythingLLMDark) : setLogo(AnythingLLM);
       setLoginLogo(DefaultLoginLogo);
       setIsCustomLogo(false);
       console.error("Failed to fetch logo:", err);
--- a/frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
@@ -388,20 +388,15 @@ function useIsDisabled() {
    */
   useEffect(() => {
     if (!window) return;
-    window.addEventListener(ATTACHMENTS_PROCESSING_EVENT, () =>
-      setIsDisabled(true)
-    );
-    window.addEventListener(ATTACHMENTS_PROCESSED_EVENT, () =>
-      setIsDisabled(false)
-    );
+    const onProcessing = () => setIsDisabled(true);
+    const onProcessed = () => setIsDisabled(false);
+
+    window.addEventListener(ATTACHMENTS_PROCESSING_EVENT, onProcessing);
+    window.addEventListener(ATTACHMENTS_PROCESSED_EVENT, onProcessed);
 
     return () => {
-      window?.removeEventListener(ATTACHMENTS_PROCESSING_EVENT, () =>
-        setIsDisabled(true)
-      );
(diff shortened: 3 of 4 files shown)


timothycarambat:
@shatfield4 I know this is not ready but some feedback:
- Microphone is busted, does not accumulate text, only replaces what is currently in the input. Also will remove any previous text when spoken into.

- Padding for buttons is wrong - the mt-6 is 24px, but the box heigh extends beyond the visible area so the padding is higher.

- The buttons/ctas have no hover state, sync with design on what that should be (maybe just an opacity??)

- Upload document (imo) should result in a **parsed file** not an embed. This will lead to a frustrating UX going back to the whole "Why cant the LLM summarize my document" UX complaint that we had before

- The input clears on send then changes UI. Is it possible we keep the text in the input or are these components the same component and it just moves? Low priority for this IMO, just a UI quirk.

We might want to make another PR after this to fix the reset message - it causes the page shift as if a chat was sent if you send it first.

frontend/src/components/lib/QuickActions/index.jsx
```diff
@@ -0,0 +1,32 @@
...
+export default function QuickActions({
+  onCreateAgent,
+  onEditWorkspace,
+  onUploadDocument,
+}) {
+  return (
+    <div className="flex flex-wrap justify-center gap-2 mt-6">
+      <QuickActionButton label="Create an Agent" onClick={onCreateAgent} />
+      <QuickActionButton label="Edit Workspace" onClick={onEditWorkspace} />
+      <QuickActionButton label="Upload a Document" onClick={onUploadDocument} />
```
angelplusultra: Should these labels have keys in i18n?

frontend/src/components/WorkspaceChat/ChatContainer/index.jsx
```diff
@@ -301,6 +329,55 @@ export default function ChatContainer({ workspace, knownHistory = [] }) {
...
+                attachments={files}
+                centered
+              />
+              {(!user || user.role !== "default") && (
+                <QuickActions
+                  onCreateAgent={() => navigate(paths.settings.agentSkills())}
+                  onEditWorkspace={() =>
+                    navigate(
+                      paths.workspace.settings.generalAppearance(workspace.slug)
+                    )
```
angelplusultra: `workspace` could be null, add some sort of guard against null workspaces. Maybe even disable the button until workspace is truthy for better UX.

frontend/src/pages/Main/Home/index.jsx
```diff
@@ -1,26 +1,295 @@
...
+
+async function getTargetWorkspace() {
+  const lastVisited = safeJsonParse(
+    localStorage.getItem(LAST_VISITED_WORKSPACE)
+  );
+  if (lastVisited?.slug) {
+    const workspace = await Workspace.bySlug(lastVisited.slug);
+    if (workspace) return workspace;
+  }
+
```
angelplusultra, line 35: Shouldnt you also get the suggested messages for the targeted workspace and display them?

frontend/src/components/WorkspaceChat/ChatContainer/index.jsx
```diff
@@ -331,3 +407,33 @@ export default function ChatContainer({ workspace, knownHistory = [] }) {
...
+        if (!text) return null;
+
+        return (
+          <div key={index}>
+            {index > 0 && (
+              <div className="border-t border-zinc-800 light:border-theme-chat-input-border" />
+            )}
+            <button
+              type="button"
+              onClick={() => sendCommand({ text, autoSubmit: true })}
```
angelplusultra: Should it really autoSubmit? Not just inject the suggested message into the input?

angelplusultra (changes requested):
Overall this looks great. I just left a few notes.

My biggest concern is the home page defaulting to the last visited workspace without clearly indicating which workspace that is. If you don’t remember where you were last, submitting a prompt from the home page becomes a bit of a dice roll. It may be worth considering a way to clearly display the targeted workspace.

frontend/src/components/WorkspaceChat/ChatContainer/index.jsx
shatfield4: Parent component already protects against workspace being null. It will not load the ChatContainer component at all unless there is a valid target workspace.

frontend/src/components/WorkspaceChat/ChatContainer/index.jsx
shatfield4: Yes, the autosubmit behavior is the same as it was before with the current suggested messages for workspace.

angelplusultra (changes requested):
Overall looks solid and works great. I have one refactor suggestion for code maintainability.


The greeting + PromptInput + QuickActions + SuggestedMessages layout is largely duplicated between `Home/index.jsx:256-293` and `ChatContainer/index.jsx:329-397`. Both render the same structure with slightly different wiring. Consider extracting this into a shared component (e.g., `EmptyStateView`) to reduce duplication and ensure visual consistency.

shatfield4:
> Overall looks solid and works great. I have one refactor suggestion for code maintainability.
> 
> The greeting + PromptInput + QuickActions + SuggestedMessages layout is largely duplicated between `Home/index.jsx:256-293` and `ChatContainer/index.jsx:329-397`. Both render the same structure with slightly different wiring. Consider extracting this into a shared component (e.g., `EmptyStateView`) to reduce duplication and ensure visual consistency.

Took a crack at extracting this to a shared component but due to the wiring and differences in props, this extraction would make the overall implementation more complex. In the future we may want to add new components to only the home page so IMO it is a good idea to keep these separate for now.

timothycarambat (changes requested):
Discovered issues:

If i upload a file I can click the "x" to remove them as parsed files but hoving over the paperclip i cannot click the X to remove it as a parsed file via the modal.

Image

The new landing page logic works for admins, but for managers and default level users this logic does not work.

Managers can create workspaces - so they should be able to create a new workspace and send a chat and auto-create a workspace. Instead it makes me choose a workspace or tells me i have none. If i open a workspace the empty state exists

Image

Image

For users, this is nonfunctional in a similar way. If i have a workspace then direct me to the new landing so i can send a chat instantly. If i have no workspaces then i should see the  "contact system admin" landing page.

Image


We need to confirm for the command prompt and agent prompt popovers are correctly laid out when the input modal is tall. I dont think this is the layout we want.

Image
Image

As an admin, if i have a workspace we can assume to send chats to - the sidebar does not show it - here i am actually about to send a prompt to MyWorkspace, which has 3 threads already. I am on the root page, but its not clear where I am about to send a chat.

> I can confirm that actually happens in single user mode too - if we are assuming a workspace we should highlight on sidebar along with thread if we are tracking it. Otherwise whatever the assumed workspace is should be expanded too.

Image

Change model is missing - when the prompt is showing and assuming a present workspace as to where we are chatting we should enable showing the model change "brain" icon. When the workspace does not yet exist - we should **not** show this icon.

----

https://github.com/Mintplex-Labs/anything-llm/pull/4931/changes#diff-635717db84d7d0d28b08a5a889f480ffdd7e8fde636d6b1235ce9c90ae45cb75R352

This record in the translations does not exist (i cant find it??) and it's not in the associated translation PR. Additionally, this same fallback string should just be a constant so its in one place. However, just making a real record in EN would make it the fallback for any `null` language entry once normalized.

-----

Create An agent quick-action is showing for managers - who cannot access this page.

shatfield4:
- [x] Fix translations for greeting
- [x] Fix removing files from `ParsedFileMenu` component in home page
- [x] Manager should be able to see new home page since they are allowed to create workspaces
- [x] If user has workspace available we should show new home page and send chat directly to it (block some features since they are not allowed)
- [x] Add brain icon change model popup back to home page
- [x] Review with design slash commands pop up/agent pop up
- [x] QA approval

angelplusultra (approved):
LGTM

PR #4981, opened 2026-02-10
fix: correct TPS calculation for Generic OpenAI provider with llama.cpp

### LINKED ISSUE #4972: [BUG]: Generic OpenAI on llama.cpp server incorrect metrics
### Stack

- AnythingLLM Desktop Windows v1.10.0.
- Docker hosted llama.cpp server in router mode.

### Issue

The displayed metrics inside AnythingLLM Desktop and the corresponding metrics for the same request inside the llama.cpp server logs.

Image

```
[38311] prompt eval time =     131.41 ms /    53 tokens (    2.48 ms per token,   403.33 tokens per second)
[38311]        eval time =    1809.32 ms /   222 tokens (    8.15 ms per token,   122.70 tokens per second)
[38311]       total time =    1940.73 ms /   275 tokens
```

The evaluation time is correct, 1.809s for 1809.32ms.
The tokens per second are incorrect, they should be 122.70 tok/s and not 4.98 tok/s.

### Details

Tested with GLM-4.7-Flash and Qwen3-VL-30B-A3B-Thinking so I suspect the issue happens with any LLM.

I have not tested on a llama.cpp server not in router mode but I don't think it makes any difference.

angelplusultra:
### Pull Request Type

- [ ] ✨ feat
- [x] 🐛 fix
- [x] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves #4972

### What is in this change?

Fixes inaccurate tokens-per-second (TPS) calculation when using the Generic OpenAI provider with a llama.cpp server.

The TPS was being calculated using wall-clock duration (which includes prompt evaluation time) instead of generation-only duration. This caused massively deflated TPS values in the UI — for example, llama.cpp would report ~15 TPS while AnythingLLM displayed ~1 TPS.

llama.cpp returns a `timings` object on the final streaming chunk with accurate generation-only metrics (`predicted_n`, `predicted_ms`). This change extracts those values and passes them to `endMeasurement`, similar to how the Ollama provider already uses `eval_duration` / `eval_count`.

The timings extraction is implemented as a private method (`#extractTimings`) that mutates the usage object in place, shared by both the streaming (`handleStream`) and non-streaming (`getChatCompletion`) code paths. This ensures accurate TPS for both modes when using a llama.cpp backend.

Non-llama.cpp servers that don't return `timings` are unaffected — the existing wall-clock fallback behavior is unchanged.

### Additional Information

The `timings` object from llama.cpp's final chunk looks like:
```json
{
  "predicted_n": 165,
  "predicted_ms": 5889.311,
  "predicted_per_second": 28.01,
  "prompt_n": 5,
  "prompt_ms": 99.914,
  "prompt_per_second": 50.04
}
```

### Developer Validations

### DIFF
--- a/server/utils/AiProviders/genericOpenAi/index.js
+++ b/server/utils/AiProviders/genericOpenAi/index.js
@@ -156,15 +156,14 @@ class GenericOpenAiLLM {
    * @param {Object} response - the API response or final streaming chunk
    * @param {Object} usage - the usage object to mutate
    */
-  #extractTimings(response, usage) {
-    if (
-      response?.timings &&
-      "predicted_n" in response.timings &&
-      "predicted_ms" in response.timings
-    ) {
-      usage.completion_tokens = response.timings.predicted_n;
-      usage.duration = response.timings.predicted_ms / 1000;
-    }
+  #extractLlamaCppTimings(response, usage) {
+    if (!response || !response.timings) return;
+
+    if (response.timings.hasOwnProperty("predicted_n"))
+      usage.completion_tokens = Number(response.timings.predicted_n);
+
+    if (response.timings.hasOwnProperty("predicted_ms"))
+      usage.duration = Number(response.timings.predicted_ms) / 1000;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


shatfield4 (approved):
LGTM

PR #5021, opened 2026-02-18
fix: GitLab connector infinite loop and rate limit crash for large repos

### LINKED ISSUE #5003: [BUG] GitLab Connector: Infinite pagination loop on large repos / JSON crash on Rate Limit

Local development

### What happened?

I have encountered two issues when using the GitLab Connector with large repositories (>10k files):

1. **Infinite Loop (Pagination Bug)**: GitLab stops sending the x-total-pages header for large repositories for performance reasons (https://gitlab.com/gitlab-org/gitlab-foss/-/merge_requests/23931, https://gitlab.com/gitlab-org/gitlab/-/issues/361362). The current logic relies on this header to detect the last page. When it is missing, the connector resets to Page 1 and loops forever.
2. **Crash on Rate Limit**: The connector attempts to download all files in a page simultaneously using Promise.all. This triggers GitLab's HTTP 429 (Rate Limit) protection. The API returns a plain text error ("Rate limit exceeded"), which causes JSON.parse() to throw a SyntaxError: Unexpected token 'R', crashing the collector.

This is also happening in Docker and using a private Repo. 

I have chosen https://gitlab.com/qemu-project/qemu just as an example, because I was searching for a publicly available repository with more than 10k files. The same would happen for https://gitlab.com/gitlab-org/gitlab-foss.

1. Start the Collector: Run the collector service (e.g., yarn dev:collector or via Docker).
2. Configure GitLab Connector:
3. Open a Workspace in AnythingLLM.  Go to the Connectors and select Gitlab. Repo URL: Enter a large public repository URL (e.g., https://gitlab.com/qemu-project/qemu)).  Access Token:  use a Personal Access Token .  (Note: Using no token hits rate limits faster, triggering the crash sooner).
4. Trigger the Bug by clicking "submit" to start the process
5. Observe the Logs:
-  Crash: The collector logs SyntaxError: Unexpected token 'R' (due to unhandled 429 Rate Limit response).
-  Infinite Loop: If it survives the crash, it reaches the last page (e.g., Page 118) and then restarts at Page 1 because x-total-pages is missing.

The endless loop. It starts with the following. You can already see, that the x-total-pages is 0 after the first message.
Image

Upon reaching the end, it continues downloading, starting right from the beginning.
Image

I made a quick attempt to solve the issue, which would require some further cleanup: anything-llm\collector\utils\extensions\RepoLoader\GitlabRepo\RepoLoader\index.js:
[index.js](https://github.com/user-attachments/files/25326714/index.js)

Two adjustments for the _fetchNextPage_ function:

1. Only relay on x-next-page
2. Check for rate-limit response. If reached, wait a minute, so the limit gets refreshed, then continue downloading (here some smaller changes are still required)

With those changes, I can download all files.

If desired, I could try to clean this up and provide a proper PR.

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5003

### Description

Fixes two bugs in the GitLab connector that occur with large repositories (>10k files):

1. **Infinite loop (pagination):** GitLab omits the `x-total-pages` header for large repos for performance reasons. The old code relied on this header to detect the last page. When missing, pagination reset to page 1 and looped forever. Fixed by using `x-next-page` as the sole pagination signal — when it's empty, we're on the last page.

2. **Crash on rate limit:** The old code checked for status `401` (Unauthorized) instead of `429` (Rate Limit), and called `response.json()` unconditionally. When GitLab returned a plain-text 429 response like `"Rate limit exceeded"`, `JSON.parse` threw a `SyntaxError`, crashing the collector. Fixed by handling 429 in both `fetchNextPage` and `fetchSingleFileContents` with retry logic that respects the `retry-after` header (defaulting to 60s, capped at 3 retries).

### Additional Information

- Also fixed the `401` log message which incorrectly said "Rate limit hit" — it now correctly says "Unauthorized".
- Added non-OK status guard before `response.json()` to prevent JSON parse errors on any unexpected error response.

### Developer Validations


shatfield4 (approved):
LGTM tested using qemu repo.

PR #5031, opened 2026-02-20
feat: Add Hostim.dev as a new self-hosting deployment option to the R…

pavel1337:
…EADME.

### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)
- [x] 📝 docs (Documentation updates)

### Description

This PR adds support for deployment on Hostim.dev.

### Developer Validations


timothycarambat:
There is an issue with however you run the containers. Restarts wipe whatever was saved as an ENV like the LLM provider.
Here is the official start command https://docs.anythingllm.com/installation-docker/quickstart

It is very important you have these volume binds:
```
```

if you do not, nothing will ever store. Additionally, whatever you have for the ENV on boot is somehow skipping onboarding totally so it just opens to the main window but every chat will fail since no LLM provider is set, which is the first screen of the onboarding to prevent exactly that.

Closing until patched

PR #5040, opened 2026-02-21
fix: use per-model output token limits for Anthropic agent provider

### LINKED ISSUE #5039: [BUG]: Anthropic agent responses truncate at 4096 tokens on modern models

All versions

### What happened?

The Anthropic agent provider hardcodes max_tokens: 4096 for all API calls. Modern Claude models 
  (Claude 3.5 and later) support output limits well above 4096 tokens, but the hardcoded value was
  set when older models had smaller limits. As a result, agent responses are silently cut off at
  4096 tokens regardless of which model is selected.

  Affected versions:
  All current versions of AnythingLLM using the Anthropic agent provider.

  Behavior:
  When an agent session produces a response longer than 4096 tokens — for example, when processing
  large inputs passed via flow variables, or when a task requires a detailed output — the response
  is truncated at 4096 tokens with no warning or error shown to the user.

  Expected behavior:
  Agent responses are not artificially limited below the selected model's actual output capacity.

Location:                                                                                         
  The issue originates in server/utils/agents/aibitat/providers/anthropic.js, where max_tokens: 4096
   is hardcoded in both the stream() and complete() methods (lines 192 and 319).                  

Fix:                                                                                            
  Replaces the hardcoded max_tokens: 4096 with a per-model lookup table mapping each Anthropic model
   to its actual API-enforced output token limit. Newer models (Claude 3.5+) get their full 8192    
  token capacity, legacy models stay safely at 4096, and any unknown or future models fall back to  
  4096 by default.  

Will make a PR for the fix.

Steps to reproduce:
  1. Configure AnythingLLM to use Anthropic with a Claude 3.5 or later model
  2. Start an agent session with a prompt that would naturally produce a long response (e.g.
  summarize a large document, or pass a large text block via a flow variable)
  3. Observe the response cuts off at approximately 4096 tokens

elevatingcreativity:
Closes #5039

## Bug

The Anthropic agent provider hardcodes `max_tokens: 4096` for all API calls. Modern Claude models (Claude 3.5 and later) support output limits well above 4096 tokens, but the hardcoded value was set when older models had smaller limits. As a result, agent responses are silently cut off at 4096 tokens regardless of which model is selected.

**Location:** `server/utils/agents/aibitat/providers/anthropic.js` — `max_tokens: 4096` is hardcoded in both the `stream()` and `complete()` methods.

## Fix

Replaces the hardcoded `max_tokens: 4096` with a `MODEL_MAX_OUTPUT_TOKENS` lookup table mapping each Anthropic model to its actual API-enforced output token limit. Claude 3.5 models use 8192, Claude 3.7 uses 64,000, legacy models retain 4096, and any unknown or future models fall back to 4096 safely.

This also prevents 400 API errors that would occur if a hardcoded value exceeded an older model's output limit — the Anthropic API rejects requests where `max_tokens` exceeds the model's maximum rather than silently clamping it.

## Test plan

- [x] Added Jest tests covering all model tiers (legacy, Claude 3.5, Claude 3.7, unknown)
- [x] Verified the correct `max_tokens` value is passed in actual API calls via mock
- [x] Run: `jest server/__tests__/utils/agents/aibitat/providers/anthropic.test.js`


elevatingcreativity:
Hi @timothycarambat I would like to bump this one, I didn't see it pulled in the latest release. It's a simple but important fix. I run into truncated output from Claude within AnythingLLM on many instances (several times today). It would be great to see this fix in the main branch of the code. Thanks for your consideration.

timothycarambat:
Moved this to be reworked in https://github.com/Mintplex-Labs/anything-llm/pull/5255 because it could be more dynamic so that we dont have to maintain a hardcoded list that will be tedious to maintain and will instantly accurate as new models roll out. 👍

PR #5073, opened 2026-02-26
fix: only scroll active sidebar items into view when not already visible

### LINKED ISSUE #4992: [FEAT]: Preserve scroll position

Image
Image

When clicking on a menu item lower on the screen, the sidebar shouldn't automatically scroll. The scroll position should be preserved

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #4992

### Description

Active sidebar items (workspace threads and settings menu options) would always scroll into center view on mount, causing jarring jumps even when the item was already visible in the scroll container.

This fix updates the `useScrollActiveItemIntoView` hook to check if the element is already fully visible within its nearest scrollable ancestor before calling `scrollIntoView`. It walks the DOM tree to find the actual scrollable parent (via `scrollHeight > clientHeight`) rather than matching by class name, making it work correctly across both the thread sidebar and settings sidebar (which has nested scroll containers).

### Developer Validations

### DIFF
--- a/frontend/src/hooks/useScrollActiveItemIntoView.js
+++ b/frontend/src/hooks/useScrollActiveItemIntoView.js
@@ -1,6 +1,19 @@
 import { useEffect, useRef } from "react";
 
 /**
+ * Walks up the DOM tree from `el` to find the nearest ancestor
+ * that is actually scrollable (i.e. its content overflows).
+ */
+function findScrollableParent(el) {
+  let node = el.parentElement;
+  while (node && node !== document.body) {
+    if (node.scrollHeight > node.clientHeight) return node;
+    node = node.parentElement;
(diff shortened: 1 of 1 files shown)


shatfield4 (changes requested):
The code looks good here but the behavior is still not ideal. The ideal behavior should be:
- Scroll position stays locked between switching pages

Currently if you click any item on the sidebar in settings, it works fine for any options that are not overflowing. But, if you scroll down below into the overflowing items and click them, the items still jump around causing an unnatural shift in ui.

Can we store a debounced scroll position in localstorage that can be used to scroll the sidebar on page load?

PR #5080, opened 2026-02-27
fix: persist agent and query-refusal chats to thread history via API

### LINKED ISSUE #5060: [BUG]: Agent interactions (@agent) are not persisted to thread history via API

AnythingLLM desktop app

### What happened?

When sending a chat request to the /v1/workspace/{slug}/thread/{threadSlug}/chat endpoint using the @agent trigger, the interaction (both prompt and response) is not saved to the thread history.

This issue breaks the continuity of the conversation when using AnythingLLM as a backend for custom applications, as the "memory" of the thread is lost for future interactions if those interactions were handled by the agent.

While the agent successfully executes its tasks and returns a response in the API call, the messages do not appear in the desktop UI for that thread, nor are they returned when fetching history via /v1/workspace/{slug}/thread/{threadSlug}/chats.

Steps to Reproduce
- Create a workspace and a specific thread.
- Send a POST request to /v1/workspace/{slug}/thread/{threadSlug}/chat with the message prefixing @agent (e.g., "message": "@agent search the customer database for...").
- Observe that the API returns a valid response from the agent.
- Check the thread history in the AnythingLLM Desktop App or via the /chats API endpoint.
- Result: The thread remains empty or missing the agent interaction.

Expected Behavior
Every interaction sent to a specific threadSlug should be persisted to the database and visible in the chat history, regardless of whether the @agent mode was triggered or a standard LLM response was generated.

Actual Behavior
- Standard chats (without @agent): Saved correctly in history.
- Agent chats (with @agent): Processed successfully but completely missing from the history/database.

Environment
- AnythingLLM desktop app v1.10.0
- LLM Provider: Google Gemini

_No response_

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5060

### Description

API chats via `chatSync` and `streamChat` had inconsistent behavior around UI visibility, thread association, and conversation history. This PR unifies both handlers so all code paths behave identically.

**Root causes:**
1. `chatSync` agent path was missing `threadId` and `user`, so agent chats were saved as orphan records with no thread association
2. `include` was hardcoded inconsistently — some paths used `true`, some `false`, some relied on the default (`true`). There was no unified logic
3. `recentChatHistory` always filtered `include: true`, which meant any chat saved with `include: false` (ephemeral session chats) was invisible to the LLM on subsequent calls — breaking conversation continuity for `sessionId`-based API usage

**Changes:**

1. **`apiChatHandler.js`** — Both `chatSync` and `streamChat` now derive `shouldIncludeInUI = !sessionId` at the top. Every `WorkspaceChats.new` call uses this value, and consistently passes `threadId`, `user`, and `apiSessionId`.

2. **`chats/index.js`** — `recentChatHistory` no longer filters on `include: true` when `apiSessionId` is present, so ephemeral session chats are still available as LLM context.

**Expected behavior with these changes:**

#### UI Visibility

| Condition | `include` | Visible in UI? |
|---|---|---|
| `sessionId` provided | `false` | No — ephemeral session, hidden from UI |
| `userId` provided, no `sessionId` | `true` | Yes — visible to that user |
| Neither provided (single-user mode) | `true` | Yes — no user scoping needed |
| Neither provided (multi-user mode) | `true` | Effectively no — `user_id` is `null` so no user's frontend query matches it |

#### LLM Conversation History

| Condition | LLM receives prior messages? | Scoped by |
|---|---|---|
| `sessionId` provided | Yes | `api_session_id` match |
| `userId` provided, no `sessionId` | Yes | `user_id` + `thread_id` + `include: true` |
| Neither provided | Yes (single-user) / No (multi-user, messages effectively vanish) | `thread_id` + `include: true` |

#### Consistency across handlers

| Code path | `chatSync` | `streamChat` |
|---|---|---|
| Agent | `shouldIncludeInUI`, `threadId` ✅, `user` ✅ | `shouldIncludeInUI`, `threadId` ✅, `user` ✅ |
| Query refusal (no embeddings) | `shouldIncludeInUI`, `threadId` ✅, `user` ✅ | `shouldIncludeInUI`, `threadId` ✅, `user` ✅ |
| Query refusal (no context) | `shouldIncludeInUI`, `threadId` ✅, `user` ✅ | `shouldIncludeInUI`, `threadId` ✅, `user` ✅ |
| Regular chat | `shouldIncludeInUI`, `threadId` ✅, `user` ✅ | `shouldIncludeInUI`, `threadId` ✅, `user` ✅ |

### Additional Information

- The frontend websocket chat handler (`stream.js`) is unaffected — it does not go through `apiChatHandler.js`
- Workspace-level API endpoints (`/v1/workspace/:slug/chat`) do not currently accept `userId` in the request body — `user` is always `null`. Only thread-level endpoints support `userId`.

### Developer Validations


server/utils/chats/apiChatHandler.js
```diff
@@ -230,7 +231,9 @@ async function chatSync({
         metrics: {},
       },
       include: false,
```
timothycarambat: why change this to true in the above but not here?

server/utils/chats/apiChatHandler.js
```diff
@@ -188,8 +188,9 @@ async function chatSync({
             type: chatMode,
             thoughts,
           },
-          include: false,
+          include: true,
           apiSessionId: sessionId,
+          threadId: thread?.id || null,
```
timothycarambat, line 206: I wonder why we hid these from the UI when originally added. I can imagine having every API chat appear in the UI in single user mode could be annoying - this also would make `sessionId` kind of useless now - since we can group them by thread now.

I believe the intention was:
- You can pass sessionId as basically a virtual thread to keep track of conversations ephemerally
- You can pass thread/workspace/user to automate chats for a user.

I think there are two use cases here that are now combined incorrectly:
- Ability to send ephemeral chats where history is controlled via sessionId matching.
- Ability to automate chats for a workspace/thread + user and have those chats also persist in the UI when using the full GUI.

For the first case `include` must still be false otherwise the UI will be heavily polluted. I think those two use cases need to be confirmed as functional in single/multi-user mode as at a glance this seems like it would break the intended use of both cases.

server/endpoints/api/workspaceThread/index.js
```diff
@@ -590,6 +606,19 @@ function apiWorkspaceThreadEndpoints(app) {
           return;
         }
 
+        if (userId && sessionId) {
```
shatfield4: Looks good here but we need to add this check to the workspace endpoints as well (`/v1/workspace/:slug/chat` and `/v1/workspace/:slug/stream-chat`). Possibly extract to util and reuse this logic?

angelplusultra: Those endpoints do not consume a userId only a session Id according to the schema

shatfield4 (approved):
LGTM

server/utils/chats/index.js
```diff
@@ -65,18 +65,20 @@ async function recentChatHistory({
   messageLimit = 20,
   apiSessionId = null,
 }) {
+  const where = {
```
timothycarambat: This is a bad name, should just be something like `conditions` or something more clear imo

server/utils/chats/index.js
```diff
@@ -65,18 +65,20 @@ async function recentChatHistory({
...
+    workspaceId: workspace.id,
+    user_id: user?.id || null,
+    thread_id: thread?.id || null,
+    api_session_id: apiSessionId || null,
+  };
+
+  // API chats using a sessionId are saved with include: false to keep them
+  // hidden from the UI. When fetching history for an API session, we omit
+  // the include filter so prior chats are still available for LLM context.
+  if (!apiSessionId) where.include = true;
```
timothycarambat: Include `include: true` to keep compatibility with the existing code prior just to be sure we are covering cases we might not be aware of.

then if apiSessionId is *not* `null` we can do `include: false`.

---

Actually, to clear something up:
- If a chat is sent with apiSessionId we hide it from the UI with include: false
- If a chat is sent with same session id, even though it is false, we should pull the historical chats anyway so that chat history exists

- What about people using sessionIDs right now where history is still true, but then upgrade. Their history is now reset for the same sessionID, correct?

In the above case, if that is correct, shouldnt we just omit/delete include totally from the query?

server/endpoints/api/workspaceThread/index.js
```diff
@@ -422,13 +423,27 @@ function apiWorkspaceThreadEndpoints(app) {
...
+            id: uuidv4(),
+            type: "abort",
+            textResponse: null,
+            sources: [],
+            close: true,
+            error:
+              "Cannot pass both userId and sessionId. Use userId to chat as a user or sessionId for ephemeral sessions.",
+          });
+          return;
+        }
```
timothycarambat: Make this a locally defined middleware check to the endpoint here to prevent code duplication since it is the same thing for both endpoints

server/utils/chats/index.js
angelplusultra: If I understand you correctly, I'm pretty sure that's what the code is doing. 

 When `apiSessionId` is present, `include` is omitted from the query entirely:

```js
// include not on conditions
  const conditions = {
    workspaceId: workspace.id,
    user_id: user?.id || null,
    thread_id: thread?.id || null,
    api_session_id: apiSessionId || null,
  };

// add include property to conditions object
  if (!apiSessionId) conditions.include = true;
```

So for non-session queries, it filters `include: true` (existing behavior). For session queries, it doesn't filter on include at all, meaning both old records (true) and new records (false) come back. No history reset on upgrade.

server/endpoints/api/workspaceThread/index.js
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/5080/commits/e7158f933fe13e288c5c3d0db6a59dbe19c25676

server/utils/chats/index.js
angelplusultra: Fix in https://github.com/Mintplex-Labs/anything-llm/pull/5080/commits/1f8b1e75af73d7f781dc62892dc70c07687396b9

PR #5136, opened 2026-03-04
fix: support OpenShift arbitrary UID/GID-0 in Docker image

petre:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Description

The Docker image currently fails to run on OpenShift (and similar platforms that use the restricted SCC) because OpenShift assigns an **arbitrary UID at runtime** while forcing **GID 0** (root group). The existing Dockerfile assumes the container will always run as the `anythingllm` user with a matching UID/GID, which causes permission errors on startup.

This PR makes the Docker image compatible with OpenShift's security model while preserving full backward compatibility with standard Docker and Docker Compose deployments.

**Dockerfile changes:**
- Add `anythingllm` user to the root group (GID 0) as a supplementary group, so files are accessible via group ownership under OpenShift's arbitrary-UID model
- Handle the `ARG_GID=0` case explicitly to avoid conflicting with the existing root group
- Set all `/app` files to `anythingllm:0` ownership with group-writable (`g+w`) permissions
- Set `chmod g=u /etc/passwd` so the entrypoint can inject a passwd entry for arbitrary UIDs
- Create additional directories (`/app/.cache`, `/app/.yarn`, `/app/server/storage`) upfront to avoid runtime permission failures
- Set `ENV HOME=/app` so tools like npm/yarn write to a writable location
- Run a final `chown -R` + `chmod -R g+w` after all `COPY` and `yarn install` steps to cover `node_modules` and build artifacts
- Apply changes to both the `amd64` and `arm64` build stages

**Entrypoint changes (`docker-entrypoint.sh`):**
- Detect when the running UID has no `/etc/passwd` entry (OpenShift arbitrary UID scenario) and dynamically inject one, so that tools like npm, Prisma, and git work correctly
- Export `HOME=/app` to ensure consistent behavior regardless of passwd state

**Compatibility matrix:**

| Platform | How access works |
|---|---|
| Docker / Docker Compose | Via user ownership (UID match) — unchanged behavior |
| OpenShift (restricted SCC) | Via group ownership (GID 0 + `g+w`) |
| Generic Kubernetes (with `fsGroup`) | Via user ownership or `fsGroup` — unchanged behavior |

### Visuals (if applicable)

N/A — infrastructure-only change, no UI impact.

### Additional Information

- No changes to application code; only `docker/Dockerfile` and `docker/docker-entrypoint.sh` are modified.
- Tested with `ARG_UID=1000 / ARG_GID=1000` (default Docker), `ARG_GID=0` (explicit root group), and OpenShift-style arbitrary UID with GID 0.

### Developer Validations

### DIFF
--- a/cloud-deployments/openshift/Dockerfile
+++ b/cloud-deployments/openshift/Dockerfile
@@ -1,3 +1,15 @@
+# OpenShift-compatible Dockerfile for AnythingLLM
+# 
+# This Dockerfile is specifically designed for OpenShift deployments which use
+# arbitrary UIDs with GID 0 (root group). Do NOT use this for standard Docker
+# or Docker Compose deployments - use the main docker/Dockerfile instead.
+#
+# Key differences from the standard Dockerfile:
+# - User is added to supplementary group 0 (root) for OpenShift compatibility
+# - All files are owned by group 0 and are group-writable (chmod g+w)
+# - /etc/passwd is made group-writable for dynamic UID injection
+# - Uses a modified entrypoint that handles arbitrary UID scenarios
+
 # Setup base image
 FROM ubuntu:noble-20251013 AS base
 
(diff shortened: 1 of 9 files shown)


timothycarambat:
For such a change, we should not be modifying the main docker entrypoint and dockerfile for a specific hosting provider. This should either be on a fork or a branch so that it does not interfere with existing images

petre:
Thanks for the feedback @timothycarambat. 

This isn't OpenShift-specific — it's a POSIX/OCI best practice. The arbitrary UID + GID 0 pattern is part of the [OCI image specification guidelines](https://docs.openshift.com/container-platform/latest/openshift_images/create-images.html#use-uid_create-images) and is used by many platforms beyond OpenShift: Podman rootless, Kubernetes with runAsNonRoot + runAsUser, Rancher, and any security-hardened container runtime that refuses to run as the UID baked into the image.

The changes are fully backward-compatible. As shown in the compatibility matrix, Docker and Docker Compose behavior is completely unchanged — access still works via UID ownership as before. The only addition is that GID-0 group permissions are also set, which is a no-op on standard Docker. There is no interference with existing images.

The entrypoint modification is minimal and completely inert on standard Docker. The passwd entry injection is only a few lines and is gated behind a condition that checks whether the current UID exists in /etc/passwd. On regular Docker and Docker Compose — where the container always runs as the anythingllm user baked into the image — this condition is never true, so the code path is never executed. It's purely a safety net, not a change in default behavior.

timothycarambat:
@petre Ah, thank you for the explanation - I will take a closer look. We once had something like this with the GUID and it absolutely blew up everything which is why I am cautious about it!

timothycarambat:
This PR modifies the main docker/Dockerfile and docker-entrypoint.sh in ways that affect every Docker deployment globally, not just OpenShift users. We have millions of users running AnythingLLM via standard Docker and Docker Compose, and I'm not willing to take even the tinyest risk breaking their deployments to support a single platform's unique requirements.

I've moved these changes to a dedicated OpenShift community template at cloud-deployments/openshift/ which includes:

- An OpenShift-specific Dockerfile (your changes)
- The modified entrypoint script
- Documentation explaining when and how to use it

This follows the same pattern as our other cloud deployment templates (AWS, Azure, GCP, Railway, Render) and allows OpenShift users to build a compatible image without imposing OpenShift-specific requirements on the entire user base.

OpenShift users can build from the repo root with:
```
docker build -f cloud-deployments/openshift/Dockerfile -t anythingllm:openshift .
```

I hope this is an acceptable compromise so OpenShift gets first-class support via a dedicated template, and existing Docker users aren't affected or at least I do not have to personally incur the wrath of breaking every persons deployment globally from changing permissions scope for containers. I am by no means a Docker expert and that is very likely where this anxiety of a change like this comes from, having been burned by "small changes to Dockerfile" a lot time ago in this project, haha.

PR #5143, opened 2026-03-04
Automatic mode for workspace (Agent mode default)

### LINKED ISSUE #4509: [UX Enhancement]: Add Toggle in Chat Bar to Automatically Enable `@agent` Mode

Currently, users must manually prepend @agent to every message in order to enable Agent Mode (which allows tool calling). 

Proposal:
Introduce a toggle switch in the chat interface (e.g., near the input bar) that, when enabled, automatically prepends `@agent` to the prompt before the message is sent.

Expected Behavior:
	•	When the toggle is on, messages are automatically sent in Agent Mode (equivalent to manually typing `@agent`...).
	•	When the toggle is off, messages are sent as normal.
	•	The toggle’s state could persist across sessions or reset per chat, depending on UX preferences.

Example:

Instead of:

`@agent summarize this document`

The user could simply type:

`summarize this document`

**—with Agent Mode toggle turned on.**

### LINKED ISSUE #4932: [FEAT]: Add "Always-On Agent" Mode for Automatic Tool Invocation Without @agent Prefix

First, thank you for building such a powerful and flexible platform! The custom agent skill system works great once activated.
However, we'd like to request a new "Always-On Agent" mode that allows the LLM to automatically analyze user intent and invoke custom skills without requiring the user to type @agent.

**Current Behavior**
Currently, the embedded chat widget does not support invoking agent skills at all, and regular end-users are generally unaware of — or unfamiliar with — the concept of AI "skills" or the need to use a special command like @agent to trigger them.

**Proposed Feature**
Add a workspace-level toggle: "Enable Always-On Agent" (off by default for safety).
When enabled, the LLM continuously evaluates whether a custom skill should be invoked based on user input—just like in systems such as Claude or Cursor.
All existing security and logging mechanisms (e.g., this.logger, skill permissions) would still apply.

**Use Case**
We’ve built a custom skill to send inquiry emails to our sales team when a user requests follow-up. We want this to happen seamlessly—e.g., when a user says “Please have someone contact me about pricing,” the agent should auto-trigger the email skill without any special syntax.

This would significantly improve usability in production deployments (especially via the embeddable chat widget) while maintaining control via explicit opt-in at the workspace level.
Thanks for considering this enhancement!

### LINKED ISSUE #5068: [FEAT]: Allow Agent Skills to Access Images Attached or Pasted in Chat
## Context
I'm building a custom agent skill for AnythingLLM that edits images using an external API (Replicate) to call an AI Image editing model. The ideal user experience is for users to simply attach or paste an image into the chat and have the agent skill process it automatically based on their prompt (send the input image and the prompt to the API).

## Current Situation
Currently, when a user sends a message to an agent skill (e.g., using `@agent`), only the text content of the message is passed to the skill handler. Any images attached or pasted into the chat are not accessible to the skill—there is no URL, file path, or base64 data provided. As far as I know.

## Problem
This limitation means users must manually convert images to base64 or upload them to a specific folder, which adds friction and makes the workflow less user-friendly. The skill cannot process images directly from the chat, which is a common and expected use case for image editing or analysis skills.

## Questions
1. Is there currently any way for agent skills to access images that users attach or paste into the chat?
2. If not, are there any recommended workarounds for this use case that don't add a lot of user friction?
3. Are there plans to support passing image data (URL, file path, or base64) to agent skills in the future?

## Desired Feature
It would be very useful and helfpuf if you could provide a way for agent skills to access images that users attach or paste into the chat, ideally as a URL, file path, or base64 string, so that skills can process them directly and provide a seamless user experience for image-based workflows.

Thanks in advance for your feedback.

timothycarambat:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

closes #3483
resolves #4932
resolves #4509
resolves #5068

### Description

Introduces a new automatic chat mode (now the default) that automatically invokes tools when the provider supports native tool calling. 

- Conditionally shows/hides the @agent command based on whether native tooling is available in the prompt input
- Update all locale translations with new mode descriptions (#5145)

Also supports automatic injection of attached documents and images to providers.

### Additional Information
 - Add supportsNativeToolCalling() to AI providers (OpenAI, Anthropic, Azure always support; others opt-in via ENV)
- Enhance translator to preserve Trans component tags
- Remove deprecated ability tags UI
- There was a lot of work on aibitat core to get this to function for all providers as well as persist messages, images, attachments, and possible stream errors for tooled/untooled and niche providers

### Developer Validations

### DIFF
--- a/server/utils/agents/ephemeral.js
+++ b/server/utils/agents/ephemeral.js
@@ -8,6 +8,7 @@ const { User } = require("../../models/user");
 const { Workspace } = require("../../models/workspace");
 const { WorkspaceChats } = require("../../models/workspaceChats");
 const { WorkspaceParsedFiles } = require("../../models/workspaceParsedFiles");
+const { DocumentManager } = require("../DocumentManager");
 const { safeJsonParse } = require("../http");
 const {
   USER_AGENT,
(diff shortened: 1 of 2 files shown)


server/models/workspace.js
```diff
@@ -205,6 +207,7 @@ const Workspace = {
       const workspace = await prisma.workspaces.create({
         data: {
           name: this.validations.name(name),
+          chatMode: "automatic",
```
timothycarambat: Do this to prevent full drop-table migration

server/utils/agents/aibitat/providers/helpers/tooled.js
```diff
@@ -35,10 +35,41 @@ function formatFunctionsToTools(functions) {
...
+  }
+
+  // Transform message with attachments into multimodal format
+  const content = [{ type: "text", text: message.content }];
+  for (const attachment of message.attachments) {
+    content.push({
+      type: "image_url",
+      image_url: {
+        url: attachment.contentString,
+      },
```
timothycarambat, line 56: Always assuming OAI format since that is what 90% of the providers use

server/utils/agents/aibitat/providers/ai-provider.js
```diff
@@ -82,6 +82,14 @@ class Provider {
     return this._client;
   }
 
+  /**
+   * Whether this provider supports native OpenAI-compatible tool calling.
+   * @returns {boolean|Promise<boolean>}
+   */
+  supportsNativeToolCalling() {
+    return false;
+  }
```
timothycarambat, line 138: This should always have a per-provider override

timothycarambat:
Blocked by #5074 since will have massive merge confs once merged.

bhtabor:
Thank you @timothycarambat! Does this support automatic chat mode on the API? I noticed that `automatic` is a valid chat mode for the API but native tool calling was not working. This is the fix that worked for me -https://github.com/bhtabor/anything-llm/tree/automatic-mode-api-fix

timothycarambat:
@bhtabor - 1aaa44aa2f4a55b15f3868690edbec1ccf39f54b thank you for that 👍 

Need to do some additional work as it doesnt support passthrough of the attachments via API, but that shouldnt be too hard to do.

bhtabor:
Hi @timothycarambat, I’m sure you have a release process in place, but do you happen to have an ETA for the next version bump that include this commit?

timothycarambat:
> Hi @timothycarambat, I’m sure you have a release process in place, but do you happen to have an ETA for the next version bump that include this commit?

Should be this week, working on it right now actually

PR #5058, opened 2026-02-24
feat: LLM API (llmapi.ai) integration

MarkosBK:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)
- [ ] ♻️ refactor (Code refactoring without changing behavior)
- [ ] 💄 style (UI style changes)
- [ ] 🔨 chore (Build, CI, maintenance)
- [ ] 📝 docs (Documentation updates)

### Relevant Issues

resolves #5057

### Description

Support code for new LLM provider - LLM API (https://llmapi.ai)

### Visuals (if applicable)

image

### Additional Information

The `/v1/models` endpoint does not require authentication, so the model
list is available before the user enters their API key.

### Developer Validations

- [x] I ran `yarn lint` from the root of the repo & committed changes
- [x] Relevant documentation has been updated (if applicable)
- [x] I have tested my code functionality
- [x] Docker build succeeds locally

PR #5148, opened 2026-03-05
feat: LLM API (llmapi.ai) integration

### LINKED ISSUE #5057: [FEAT]: LLM API (llmapi.ai) integration
## Problem
AnythingLLM doesn't support [LLM API](https://llmapi.ai) — an AI model aggregator with an OpenAI-compatible API.

## Solution

Add LLM API as a supported LLM provider.

**Details:**
- Base URL: `https://internal.llmapi.ai/v1`
- OpenAI-compatible API (drop-in replacement, no special SDK needed)
- Dynamic model loading from `/v1/models` (no API key required to fetch models)
- Supports chat, RAG, and agents

MarkosBK:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5057

### Description

Support code for new LLM provider - LLM API (https://llmapi.ai)

### Visuals (if applicable)

image

### Additional Information

The `/v1/models` endpoint does not require authentication, so the model
list is available before the user enters their API key.

### Developer Validations


MarkosBK:
@timothycarambat #5058 resolves

frontend/src/components/ProviderPrivacy/constants.js
```diff
@@ -252,6 +253,11 @@ const LLM_PROVIDER_PRIVACY_MAP = {
     ],
     logo: LemonadeLogo,
   },
+  llmapi: {
+    name: "LLM API",
+    policyUrl: "https://llmapi.ai/privacy-policy",
```
timothycarambat: Not a valid policy URL

server/utils/agents/aibitat/providers/llmapi.js
```diff
@@ -0,0 +1,85 @@
...
+
+  async stream(messages, functions = [], eventHandler = null) {
+    return await UnTooled.prototype.stream.call(
+      this,
+      messages,
+      functions,
+      this.#handleFunctionCallStream.bind(this),
+      eventHandler
+    );
+  }
```
timothycarambat, line 131: Does LLMAPI support native tool calling? Using `Untooled` basically makes a model tool-calling capable and is basically a legacy patch since we have been around since before tool calling ever existed. If the API accepts a regular OpenAI Spec tools/functions that should be used as the performance of tooling is 10x better.

server/utils/AiProviders/llmapi/index.js
```diff
@@ -0,0 +1,287 @@
...
+  LLMPerformanceMonitor,
+} = require("../../helpers/chat/LLMPerformanceMonitor");
+const { v4: uuidv4 } = require("uuid");
+const {
+  writeResponseChunk,
+  clientAbortedHandler,
+} = require("../../helpers/chat/responses");
+
+const LLMAPI_BASE_URL = "https://api.llmapi.ai/v1";
+const LLMAPI_DEFAULT_CONTEXT_WINDOW = 131072;
```
timothycarambat: Not configurable by user or detectable via model tag on API?

server/utils/AiProviders/llmapi/index.js
```diff
@@ -0,0 +1,287 @@
...
+const {
+  LLMPerformanceMonitor,
+} = require("../../helpers/chat/LLMPerformanceMonitor");
+const { v4: uuidv4 } = require("uuid");
+const {
+  writeResponseChunk,
+  clientAbortedHandler,
+} = require("../../helpers/chat/responses");
+
+const LLMAPI_BASE_URL = "https://api.llmapi.ai/v1";
```
timothycarambat: Not reused, can just be in the constructor client creation.

server/utils/AiProviders/llmapi/index.js
```diff
@@ -0,0 +1,287 @@
...
+    if (!process.env.LLMAPI_LLM_API_KEY)
+      throw new Error("No LLM API key was set.");
+    this.className = "LLMApiLLM";
+    const { OpenAI: OpenAIApi } = require("openai");
+
+    this.openai = new OpenAIApi({
+      baseURL: LLMAPI_BASE_URL,
+      apiKey: process.env.LLMAPI_LLM_API_KEY,
+      defaultHeaders: {
+        "x-source": "anythingllm",
```
timothycarambat: We have a method called `getAnythingLLMUserAgent()` that might be useful here.

server/utils/AiProviders/llmapi/index.js
```diff
@@ -0,0 +1,287 @@
...
+      messages,
+      runPromptTokenCalculation: false,
+      modelTag: this.model,
+      provider: this.className,
+    });
+
+    return measuredStreamRequest;
+  }
+
+  async handleStream(response, stream, responseProps) {
```
timothycarambat, line 229: Is the API truly OpenAI spec? Might be able to just use the traditional `handleDefaultStreamResponseV2` function to handle streams if the usage schema and response schema is chat/completions schema exactly.

server/utils/AiProviders/llmapi/index.js
```diff
@@ -0,0 +1,287 @@
...
+    const handleAbort = () => {
+      stream?.endMeasurement(usage);
+      clientAbortedHandler(resolveStream, fullText);
+    };
+    response.on("close", handleAbort);
+
+    try {
+      for await (const chunk of stream) {
+        const message = chunk?.choices?.[0];
+        const token = message?.delta?.content;
```
timothycarambat: Just asking, does this API support thinking models, and if does - are those chunks coming back via delta content or in a separate reasoning block?

server/utils/helpers/customModels.js
```diff
@@ -1004,6 +1007,44 @@ async function getSambaNovaModels(_apiKey = null) {
...
+ */
+async function getLLMApiModels(_apiKey = null) {
+  try {
+    const apiKey =
+      _apiKey === true
+        ? process.env.LLMAPI_LLM_API_KEY
+        : _apiKey || process.env.LLMAPI_LLM_API_KEY || null;
+
+    const headers = { "Content-Type": "application/json" };
+    if (apiKey) headers["Authorization"] = `Bearer ${apiKey}`;
```
timothycarambat: No auth on the endpoint so no need to pass `apiKey` at all into this function

server/utils/helpers/customModels.js
```diff
@@ -1004,6 +1007,44 @@ async function getSambaNovaModels(_apiKey = null) {
...
+
+    const response = await fetch("https://api.llmapi.ai/v1/models", {
+      headers,
+      signal: AbortSignal.timeout(5000),
+    });
```
timothycarambat: Since spec if OpenAI compatible, you can just use OpenAI client and `.models()` with the custom BaseURL instead of a custom fetch call

server/utils/AiProviders/llmapi/index.js
timothycarambat: Just checked and some models do indeed have `context_length` and that should be retrieved here dynamically with a new class method `assertContextLength` to lazy-load the context window on demand (see ollama or lmstudio on how this is done).

This ensures we can properly manage the window for the mode. The current default of `131072` is larger than the API stated 128K for `gpt-4o-mini` which I would not be surprised totally breaks if a larger window is sent than is possible to be called.

timothycarambat (changes requested):
I cant comment on an image, but the provider image is super small and should be:
- at least 330x330
- white bg (if provider does not have brand color bg)

frontend/src/components/ProviderPrivacy/constants.js
MarkosBK: Provider logo — replaced with a 330×330px version with white background.
Privacy policy URL — updated to https://llmapi.ai/terms.

server/utils/agents/aibitat/providers/llmapi.js
MarkosBK: UnTooled → native tool calling — refactored aibitat/providers/llmapi.js to follow the same pattern as groq.js: supports native OpenAI-compatible tool calling when PROVIDER_SUPPORTS_NATIVE_TOOL_CALLING=llmapi is set in ENV, falls back to UnTooled otherwise. Added llmapi to the example in .env.example.

server/utils/AiProviders/llmapi/index.js
MarkosBK: Context window — removed the hardcoded constant. Added cacheContextWindows() that fetches context_length per model from /v1/models on startup (159/161 models expose this field), and assertModelContextLimits() for lazy initialization. Falls back to 128 000 for models without it.

server/utils/AiProviders/llmapi/index.js
MarkosBK: moved inline to the constructor.

server/utils/AiProviders/llmapi/index.js
MarkosBK: getAnythingLLMUserAgent() — replaced the custom x-source header.

server/utils/AiProviders/llmapi/index.js
MarkosBK: handleStream — yes, the API is OpenAI-compatible. However a custom handleStream is required — see comment about thinking models.

server/utils/AiProviders/llmapi/index.js
MarkosBK: LLMAPI proxies reasoning models from multiple providers, each with different streaming behavior. Some put their entire response in delta.reasoning with no delta.content at all, and some don't send finish_reason. Using handleDefaultStreamResponseV2 results in empty responses or hanging promises for these models. The custom handleStream handles both fields (delta.content || delta.reasoning) and includes a fallback close when no finish_reason is received — consistent with how openRouter/index.js solves the same problem.

MarkosBK:
@timothycarambat hi, I fixed that. Can you review again pls

Issue #4352, opened 2025-09-01
[FEAT]: Contextual RAG & Document Structuring Improvements 🚀


trade-magnet:
### What would you like to see?

First off – bloody amazing product 🙌. Absolutely love AnythingLLM. A huge kudos to the team.  

I genuinely think that for the success of this product to keep growing, its biggest strength (and main differentiator) is in the **RAG component**. This is essentially why you’d choose AnythingLLM over something like an OpenWebUI. Strengthening how retrieval works is what will really set this project apart from other products in the space.  

That said, I’d love to suggest a couple of improvements around **contextual RAG and semantic search**:  

### 1. Document Structuring Before Embedding  
At the moment, when docs are uploaded, it looks like they’re just scraped top-to-bottom without much structure. This can hurt retrieval quality.  

**Suggestion:**  
- Allow the ability to plug in an external API endpoint (e.g. Mistral’s OCR/structuring endpoint) that can transform the document into clean Markdown or structured text *before* embeddings are generated.  
- This extra step would massively improve consistency and reliability of the RAG pipeline.  

### 2. Contextual RAG / Chunk Enrichment  
One of the biggest wins I’ve found in my own testing is **contextualizing each chunk**. For example:  
- Appending a contextual sentence at the beginning of each chunk (summarizing what the chunk relates to in the broader doc).  
- Selecting how many pages/sections of a doc should be used for contextual grounding (e.g. first 10 pages only, or user-defined).  

This approach has boosted context awareness for me dramatically — in some cases, it’s the difference between a vague answer and a highly relevant one.  

### 3. Future Ideas (Optional)  
- Knowledge graph integration to map relationships between entities across docs.  
- Advanced page/chunk selection options for power users.  

---

Even if AnythingLLM doesn’t go all-in on contextual RAG, I think doubling down here would be a huge win. It’s the RAG component that makes this tool shine — and a few targeted improvements could push retrieval performance through the roof.  

Thanks again for such an awesome project 🚀

timothycarambat:
> Allow the ability to plug in an external API endpoint (e.g. Mistral’s OCR/structuring endpoint) that can transform the document into clean Markdown or structured text before embeddings are generated.
This extra step would massively improve consistency and reliability of the RAG pipeline.

Agreed.

> One of the biggest wins I’ve found in my own testing is contextualizing each chunk. For example:
> 
[...]

Issue #4072, opened 2025-06-30
RAG retrieval process chat history (previous issue)


Alminc91:
I am referencing your comment on the closed issue #4071 
I will formulate my question better and answer. Thanks for your help:

I tried it with 32k context window and it is only one follow-up question (so maybe about 6-8k tokens).
I use Qwen2.5-72B with AWQ 4-bit quantization. 

It still fails at that. It seems this technique is called **Conversational RAG**.
Maybe another example is helpful:

I ask for three courses with instructors. In the next prompt, I ask for more courses from the second instructor. This works with providers like Chatbase.

I assume: they query the model two times with the whole context for a query rewrite, so this is a bit more complex and not easily integrated into AnythingLLM, I just assume. 

So that would be a system issue, but quite complex and more for commercial, advanced systems? Thank you.

P.S.: My mistake previously: The courses were not hallucinated, they exist. They are just not sports courses.

The original comment:

> This isn't user error nor an issue with the system, this is a model issue! The hallunications are the model behavior, likely the model is too small or over-quantized to get coherent results. In general, small over-quantized models get _worse_ the longer a chat goes as the context window fills up.
> 
> As for RAG results, each query gets its own results. If none are found, previous citations will backfill this. Then, on top of this, the chat history is always included. **However** this is all assuming your context window is not tiny. To not overflow the context window, crashing the model, we will auto-prune the history. If the context window for your model is set to a low value youll see behavior like this.
> 
> You didnt mention what provider or model or system specs so I am making assumptions that usually explain the above for most people 

 _Originally posted by @timothycarambat in [#4071](https://github.com/Mintplex-Labs/anything-llm/issues/4071#issuecomment-3019259082)

timothycarambat:
> I ask for three courses with instructors. In the next prompt, I ask for more courses from the second instructor. This works with providers like Chatbase.

They are putting the entire context into the chat window. This is a huge burden on the system running the LLM because you are injecting tokens that ultimately are "useless" to have in context. However, you get more coherent responses when requiring full-text comprehension. This can be solved by [pinning the document](https://docs.anythingllm.com/llm-not-using-my-docs#document-pinning).

We dont do this by default since it blows up your context window needlessly, but it will give you better results. An "inbetween" solution is just to change the chat search preference to "Accuracy optimized" which will then rerank document snippets - this gives you more context than normal rag without injecting the entire corpus of text
[...]

Issue #4071, opened 2025-06-30
[FEAT]: RAG retrieval process misses reference to chat history


Alminc91:
### What would you like to see?

Dear AnythingLLM team,

I have a feeling that the RAG pipeline does not reference the chat history the LLM model is aware of.

For example, imagine that I have fed data into my chatbot for an online course adviser. Let's assume that I have multiple sports courses in my repertoire.
When I ask for sports courses, the chatbot lists three courses.

However, if I then ask for more, the sources provided are completely irrelevant to my previous request. The LLM knows the chat history, but the RAG pipeline for retrieving suitable sources does not reference it.

My example:

Here, I ask for sports courses in German. The sources and the answer are relevant and relate to the request.

![Image](https://github.com/user-attachments/assets/b8ebebe6-8f1a-4ce7-b91d-f01f739c7ff7)

Now, I ask for some more courses (related to the last request).

![Image](https://github.com/user-attachments/assets/19931748-3065-4802-8d53-c4cd253d266c)

In this case, the LLM tries to recommend additional sports courses that do not exist. Therefore, it hallucinates easily. This is because the sources are arbitrary; in my case, some random computer courses are retrieved.

This makes the chatbot far less useful for real conversations as I cannot really refer to previous requests using RAG.
I think it would be a valuable addition to AnythingLLM. Would this be very difficult? If there are any suggestions on how to do it, I could try to approach the problem.

Or is it just a user error on my part? 

Thank you very much for your help and for the great software!

timothycarambat:
This isn't user error nor an issue with the system, this is a model issue! The hallunications are the model behavior, likely the model is too small or over-quantized to get coherent results. In general, small over-quantized models get _worse_ the longer a chat goes as the context window fills up.

As for RAG results, each query gets its own results. If none are found, previous citations will backfill this. Then, on top of this, the chat history is always included. **However** this is all assuming your context window is not tiny. To not overflow the context window, crashing the model, we will auto-prune the history. If the context window for your model is set to a low value youll see behavior like this.

You didnt mention what provider or model or system specs so I am making assumptions that usually explain the above for most people

PR #5171, opened 2026-03-08
feat: LLM-based query rewriting for contextual RAG

Alminc91:
## Problem

In multi-turn conversations, follow-up queries lack context for vector search, causing RAG to retrieve irrelevant or no results.

**Example conversation:**
1. User: "What German courses do you offer?" → RAG finds German courses ✅
2. Bot: "Here are some German courses... Are you looking for A1 or B1?"
3. User: "Yes, B1" → RAG searches for "Yes, B1" → finds **nothing** ❌
4. User: "Is this course also available on weekends?" → RAG searches literally → **nothing** ❌

The chat history is already loaded and sent to the LLM for response generation — but it is **not used for the vector search**. The raw user query goes directly into `performSimilaritySearch`, which fails for context-dependent follow-ups.

This has been reported multiple times:
- #4352 (Contextual RAG)
- #4071 (RAG retrieval misses chat history reference)
- #4072 (follow-up)

## Solution

A lightweight LLM-based query rewriter that runs **before** vector search. It takes the user's message + recent chat history and rewrites it into a standalone search query.

### How it works
1. **Trigger**: Only when chat history exists AND query is ≤12 words (configurable). First messages and long self-contained queries skip rewriting entirely.
2. **Rewrite**: Sends a short prompt (~550 tokens total) with the last 2 turn-pairs to the already-loaded LLM. Assistant messages are truncated to 150 words to keep input small.
3. **Use**: The rewritten query goes to `performSimilaritySearch`. The **original** message is still used for the LLM chat completion and stored in chat history.
4. **Fallback**: On any error, the original query is used unchanged. Zero risk of breaking existing behavior.

### Selective rewriting with UNCHANGED signal

Self-contained queries (ones that already contain their own subject/topic) are **not rewritten**. The LLM responds with a single token `UNCHANGED` instead of reproducing the full query, which provides two benefits:

- **Latency**: 1 output token (~125ms) vs. reproducing the query (5-15 tokens, ~300ms) — roughly **40% faster** for queries that don't need rewriting
- **Correctness**: No risk of the LLM subtly altering a self-contained query (e.g., changing "What yoga courses are there?" to "Which yoga courses are available?" — same meaning, different search vector)

### 3-layer output validation (model-agnostic)

LLM output is non-deterministic. Different models (GPT-4, Mistral, Llama, Qwen, local models) may produce meta-responses like "no rewrite needed" or "The query is already self-contained" instead of clean output. If such text reaches vector search, it returns nothing — a silent failure.

The validation guarantees that `rewriteQueryForSearch()` only ever returns the **original query** or a **valid rewritten query**, never meta-text:

| Layer | Check | Catches |
|-------|-------|---------|
| 1 | `startsWith("UNCHANGED")` | Explicit signal from compliant LLMs (fast path, 1 token) |
| 2 | Verbatim string comparison | LLM copies the query exactly (fallback for less capable models) |
| 3 | Content word overlap | Valid rewrites share topic words (>3 chars) with the conversation context. Meta-responses like "no rewrite needed" do not. |

Layer 3 is the core safety net — it works across all languages and models without keyword lists. A valid rewrite necessarily contains subject/topic words from the conversation ("German courses", "B1", "weekends"), while meta-responses contain only generic words ("rewrite", "needed", "query") that never appear in the conversation context.

**Universal script support**: For non-space-separated scripts (CJK, Thai, etc.), word-level matching is supplemented with character-level overlap (characters with charCode > 127). This ensures the validation works for Chinese, Japanese, Korean, Thai, and other languages where spaces don't delimit words.

### Production results

Tested in production across 65 RAG chatbot instances. Actual rewrite logs:

```
[QueryRewrite] "Yes, B1 please"                            → "German courses B1 integration course"      (339ms)
[QueryRewrite] "Is this course also available on weekends?" → "Integration courses German B1 on weekends" (258ms)
[QueryRewrite] "What about regular German courses?"         → "German courses B1 on weekends"             (219ms)
[QueryRewrite] "Let's search for Spanish beginner courses!" → "UNCHANGED"                                 (125ms)  ← self-contained, no rewrite
[QueryRewrite] "Maybe I'll try B1 after all!"               → "Spanish courses B1 advanced"               (369ms)
[QueryRewrite] "Is that the only one???"                    → "Spanish courses B1 available"               (332ms)
```

**Overhead: 125–370ms per query** (~550 tokens in, 1-15 tokens out). Self-contained queries cost ~125ms (1 token). The rewrite uses the same LLM that's already loaded for chat — no additional model required.

## Configuration

Three environment variables, no database changes:

| Variable | Default | Description |
|----------|---------|-------------|
| `DISABLE_QUERY_REWRITING` | `false` | Kill switch to disable entirely |
| `QUERY_REWRITE_WORD_THRESHOLD` | `12` | Queries above N words skip rewriting |
| `QUERY_REWRITE_MAX_HISTORY` | `2` | Number of turn-pairs used for context |

## Changes

| File | Change |
|------|--------|
| `server/utils/helpers/chat/queryRewriter.js` | **New** — rewrite logic, 3-layer validation, universal script support |
| `server/utils/chats/embed.js` | Import + rewrite before vector search |
| `server/utils/chats/stream.js` | Import + rewrite before vector search |
| `server/utils/chats/apiChatHandler.js` | Import + rewrite before vector search (2 handlers) |
| `server/utils/chats/openaiCompatible.js` | Import + rewrite before vector search (2 handlers) |

All 6 `performSimilaritySearch` call sites across all chat interfaces (embed widget, workspace UI, API sync, API stream, OpenAI-compatible sync/stream) are covered.

## Why this approach

| Alternative | Why not |
|------------|---------|
| String concatenation (keywords from history) | Fragile, unreliable for real conversations |
| Embed history + query together | Dilutes the embedding signal |
| Conditional on word count only | Misses context-dependent longer queries ("Is **this** course available on weekends?") |
| Always rewrite | Wastes compute on first messages and long self-contained queries |
| Trust LLM output directly | Different models produce unpredictable meta-text — needs validation |

The LLM-based approach with a word-count threshold is the industry standard (LangChain's `create_history_aware_retriever`, Open WebUI's search query generation). This implementation goes further with selective rewriting (UNCHANGED signal) and model-agnostic output validation (3-layer check). One file, ~115 lines of logic, no dependencies.

### DIFF
--- a/server/utils/helpers/chat/queryRewriter.js
+++ b/server/utils/helpers/chat/queryRewriter.js
@@ -86,7 +86,18 @@ async function rewriteQueryForSearch({ userQuery, chatHistory, LLMConnector }) {
       .toLowerCase()
       .split(/\s+/)
       .filter((w) => w.length > 3);
-    const hasOverlap = rewriteWords.some((w) => contextWords.has(w));
+    let hasOverlap = rewriteWords.some((w) => contextWords.has(w));
+
+    // Fallback for non-space-separated scripts (CJK, Thai, etc.):
+    // Word-level matching fails because these languages don't use spaces.
+    // Individual characters carry meaning, so check for shared characters.
+    if (!hasOverlap) {
+      const contextChars = new Set([...(userQuery + historyText)]);
+      hasOverlap = [...rewritten].some(
+        (c) => c.charCodeAt(0) > 127 && contextChars.has(c)
+      );
+    }
+
     if (!hasOverlap) {
       console.log(
         `\x1b[35m[QueryRewrite]\x1b[0m Rejected meta-response: "${rewritten}"`


timothycarambat:
We will not be doing this issue as it introduces a massive change to the core of the system that is solved via reranking mostly. I know this process or processes similar to this can be useful but with smaller on device models this will add a lot of latency for a non-deterministic output that could even be worse in some instances. Something this large in change should be maintained on an independent fork!

PR #5190, opened 2026-03-11
Telegram bot connector

### LINKED ISSUE #5189: [FEAT]: Telegram connection/remote control

- Build Telegram connector per internal spec

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5189

### Description

- Add telegram bot integration as a new page under "Channels" -> "Telegram"
- Add QR code to take user to BotFather to create new bot
- User verification system with pairing codes
- Voice message support with multiple response modes (text only, mirror, always voice) *hidden for now*
- Photo/media handling for Telegram chats
- Chat workspace and thread switching via Telegram slash commands (/switch, /history)
- Model change support in chat
- `Auto` chat mode support and native tool calling
- @agent support with custom PNG chart renderer for telegram
- Bot token encryption
- Offloads Telegram chat processing to background worker jobs
- Kills Telegram bot on enabling of multiuser mode (Telegram bot only active in single user/single user password protected modes)
- Reconnect UI if token expires or becomes invalid
- Rendering of rich text responses for markdown elements
- handling of `think` blocks and split-block handling
- Full user GUI for management of connection and approved users

### Telegram Commands
- /start - Welcome message with current workspace info
- /switch - Switch workspace or thread
- /new - Start a new thread in the current workspace
- /history [count] - Show recent messages (default 10, max 50)
- /status - Show current workspace, thread, provider, and model
- /reset - Clear chat history context for the LLM (messages stay visible but aren't used as context)
- /proof - show citations from previous chat
- /model - select from model list to change mid chat
- /help - Show available commands

### Developer Validations

### DIFF
--- a/frontend/src/locales/lt/common.js
+++ b/frontend/src/locales/lt/common.js
@@ -51,7 +51,6 @@ const TRANSLATIONS = {
   },
   common: {
     "workspaces-name": "Darbo srities pavadinimas",
-    user: "Vartotojas",
     selection: "Modelio pasirinkimas",
     saving: "Saugoma...",
     save: "Išsaugoti pakeitimus",
@@ -112,6 +111,10 @@ const TRANSLATIONS = {
     contact: "Susisiekti su pagalba",
     "browser-extension": "Naršyklės plėtinys",
     "mobile-app": "AnythingLLM mobiliesiems",
+    channels: "Kanalai",
+    "available-channels": {
+      telegram: "„Telegram“",
+    },
   },
   login: {
     "multi-user": {
@@ -195,18 +198,12 @@ const TRANSLATIONS = {
       title: "Pokalbio režimas",
       automatic: {
         title: "Auto",
-        description:
-          "automatiškai naudos įrankius, jei modelis ir tiekėjas palaiko natūralų įrankių iškvietimą.<br />Jei natūralus įrankių iškvietimas nepalaikomas, norėdami naudoti įrankius turėsite naudoti @agent komandą.",
       },
       chat: {
         title: "Pokalbis",
-        description:
-          "pateiks atsakymus remdamasis LLM bendrosiomis žiniomis <b>ir</b> rastu dokumentų kontekstu.<br />Norėdami naudoti įrankius, turėsite naudoti @agent komandą.",
       },
       query: {
         title: "Užklausa",
-        description:
-          "pateiks atsakymus <b>tik</b> jei bus rastas dokumentų kontekstas.<br />Norėdami naudoti įrankius, turėsite naudoti @agent komandą.",
       },
     },
     history: {
@@ -834,7 +831,6 @@ const TRANSLATIONS = {
     source_count_other: "{{count}} nuorodos",
     document: "Dokumentas",
     similarity_match: "atitikimas",
-    pause_tts_speech_message: "Sustabdyti garsinį žinutės skaitymą",
     fork: "Atskirti (Fork)",
     delete: "Ištrinti",
     cancel: "Atšaukti",
@@ -866,7 +862,6 @@ const TRANSLATIONS = {
     normal: "Normalus",
     large: "Didelis",
     tools: "Įrankiai",
-    browse: "Naršyti",
     text_size_label: "Teksto dydis",
     select_model: "Pasirinkti modelį",
     slash_commands: "Komandos su „/“",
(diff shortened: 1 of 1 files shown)


frontend/src/pages/GeneralSettings/Connections/TelegramBot/index.jsx
```diff
@@ -0,0 +1,501 @@
...
+      {pendingUsers.length > 0 && (
+        <UsersTable
+          title="Pending Approval"
+          description="Users waiting to be verified. Match the pairing code shown here with the one displayed in their Telegram chat."
+          users={pendingUsers}
+          isPending
+          onApprove={handleApprove}
+          onDeny={handleDeny}
+        />
+      )}
```
timothycarambat: Since you pass in `pendingUsers` as a prop, just check it at the top of the component and return `null` when length is 0

frontend/src/pages/GeneralSettings/Connections/TelegramBot/index.jsx
```diff
@@ -0,0 +1,501 @@
...
+                            onClick={() => onDeny(chatId)}
+                            className="border-none flex items-center justify-center text-xs font-medium text-white/80 light:text-black/80 rounded-lg p-1 hover:bg-white hover:bg-opacity-10 hover:light:bg-red-50 hover:light:text-red-500 hover:text-red-300"
+                          >
+                            Deny
+                          </button>
+                        </>
+                      ) : (
+                        <button
+                          onClick={() => onRevoke(chatId)}
+                          className="border-none flex items-center justify-center text-xs font-medium text-white/80 light:text-black/80 rounded-lg p-1 hover:bg-white hover:light:bg-red-50 hover:bg-opacity-10"
```
timothycarambat: These buttons share a lot of their classes, so maybe lets just make a generic we can hoist in or out of the component to render these just to keep is simple.

Thanks for using border-none - saves me trouble later

frontend/src/pages/GeneralSettings/Connections/TelegramBot/index.jsx
```diff
@@ -0,0 +1,501 @@
...
+              to @BotFather
+            </p>
+            <p>3. Choose a name and username for your bot</p>
+            <p>4. Copy the API token you receive</p>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+}
```
timothycarambat, line 97: This is a super hefty component, i know this is WIP but lets be sure to break this out into small sub-components

server/endpoints/telegram.js
```diff
@@ -0,0 +1,283 @@
...
+    "/telegram/config",
+    [validatedRequest, flexUserRoleValid([ROLES.admin])],
+    async (_request, response) => {
+      try {
+        if (await SystemSettings.isMultiUserMode()) {
+          return response.status(403).json({
+            config: null,
+            error: "Telegram bot is only available in single-user mode.",
+          });
+        }
```
timothycarambat: Dont need to hit the DB again to check this
💡 https://github.com/Mintplex-Labs/anything-llm/blob/master/server/utils/middleware/multiUserProtected.js

-> sets `response.locals?.multiUserMode` (boolean)

server/endpoints/telegram.js
```diff
@@ -0,0 +1,283 @@
...
+          return response.status(500).json({ success: false, error });
+        }
+
+        // Start the bot with the plaintext token
+        const service = new TelegramBotService();
+        await service.start({ ...storedConfig, bot_token });
+
+        await EventLogs.logEvent("telegram_bot_connected", {
+          bot_username: verification.username,
+        });
```
timothycarambat, line 120: Good idea, lets add a Telemetry entry too

server/utils/telegramBot/index.js
```diff
@@ -0,0 +1,248 @@
...
+    );
+    this.#bot.onText(/\/reset/, (msg) =>
+      guard(msg, () => handleReset(ctx, msg.chat.id))
+    );
+    this.#bot.onText(/\/clear/, (msg) =>
+      guard(msg, () => handleClear(ctx, msg.chat.id))
+    );
+    this.#bot.onText(/\/resume/, (msg) =>
+      guard(msg, () => handleResume(ctx, msg.chat.id))
+    );
```
timothycarambat: You can probably genericize this by grabbing the slash commands we support and doing a IF(IN_REGES) and then just have a map of command -> handler

frontend/src/pages/GeneralSettings/Connections/TelegramBot/ConnectedView/index.jsx
```diff
@@ -0,0 +1,282 @@
...
+                  <option key={opt.value} value={opt.value}>
+                    {t(opt.key)}
+                  </option>
+                ))}
+              </select>
+            </div>
+          </div>
+        )}
+        <button
+          onClick={handleDisconnect}
```
angelplusultra: I think this button should display a confirmation modal which which handles the disconnection if confirmed.

frontend/src/locales/en/common.js
```diff
@@ -111,6 +111,8 @@ const TRANSLATIONS = {
     contact: "Contact Support",
     "browser-extension": "Browser Extension",
     "mobile-app": "AnythingLLM Mobile",
+    connections: "Connections",
```
angelplusultra: IMO this should be called "Integrations" as that's usually what this kinda stuff is categorized as. 

https://linear.app/integrations
https://www.notion.com/integrations/all
https://openclaw.ai/integrations
https://github.com/integrations

server/prisma/schema.prisma
```diff
@@ -385,3 +385,12 @@ model workspace_parsed_files {
   @@index([workspaceId])
   @@index([userId])
 }
+
+model external_communication_connectors {
```
angelplusultra, line 389: IMO this table name could be better,  simply `integrations` or 'integration_configs` is more clear. Same with the model wrapper.

server/utils/telegramBot/index.js
```diff
@@ -0,0 +1,365 @@
...
+      bot: this.#bot,
+      config: this.#config,
+      getState: (chatId) => this.#getState(chatId),
+      setState: (chatId, updates) => this.#setState(chatId, updates),
+      log: (text, ...args) => this.#log(text, ...args),
+    };
+  }
+
+  async approveUser(chatId) {
+    await approveUser(this.#bot, chatId, this.#config, this.#pendingPairings);
```
angelplusultra, line 291: Why this indirection? Cant the body of this util just fit into this method, there's no other call sites of the util method `approveUser` but here.

server/utils/EncryptionManager/index.js
```diff
@@ -39,7 +39,8 @@ class EncryptionManager {
       this.log("Self-assigning key & salt for encrypting arbitrary data.");
       process.env[this.#keyENV] = crypto.randomBytes(32).toString("hex");
       process.env[this.#saltENV] = crypto.randomBytes(32).toString("hex");
-      if (process.env.NODE_ENV === "production") dumpENV();
+      // Persist keys in dev so they survive server restarts
+      if (process.env.NODE_ENV !== "production") dumpENV();
```
angelplusultra: Is this correct? This means in production if the server restarts the set keys are lost no? Maybe this shouldn't even be conditional.

server/endpoints/telegram.js
```diff
@@ -0,0 +1,335 @@
...
+            default_workspace: connector.config.default_workspace || null,
+            bot_token_masked:
+              ExternalCommunicationConnector.maskToken(plainToken),
+            voice_response_mode:
+              connector.config.voice_response_mode || "text_only",
+          },
+        });
+      } catch (e) {
+        console.error(e.message, e);
+        response.sendStatus(500).end();
```
angelplusultra: `sendStatus()` already closes the response adding `end()` is redundant and could cause "Cannot set headers after they are sent to the client" errors.

server/utils/telegramBot/utils/index.js
```diff
@@ -0,0 +1,82 @@
...
+ * so we delete messages one by one going backwards.
+ * @param {TelegramBot} bot
+ * @param {number} chatId
+ */
+async function clearTelegramChat(bot, chatId) {
+  const marker = await bot.sendMessage(chatId, "Clearing...");
+  const msgId = marker.message_id;
+
+  // Try to delete the last 100 messages (Telegram limits to 48hr old messages)
+  for (let i = msgId; i > Math.max(msgId - 100, 0); i--) {
```
angelplusultra: Deleting 100 messages one-by-one with individual API calls is slow. See if its possible to batch with `Promise.all()` in chunks (e.g., 10 at a time).

server/utils/telegramBot/utils/media.js
```diff
@@ -0,0 +1,100 @@
...
+}
+
+/**
+ * Convert text to speech and send as an audio message in Telegram.
+ * Silently does nothing if TTS is not configured.
+ * @param {TelegramBot} bot
+ * @param {number} chatId
+ * @param {string} text
+ */
+/**
```
angelplusultra, line 127: There are two JSDoc blocks.  Merge them.

server/utils/telegramBot/commands.js
```diff
@@ -0,0 +1,267 @@
...
+ * @typedef {object} BotContext
+ * @property {TelegramBot} bot
+ * @property {object} config
+ * @property {function} getState - (chatId) => { workspaceSlug, threadSlug }
+ * @property {function} setState - (chatId, updates) => void
+ * @property {function} log - (text, ...args) => void
+ */
+
+/** /start */
+async function handleStart(ctx, chatId) {
```
angelplusultra: The `/start` welcome message is never seen by a new user. Every command goes through the guard function `(telegramBot/index.js:170-174)`, which checks `isVerified()` before running the handler. Since `/start` is the first thing Telegram sends when a user opens a bot, they will always fail the verification check and get the pairing request instead.

  The flow:
  1. User opens bot → Telegram automatically sends `/start`
  2. guard checks `isVerified()` → `false` (new user)
  3. `sendPairingRequest()` runs instead → user sees the pairing code message
  4. Admin approves → user gets "You've been approved! Send a message to start chatting." `(verification.js:90-92)`
  5. The welcome message from `handleStart` is never shown

  The only way a user would ever see it is if they manually type `/start` after being approved, which no one would think to do. And even then, it's just a less useful version of `/status`, it only shows the workspace name, while `/status` shows workspace, thread, provider, and model.

Image

server/utils/telegramBot/utils/verification.js
```diff
@@ -0,0 +1,140 @@
...
+    [
+      "You need to be approved before using this bot.",
+      "",
+      `Your pairing code is: ${code}`,
+      "",
+      "Ask the AnythingLLM admin to go to Settings → Connections → Telegram and approve your request.",
+      "",
+      "Make sure the pairing code shown here matches what is displayed in the settings page. This ensures no one else is trying to connect on your behalf.",
+    ].join("\n")
+  );
```
angelplusultra, line 89: What is the pairing code for? It's never actually verified, the admin just clicks "Approve" without entering or confirming the code anywhere. This seems like unnecessary complexity that could be removed without changing how approval works.

shatfield4: Used to confirm the code displayed to the user because there is a possibility that someone could create a username very similar to the person trying to connect. This adds extra security to ensure the user sees the same code on both the telegram chat and device running AnythingLLM. Openclaw does something similar to this with pairing.

server/utils/telegramBot/commands.js
shatfield4: It's useful to keep here because if a user uses the `/clear` command, we clear all messages from the chat. Then Telegram pops up a button in their UI to `Start` as seen here:

Screenshot 2026-03-19 at 5 27 23 PM


When clicking the `Start` button it automatically sends the `/start` command welcoming the user again.

server/utils/telegramBot/utils/index.js
shatfield4: Looks like this was removed by Tim, resolving this.

timothycarambat:
@shatfield4 This is ready now. We need a docs page on this entire feature and how to use it. You can use the current UI for now for display purposes for screeshots/screen-recordings.

https://github.com/Mintplex-Labs/anythingllm-docs/issues/223

PR #5192, opened 2026-03-11
feat: Move  Native Embedder and Reranker Into Isolated Workers w/ Job Queues | Add Document Embedding Status Events

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #

### Description

Moves the native embedder and reranker from the main process into isolated child processes using `child_process.fork()` with a serial job queue. This prevents OOM crashes from large document batches from taking down the entire server, while adding real-time document and chunk-level progress reporting to the UI via SSE.

**Architecture:**
- `WorkerQueue` class manages a forked child process with serial FIFO job processing, auto-fork on first job, and configurable TTL that kills the worker when idle
- `EmbeddingProgressBus` (singleton EventEmitter) acts as the central hub for progress events between `Document.addDocuments` and SSE endpoint listeners, with event buffering for late-joining clients after page reloads
- Separate queue instances for embedding and reranking workers — each gets its own forked process
- Query embedding (lightweight, single text) still runs in-process; only bulk document embedding is routed through the worker queue

**Progress Reporting:**
- SSE endpoint at `/workspace/:slug/embed-progress` streams document-level events (`batch_starting`, `doc_starting`, `doc_complete`, `doc_failed`, `all_complete`) and chunk-level events (`chunk_progress`)
- `EmbeddingProgressContext` (React Context) manages progress state globally and connects SSE on component mount — server-side event replay catches up on any in-progress jobs without needing client-side persistence
- Progress is scoped per-user — each user only sees their own embedding jobs, even when multiple users embed into the same workspace concurrently
- Progress UI shows in the document management modal with per-file status (Queued → Progress Bar → Complete/Failed) and auto-clears 5 seconds after completion
- When no embedding is in progress, the SSE connection stays open silently and waits — no premature signals are sent, avoiding race conditions between the SSE connection and the embed API call

**Chunk-Level Progress:**
- `embedChunksInProcess` reports progress after each chunk group via an `onProgress` callback
- The embedding worker bridges this to the parent process via IPC (`process.send`)
- `WorkerQueue` receives the IPC messages and forwards them via its `onProgress` callback
- `Document.addDocuments` sets the current document context (`setEmbeddingContext`) before each vectorization call so that chunk progress events can be attributed to the correct document, workspace, and user — this is safe because the embedding queue is serial (one job at a time)
- The frontend renders a progress bar with percentage for each document being embedded

**Worker TTL:**
- Configurable via environment variables (`NATIVE_EMBEDDING_WORKER_TTL`, `NATIVE_RERANKING_WORKER_TTL`) with defaults of 300s and 900s respectively
- TTL controls how long the worker stays alive after finishing work to keep it hot for subsequent jobs
- TTL values are re-read from env before each job so changes take effect without server restart
- Not exposed in the UI — most users won't need to change these since Node worker startup time is negligible

**Other Changes:**
- `NativeEmbedder.embedChunks()` routes through the worker queue — no changes needed in vector DB providers or non-native embedding engines
- `NativeEmbeddingReranker.rerankViaWorker()` added as static method to route through the queue from LanceDB
- `batch_starting` event emitted at the start of a batch with the full file list, so SSE history replay can seed all files as "pending" for late-joining clients
- `doc_failed` event emitted when `fileData()` returns null or when the database write fails (previously files were silently skipped, stuck as "pending" in UI)

### Visuals (if applicable)

https://github.com/user-attachments/assets/01e5d8fd-b27c-4e13-892d-464d6cf6505a

The document management modal now shows real-time embedding progress when documents are being embedded into a workspace. Each file shows a progress bar with percentage as its chunks are embedded, then transitions to Complete or Failed.

### Additional Information

- Worker TTL can be set to 0 for immediate shutdown after work completes
- The reranking worker has a longer default TTL (900s) since it runs on every chat query when accuracy-optimized search is enabled with LanceDB, and frequent cold starts would add overhead
- Event replay on SSE reconnect ensures no stale "Queued" states after page refresh — the server buffers all document-level and chunk-level events and replays them to new subscribers
- Multi-user: progress is scoped per-user via SSE userId filtering. No cross-user visibility or document locking.

### Developer Validations

### DIFF
--- a/collector/__tests__/utils/downloadURIToFile/index.test.js
+++ b/collector/__tests__/utils/downloadURIToFile/index.test.js
@@ -1,0 +1,96 @@
+const path = require("path");
+const { SUPPORTED_FILETYPE_CONVERTERS } = require("../../../utils/constants");
+const { mimeToExtension } = require("../../../utils/downloadURIToFile");
+
+/**
+ * Simulates the filename-building logic from downloadURIToFile
+ * to verify extension inference works correctly.
+ */
+function buildFilenameWithExtension(sluggedFilename, contentType) {
+  const existingExt = path.extname(sluggedFilename).toLowerCase();
+  if (!SUPPORTED_FILETYPE_CONVERTERS.hasOwnProperty(existingExt)) {
+    const mimeType = contentType?.toLowerCase()?.split(";")[0]?.trim();
+    const inferredExt = mimeToExtension(mimeType);
+    if (inferredExt) {
+      return sluggedFilename + inferredExt;
+    }
+  }
+  return sluggedFilename;
+}
+
+describe("mimeToExtension", () => {
+  test("returns null for invalid or unknown input", () => {
+    expect(mimeToExtension(null)).toBeNull();
+    expect(mimeToExtension(undefined)).toBeNull();
+    expect(mimeToExtension("application/octet-stream")).toBeNull();
+  });
+
+  test("returns first extension from ACCEPTED_MIMES for known types", () => {
+    expect(mimeToExtension("application/pdf")).toBe(".pdf");
+  });
+});
+
+describe("buildFilenameWithExtension", () => {
+  test("appends .pdf when URL path has no recognized extension (arxiv case)", () => {
+    // Simulates: https://arxiv.org/pdf/2307.10265
+    // slugify produces something like "arxiv.org-pdf-230710265"
+    const filename = "arxiv.org-pdf-230710265";
+    const result = buildFilenameWithExtension(filename, "application/pdf");
+    expect(result).toBe("arxiv.org-pdf-230710265.pdf");
+  });
+
+  test("appends .pdf when URL has numeric-looking extension", () => {
+    // path.extname("arxiv.org-pdf-2307.10265") => ".10265" which is not in SUPPORTED_FILETYPE_CONVERTERS
+    const filename = "arxiv.org-pdf-2307.10265";
+    const result = buildFilenameWithExtension(
+      filename,
+      "application/pdf; charset=utf-8"
+    );
+    expect(result).toBe("arxiv.org-pdf-2307.10265.pdf");
+  });
+
+  test("does NOT append extension when file already has a supported extension", () => {
+    const filename = "example.com-document.pdf";
+    const result = buildFilenameWithExtension(filename, "application/pdf");
+    expect(result).toBe("example.com-document.pdf");
+  });
+
+  test("does NOT append extension when file has .txt extension", () => {
+    const filename = "example.com-readme.txt";
+    const result = buildFilenameWithExtension(filename, "text/plain");
+    expect(result).toBe("example.com-readme.txt");
+  });
+
+  test("does not append extension for unknown content type", () => {
+    const filename = "example.com-binary-blob";
+    const result = buildFilenameWithExtension(
+      filename,
+      "application/octet-stream"
+    );
+    expect(result).toBe("example.com-binary-blob");
+  });
+
+  test("does not append extension when content type is null", () => {
+    const filename = "example.com-unknown";
+    const result = buildFilenameWithExtension(filename, null);
+    expect(result).toBe("example.com-unknown");
+  });
+
+  test("appends .docx for word document MIME type", () => {
+    const filename = "sharepoint.com-documents-report";
+    const result = buildFilenameWithExtension(
+      filename,
+      "application/vnd.openxmlformats-officedocument.wordprocessingml.document"
+    );
+    expect(result).toBe("sharepoint.com-documents-report.docx");
+  });
+
+  test("handles content type with charset parameter correctly", () => {
+    const filename = "api.example.com-export-data";
+    const result = buildFilenameWithExtension(
+      filename,
+      "text/csv; charset=utf-8"
+    );
+    expect(result).toBe("api.example.com-export-data.csv");
+  });
+});
--- a/collector/processLink/helpers/index.js
+++ b/collector/processLink/helpers/index.js
@@ -6,6 +6,16 @@ const { ACCEPTED_MIMES } = require("../../utils/constants");
 const { validYoutubeVideoUrl } = require("../../utils/url");
 
 /**
+ * Parse a Content-Type header value and return the MIME type without charset or other parameters.
+ * @param {string|null} contentTypeHeader - The raw Content-Type header value
+ * @returns {string|null} - The MIME type (e.g., "application/pdf") or null
+ */
+function parseContentType(contentTypeHeader) {
+  if (!contentTypeHeader) return null;
+  return contentTypeHeader.toLowerCase().split(";")[0].trim() || null;
+}
+
+/**
  * Get the content type of a resource
  * - Sends a HEAD request to the URL and returns the Content-Type header with a 5 second timeout
  * @param {string} url - The URL to get the content type of
@@ -34,8 +44,9 @@ async function getContentTypeFromURL(url) {
         contentType: null,
       };
 
-    const contentType = res.headers.get("Content-Type")?.toLowerCase();
-    const contentTypeWithoutCharset = contentType?.split(";")[0].trim();
+    const contentTypeWithoutCharset = parseContentType(
+      res.headers.get("Content-Type")
+    );
     if (!contentTypeWithoutCharset)
       return {
         success: false,
@@ -171,6 +182,7 @@ async function processAsFile({ uri, saveAsDocument = true }) {
 }
 
 module.exports = {
+  parseContentType,
   returnResult,
   getContentTypeFromURL,
   determineContentType,
--- a/collector/utils/downloadURIToFile/index.js
+++ b/collector/utils/downloadURIToFile/index.js
@@ -1,9 +1,25 @@
-const { WATCH_DIRECTORY } = require("../constants");
+const { WATCH_DIRECTORY, ACCEPTED_MIMES } = require("../constants");
 const fs = require("fs");
 const path = require("path");
 const { pipeline } = require("stream/promises");
 const { validURL } = require("../url");
 const { default: slugify } = require("slugify");
+
+// Add a custom slugify extension for slashing to handle URLs with paths.
+slugify.extend({ "/": "-" });
+
+/**
+ * Maps a MIME type to the preferred file extension using ACCEPTED_MIMES.
+ * Returns null if the MIME type is not recognized or if there are no possible extensions.
+ * @param {string} mimeType - The MIME type to resolve (e.g., "application/pdf")
+ * @returns {string|null} - The file extension (e.g., ".pdf") or null
+ */
+function mimeToExtension(mimeType) {
+  if (!mimeType || !ACCEPTED_MIMES.hasOwnProperty(mimeType)) return null;
+  const possibleExtensions = ACCEPTED_MIMES[mimeType] ?? [];
+  if (possibleExtensions.length === 0) return null;
+  return possibleExtensions[0];
+}
 
 /**
  * Download a file to the hotdir
@@ -33,10 +49,29 @@ async function downloadURIToFile(url, maxTimeout = 10_000) {
       .finally(() => clearTimeout(timeout));
 
     const urlObj = new URL(url);
-    const filename = `${urlObj.hostname}-${slugify(
-      urlObj.pathname.replace(/\//g, "-"),
-      { lower: true }
-    )}`;
+    const sluggedPath = slugify(urlObj.pathname, { lower: true });
+    let filename = `${urlObj.hostname}-${sluggedPath}`;
+
+    const existingExt = path.extname(filename).toLowerCase();
+    const { SUPPORTED_FILETYPE_CONVERTERS } = require("../constants");
+
+    // If the filename does not already have a supported file extension,
+    // try to infer one from the response Content-Type header.
+    // This handles URLs like https://arxiv.org/pdf/2307.10265 where the
+    // path has no explicit extension but the server responds with
+    // Content-Type: application/pdf.
+    if (!SUPPORTED_FILETYPE_CONVERTERS.hasOwnProperty(existingExt)) {
+      const { parseContentType } = require("../../processLink/helpers");
+      const contentType = parseContentType(res.headers.get("Content-Type"));
+      const inferredExt = mimeToExtension(contentType);
+      if (inferredExt) {
+        console.log(
+          `[Collector] URL path has no recognized extension. Inferred ${inferredExt} from Content-Type: ${contentType}`
+        );
+        filename += inferredExt;
+      }
+    }
+
     const localFilePath = path.join(WATCH_DIRECTORY, filename);
     const writeStream = fs.createWriteStream(localFilePath);
     await pipeline(res.body, writeStream);
@@ -51,4 +86,5 @@ async function downloadURIToFile(url, maxTimeout = 10_000) {
 
 module.exports = {
   downloadURIToFile,
+  mimeToExtension,
 };


frontend/src/components/Modals/ManageWorkspace/Documents/WorkspaceDirectory/index.jsx
```diff
@@ -467,4 +493,75 @@ function WorkspaceDocumentTooltips() {
...
 }
 
+function EmbeddingFileRow({ filename, status }) {
+  // Extract a readable display name from the filename path
+  // e.g. "custom-documents/my-doc.json" -> "my-doc"
+  const displayName =
+    filename
+      .split("/")
+      .pop()
+      ?.replace(/\.json$/, "") || filename;
```
timothycarambat: Extract to function outside component

frontend/src/components/Modals/ManageWorkspace/Documents/WorkspaceDirectory/index.jsx
```diff
@@ -467,4 +493,75 @@ function WorkspaceDocumentTooltips() {
...
+      <CheckCircle
+        size={16}
+        className="text-green-400 shrink-0"
+        weight="fill"
+      />
+    ),
+    failed: (
+      <XCircle size={16} className="text-red-400 shrink-0" weight="fill" />
+    ),
+  };
```
timothycarambat: Can be hoisted outside of component

frontend/src/components/Modals/ManageWorkspace/Documents/WorkspaceDirectory/index.jsx
```diff
@@ -467,4 +493,75 @@ function WorkspaceDocumentTooltips() {
...
+
+  const statusLabel = {
+    pending: "Queued",
+    embedding: "Embedding...",
+    complete: "Complete",
+    failed: "Failed",
+  };
```
timothycarambat: Can be outside component, but might want a getter factory so we can later use translations + render the right component as needed

frontend/src/components/Modals/ManageWorkspace/Documents/WorkspaceDirectory/index.jsx
```diff
@@ -467,4 +493,75 @@ function WorkspaceDocumentTooltips() {
...
+              ? "text-red-400"
+              : status.status === "complete"
+                ? "text-green-400"
+                : status.status === "embedding"
+                  ? "text-sky-400"
+                  : "text-white/40"
+          }`}
+        >
+          {statusLabel[status.status] || "Queued"}
+        </p>
```
timothycarambat: The className triple-ternary is a bad code smell - lets simplify this to be more readable. I cannot parse this just from looking at this class name right now 🫢

frontend/src/components/Modals/ManageWorkspace/Documents/index.jsx
```diff
@@ -86,36 +109,49 @@ export default function DocumentSettings({ workspace, systemSettings }) {
   const updateWorkspace = async (e) => {
     e.preventDefault();
     setLoading(true);
-    showToast("Updating workspace...", "info", { autoClose: false });
     setLoadingMessage("This may take a while for large documents");
 
-    const changesToSend = {
-      adds: movedItems.map((item) => `${item.folderName}/${item.name}`),
-    };
+    const filenames = movedItems.map(
```
timothycarambat, line 114: Is this used somewhere else? The code was changed to put this into a const, but ultimately does the same thing old code used to do at this section

frontend/src/components/Modals/ManageWorkspace/Documents/index.jsx
```diff
@@ -86,36 +109,49 @@ export default function DocumentSettings({ workspace, systemSettings }) {
...
+    const embedPromise = Workspace.modifyEmbeddings(
+      workspace.slug,
+      changesToSend
+    );
+    startEmbedding(workspace.slug, filenames);
+
+    embedPromise
+      .then(async (res) => {
+        if (res.message) {
           showToast(`Error: ${res.message}`, "error", { clear: true });
```
timothycarambat: If a bunch of files fail at once will this spam the showToast with a ton of message? We report failures in the UI - no need to show additional errors via toast. Same for succcess, we can just remove them

frontend/src/components/Modals/ManageWorkspace/Documents/index.jsx
```diff
@@ -224,6 +260,7 @@ export default function DocumentSettings({ workspace, systemSettings }) {
         saveChanges={updateWorkspace}
         embeddingCosts={embeddingsCost}
         movedItems={movedItems}
+        embeddingProgress={embeddingProgress}
```
timothycarambat, line 255: Since we are in a context, cant we just modify the context so that anything having to do with embed promises is self-contained?

frontend/src/App.jsx
```diff
@@ -30,7 +31,9 @@ export default function App() {
               <LogoProvider>
                 <PfpProvider>
                   <I18nextProvider i18n={i18n}>
-                    <Outlet />
+                    <EmbeddingProgressProvider>
```
timothycarambat: Does the context need to exist around the _entire app_? We can just wrap it around the document picker UI I would think?

frontend/src/EmbeddingProgressContext.jsx
```diff
@@ -0,0 +1,170 @@
...
+
+    try {
+      const token = window.localStorage.getItem(AUTH_TOKEN);
+      const progressUrl = new URL(
+        `${API_BASE}/workspace/${slug}/embed-progress`
+      );
+      if (token) progressUrl.searchParams.set("token", token);
+
+      const eventSource = new EventSource(progressUrl.toString());
+      eventSourcesRef.current[slug] = eventSource;
```
timothycarambat: You should use the `fetchEventSource` lib here we already have for SSE event streaming for chats. Handles a lot of the edge cases, complexity, timeouts and can pass auth

frontend/src/EmbeddingProgressContext.jsx
```diff
@@ -0,0 +1,170 @@
...
+                  delete next[slug];
+                  return next;
+                });
+                delete cleanupTimeoutsRef.current[slug];
+              }, 5000);
+              break;
+          }
+        } catch {
+          // ignore parse errors
+        }
```
timothycarambat: Should be done either in a per-case handler util functions or one large handler function

server/endpoints/workspaces.js
```diff
@@ -1064,6 +1064,56 @@ function workspaceEndpoints(app) {
...
+  // and inject it into the Authorization header before the middleware runs.
+  app.get(
+    "/workspace/:slug/embed-progress",
+    [
+      (req, _res, next) => {
+        if (!req.headers.authorization && req.query.token) {
+          req.headers.authorization = `Bearer ${req.query.token}`;
+        }
+        next();
+      },
```
timothycarambat: Ad-hoc middlewares should at least be local hoisted functions just to keep readable code contained and not inline - especially for security middleware

server/endpoints/workspaces.js
```diff
@@ -1064,6 +1064,56 @@ function workspaceEndpoints(app) {
...
+        response.setHeader("Content-Type", "text/event-stream");
+        response.setHeader("Access-Control-Allow-Origin", "*");
+        response.setHeader("Connection", "keep-alive");
+        response.flushHeaders();
+
+        const { embeddingProgressBus } = require("../utils/WorkerQueue");
+        const { unsubscribe } = embeddingProgressBus.subscribe(
+          { workspaceSlug: workspace.slug, userId },
+          (event) => {
+            response.write(`data: ${JSON.stringify(event)}\n\n`);
```
timothycarambat: https://github.com/Mintplex-Labs/anything-llm/blob/master/server/utils/helpers/chat/responses.js#L222-L225

server/utils/EmbeddingEngines/native/index.js
```diff
@@ -241,6 +244,14 @@ class NativeEmbedder {
   // While this does take a while, it is zero set up and is 100% free and on-instance.
   // It still may crash depending on other elements at play - so no promises it works under all conditions.
   async embedChunks(textChunks = []) {
```
timothycarambat: This entire function was previously relying on file streaming and such. Workers should make that irrelevant since their is no way to force in-process embedding now with Native due to this PR.

So all the tempfile stuff should be removed

server/utils/EmbeddingRerankers/native/index.js
```diff
@@ -213,6 +213,25 @@ class NativeEmbeddingReranker {
...
+      return instance.rerank(query, documents, options);
+    }
+    // Otherwise, queue it in an isolated worker process
+    const { queueReranking } = require("../../WorkerQueue");
+    return await queueReranking({
+      query,
+      documents,
+      topK: options.topK,
+    });
+  }
```
timothycarambat, line 228: Same here, Is there a way to embed now outside of workers?

server/utils/vectorDbProviders/lance/index.js
```diff
@@ -126,8 +125,9 @@ class LanceDb extends VectorDatabase {
       .limit(searchLimit)
       .toArray();
 
-    await reranker
-      .rerank(query, vectorSearchResults, { topK: topN })
+    await NativeEmbeddingReranker.rerankViaWorker(query, vectorSearchResults, {
+      topK: topN,
+    })
```
timothycarambat, line 130: 👍

server/utils/WorkerQueue/index.js
```diff
@@ -0,0 +1,352 @@
...
+    ) * 1000;
+  const { result } = await rerankingQueue.enqueue({ payload });
+  return result.reranked;
+}
+
+module.exports = {
+  queueEmbedding,
+  queueReranking,
+  embeddingProgressBus,
+};
```
timothycarambat, line 81: For this whole file, lets find a better way to break this up - this file is pretty messy and lots of different things going on in the same file.

server/utils/EmbeddingEngines/native/index.js
timothycarambat: I actually see why this is here now since you do in-proc embedding when in the worker but use the same method.

Break in process embedding and then the queue worker into two different methods. Right now, this behavior is based on `typeof process.send` which is something we should definitely be doing explictly and not implying on method call.

frontend/src/components/Modals/ManageWorkspace/Documents/WorkspaceDirectory/index.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/eafa12caad6ef246b37d5bcbcf2feb9ec40d448e

frontend/src/components/Modals/ManageWorkspace/Documents/WorkspaceDirectory/index.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/eafa12caad6ef246b37d5bcbcf2feb9ec40d448e

frontend/src/components/Modals/ManageWorkspace/Documents/WorkspaceDirectory/index.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/eafa12caad6ef246b37d5bcbcf2feb9ec40d448e

frontend/src/components/Modals/ManageWorkspace/Documents/WorkspaceDirectory/index.jsx
angelplusultra: this better? https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/eafa12caad6ef246b37d5bcbcf2feb9ec40d448e

angelplusultra:
> * Timeout ≠ TTL. This looks like we are talking about how long we should keep workers alive after doing some work just to keep the worker hot. This is a TTL, so lets rename that.
> 
>     * Lets remove the UI components to specify TTL for now. Most people will not ever want to touch these nor want to even change it. Node Worker time to start is small enough to shoulder here.
> 
>     * Lets also then remove the associated `systemSettings` key entires since we wont be sending them to the UI. We should keep the `protectedKeys` in dumpENV just in case people DO want to set them manually.
> 
> 
> Clarifying questions:
> 
> How are embedding jobs user segmented? If user A is embedding 10 docs and user B is embedding 1 - when A is queued and B's job finishes and gets `all_complete` doesnt A get that event back at the same time or is this simply based on the job ref in their renderer processes

 Each user's "Save" triggers its own `addDocuments` call, which runs its own loop and emits its own `batch_starting` → `doc_starting` → `doc_complete` → `all_complete` lifecycle. User B's `all_complete` only signals the end of User B's batch — it doesn't touch User A's. On the SSE side, every event is tagged with `userId`, and the subscriber filters on it. So User A's SSE connection only receives events where `event.userId` matches their own. User B's `all_complete` is simply never delivered to User A. The two batches are independent event streams that happen to flow through the same bus. They don't interfere with each other.
 
 You can see this in action in the `EmbeddingProgressBus`:
 
 ```js
   /**
   * Register an SSE listener filtered by workspace and user.
   * Replays any buffered events for the workspace before subscribing to live events.
   * @param {{ workspaceSlug: string, userId?: number }} filter
   * @param {function} callback - receives the progress event payload
   * @returns {{ unsubscribe: function }}
   */
  subscribe(filter, callback) {
    // Replay buffered events so reconnecting clients catch up.
    if (filter.workspaceSlug && this.#history.has(filter.workspaceSlug)) {
      for (const event of this.#history.get(filter.workspaceSlug)) {
        if (filter.userId && event.userId && event.userId !== filter.userId)
          continue;
        callback(event);
      }
    }

    const handler = (event) => {
      if (filter.workspaceSlug && event.workspaceSlug !== filter.workspaceSlug)
        return;
      if (filter.userId && event.userId && event.userId !== filter.userId)
        return;
      callback(event);
    };
    this.on("progress", handler);
    return {
      unsubscribe: () => this.off("progress", handler),
    };
  }

```

File: frontend/src/components/Modals/ManageWorkspace/Documents/index.jsx (hunk 5 above)
Comment (angelplusultra) [reply frontend/src/components/Modals/ManageWorkspace/Documents/index.jsx:114]: Yes Line 130

```js
startEmbedding(workspace.slug, filenames);

```

frontend/src/App.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/f335e2ba760d7104dd1bcb14ebcf276333bb8443

frontend/src/EmbeddingProgressContext.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/2b7bce8ef6c1e1808a7091bc6305d39f1e26e706

server/endpoints/workspaces.js
angelplusultra: Removed it since refactoring to use `fetchEventSource` https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/2b7bce8ef6c1e1808a7091bc6305d39f1e26e706

frontend/src/EmbeddingProgressContext.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/4ed2304900a0927c960ace96fe73078e1ddbc371

server/endpoints/workspaces.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/904ed163ccecd0d26d6b6fc9116528b2a9836957

server/utils/EmbeddingEngines/native/index.js
angelplusultra: is this better? https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/b1d2dd6639f7a43c1336aa2e120d45b6ac36fbbc

server/utils/EmbeddingRerankers/native/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/b1d2dd6639f7a43c1336aa2e120d45b6ac36fbbc

frontend/src/App.jsx
timothycarambat: You could probably even go one level deeper around https://github.com/Mintplex-Labs/anything-llm/commit/f335e2ba760d7104dd1bcb14ebcf276333bb8443#diff-dc25dfd5a8fa0ccbfe8c8fdc01786dfe0811d7e39d9e02ee6bfde5500c280a6fR107-R110 since that is the only component that uses this? I suppose though that could break things if it unmounts by switch tabs - so maybe we keep it here if that is the case.

server/utils/WorkerQueue/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/9a63f98d8baaf0373512dd5c30b71b45b7068c69

frontend/src/components/Modals/ManageWorkspace/Documents/index.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/cf8c6817b51eb1cdb38c05fec95c9075c1652461

server/utils/WorkerQueue/index.js
```diff
@@ -0,0 +1,102 @@
...
+// Solution: Document.addDocuments calls setEmbeddingContext() with the current
+// document's { workspaceSlug, filename, userId } before starting vectorization,
+// and clears it (null) after. When the worker sends chunk progress back via IPC,
+// the onProgress callback below reads this context to emit a fully-attributed
+// "chunk_progress" event on the EmbeddingProgressBus.
+//
+// This is safe because the embedding queue processes jobs serially — only one
+// document is ever being embedded at a time, so the context always matches
+// the active job.
+// ---------------------------------------------------------------------------
```
shatfield4: I see this comment here about the `setEmbeddingContext`. Is there anything stopping us from putting the context inside the payload of the job? I think if we did that then we can remove all the context code we have here and just read directly from the job payload without having to maintain a global state. Having the global state for `setEmbeddingContext` just seems fragile.

server/utils/WorkerQueue/WorkerQueue.js
```diff
@@ -0,0 +1,250 @@
...
+  killWorker() {
+    this.#clearTTLTimer();
+    if (!this.#worker) return;
+
+    try {
+      if (this.#worker.connected) this.#worker.send({ type: "shutdown" });
+    } catch {
+      // Worker may already be disconnected
+    }
+
```
shatfield4, line 72: Let's make sure to call `killWorker` with a shutdown hook in order to prevent orphaned child processes on server shutdown so we can have a graceful shutdown.

Something like this:
```
process.on("SIGTERM", worker.killWorker())
```

server/utils/WorkerQueue/index.js
angelplusultra: It was just a workaround for having to modify like 10 call sites across `addDocumentsToNamespace` and `embedChunks()` to consume the context and only one of those call sites actually uses the context (the native embedder).

Here's passing it directly down the chain. Lemme know if you prefer it: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/77028b2b13c0a1cad3a56c2e3de676aba632c0d5

server/utils/WorkerQueue/WorkerQueue.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5192/commits/d219e03574578e97e648b7f740c9339f4ee3ad01

PR #5228, opened 2026-03-18
feat: Add File Search agent skill

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #

### Description

Adds a new **File Search** agent skill that allows the agent to search and read files from a mounted file storage directory (`/anythingllm-files` in Docker, `<repo>/anythingllm-files` in development).

**How it works:**
- The agent receives `search_terms`, optional `file_types`, and `max_results` parameters from user queries
- Files are discovered and searched using [ripgrep](https://github.com/BurntSushi/ripgrep) (`@vscode/ripgrep`) for fast filename listing and content searching
- Results are ranked using a multi-strategy scoring system: filename match (+2), path match (+1), content match (+3), and a recency bonus for files modified within the last 7 days (+1)
- File content is extracted through the existing collector document parsing pipeline, supporting all formats the collector already handles (PDF, DOCX, XLSX, etc.)
- Token-aware output budgets context per file and will truncate or summarize if the total exceeds the model's context limit
- A wildcard `*` search term returns all files sorted by recency

**How `@vscode/ripgrep` works:**
- The package has a postinstall hook that downloads a platform-specific prebuilt ripgrep binary
- The binary is stored at `server/node_modules/@vscode/ripgrep/bin/rg`
- The package exports `rgPath` which points to this binary — no app-level ripgrep installation required

**Security:**
- Path traversal protection via `isWithin()` validation on all resolved paths
- Search is sandboxed to the configured `FILE_SEARCH_PATH` only

**Changes:**
- New plugin: `server/utils/agents/aibitat/plugins/file-search.js`
- Registered in plugin index with slug alias
- Frontend skill toggle added to Admin > Agents settings
- Dockerfile updated to create and own `/anythingllm-files` directory in both build stages
- Added `@vscode/ripgrep` dependency

Translations PR: #5229 

### Additional Information

- File content extraction reuses the collector's `/parse` endpoint, so parsed files land in `direct-uploads/` and are cleaned up by the existing orphan job
- The skill reports citations for each file whose content is extracted

### Developer Validations


server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,582 @@
...
+           * @param {string[]} fileTypes
+           * @returns {Promise<{fullPath: string, relativePath: string, name: string, size: number, mtime: Date}[]>}
+           */
+          listCandidateFiles: async function (searchRoot, fileTypes = []) {
+            const args = [
+              "--files",
+              "--null",
+              "--no-ignore",
+              // "--max-depth",
+              // `${MAX_SEARCH_DEPTH}`,
```
shatfield4: Remove here too.

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,582 @@
...
+
+const execFileAsync = promisify(execFile);
+
+// In development files live under server/storage; in production they're
+// mounted at /anythingllm-files via Docker volume.
+const FILE_SEARCH_PATH =
+  process.env.NODE_ENV === "development"
+    ? path.resolve(__dirname, "../../../../storage/anythingllm-files")
+    : "/anythingllm-files";
+// const MAX_SEARCH_DEPTH = 5;
```
shatfield4: Remove these comments if `MAX_SEARCH_DEPTH` is not used.

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,582 @@
...
+              .filter(Boolean);
+            if (terms.length === 0) return new Set();
+
+            const args = [
+              "-l",
+              "--null",
+              "--no-ignore",
+              "--no-binary",
+              // "--max-depth",
+              // `${MAX_SEARCH_DEPTH}`,
```
shatfield4: Here too.

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,582 @@
...
+            const files = await this.listCandidateFiles(searchRoot, fileTypes);
+
+            if (files.length > 0) return files;
+
+            // When filtering by type returned nothing, check if the directory
+            // itself is empty vs just having no matches for the requested types.
+            if (fileTypes.length > 0) {
+              const allFiles = await this.listCandidateFiles(searchRoot);
+              if (allFiles.length > 0) return files; // directory has files, just none matching the filter
+            }
```
shatfield4, line 204: We can just remove this section of logic because lines 195-197 already do the same logic here and this section will not change what is returned by the `getCandidates` function.

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,582 @@
...
+const { isWithin, hotdirPath } = require("../../../files");
+
+const execFileAsync = promisify(execFile);
+
+// In development files live under server/storage; in production they're
+// mounted at /anythingllm-files via Docker volume.
+const FILE_SEARCH_PATH =
+  process.env.NODE_ENV === "development"
+    ? path.resolve(__dirname, "../../../../storage/anythingllm-files")
+    : "/anythingllm-files";
```
shatfield4, line 19: Let's use the `STORAGE_DIR` env for this to ensure it works in production if users have specified a different path than our default. See other implementations around our backend how we switch back and forth from dev and prod environments and be sure to match the same pattern here.

Should look like this:

```
const FILE_SEARCH_PATH =                                                                                                       
    process.env.NODE_ENV === "development"                                                                                     
      ? path.resolve(__dirname, "../../../../storage/anythingllm-files")                                                         
      : path.resolve(process.env.STORAGE_DIR, "anythingllm-files");
```

shatfield4 (changes requested):
A few minor changes.

server/utils/agents/aibitat/plugins/file-search.js
angelplusultra: This block distinguishes between an empty directory and a type filter with no matches. When `fileTypes` returns nothing but `allFiles` finds files, we skip the "directory is empty" introspect message, because the directory isn't empty, the filter just didn't match anything.

Line 195 invokes `listCandidateFiles` with a filter 

```js
const files = await this.listCandidateFiles(searchRoot, fileTypes);
```


Line 202 invokes it without a filter 
```js
const allFiles = await this.listCandidateFiles(searchRoot);
```

timothycarambat:
@angelplusultra Needs a docs PR in https://github.com/Mintplex-Labs/anythingllm-docs
https://github.com/Mintplex-Labs/anythingllm-docs/issues/222?reload=1

Can go in the https://docs.anythingllm.com/agent/usage as a new sub-header item

Should also break out Docker with current instructions to use (using volume binding) - for Desktop sub-header just put `TBD`

docker/Dockerfile
```diff
@@ -43,7 +43,8 @@ RUN (getent passwd "$ARG_UID" && userdel -f "$(getent passwd "$ARG_UID" | cut -d
     (getent group "$ARG_GID" && groupdel "$(getent group "$ARG_GID" | cut -d: -f1)") || true && \
     groupadd -g "$ARG_GID" anythingllm && \
     useradd -l -u "$ARG_UID" -m -d /app -s /bin/bash -g anythingllm anythingllm && \
-    mkdir -p /app/frontend/ /app/server/ /app/collector/ && chown -R anythingllm:anythingllm /app
+    mkdir -p /app/frontend/ /app/server/ /app/collector/ && chown -R anythingllm:anythingllm /app && \
+    mkdir -p /anythingllm-files && chown anythingllm:anythingllm /anythingllm-files
```
timothycarambat, line 47: Instead of doing this in the image - which wont carry to the container anyway - lets just create the folder as needed on query when the tool is being used.

Part of the instruction to use this tool will be that the directory exist in the host container for example:
```bash
export STORAGE_LOCATION=$HOME/anythingllm && \
mkdir -p $STORAGE_LOCATION && \
touch "$STORAGE_LOCATION/.env" && \
docker run -d -p 3001:3001 \
--cap-add SYS_ADMIN \
-v /path/for/me:/app/server/storage/anythingllm-files/my-first-folder \ 
-e STORAGE_DIR="/app/server/storage" \
mintplexlabs/anythingllm
```

Will create the folder in host and it will bind to that location in docker anyway even though it doesnt even exist, but the mapping is already live.

timothycarambat: So you can remove this.

docker/Dockerfile
```diff
@@ -112,7 +113,8 @@ RUN (getent passwd "$ARG_UID" && userdel -f "$(getent passwd "$ARG_UID" | cut -d
...
```
timothycarambat, line 117: This too

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,574 @@
...
+              call: JSON.stringify({
+                search_terms: ["README"],
+              }),
+            },
+            {
+              prompt: "What documents do I have?",
+              call: JSON.stringify({
+                search_terms: ["*"],
+              }),
+            },
```
timothycarambat, line 65: Lets keep it to 3 distinct exampls

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,574 @@
...
+            } catch (error) {
+              // Exit code 1 means "no matches" — not an error for our purposes
+              if (error.code === 1) return error.stdout || "";
+              throw new Error(
+                error.stderr?.trim() ||
+                  error.message ||
+                  "Ripgrep search failed."
+              );
+            }
+          },
```
timothycarambat, line 319: We should be using an interface for this so we get typehinting and easily can ensure return of objects and such
https://www.npmjs.com/package/ripgrep-node

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,574 @@
...
+
+          /**
+           * Extract content from ranked files and normalize the response payload.
+           * @param {{fullPath: string, relativePath: string, name: string, size: number, mtime: Date}[]} files
+           * @returns {Promise<{name: string, path: string, size: number, modified: string, content: string}[]>}
+           */
+          extractResults: async function (files) {
+            const results = [];
+            for (const file of files) {
```
timothycarambat, line 219: Would make more sense here to parallelize this to get concurrency since parsing files is very trivial

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,574 @@
...
+                  "Abort was triggered, exiting summarization early."
+                );
+                this.controller.abort();
+              });
+              return summarizeContent({
+                provider: this.super.provider,
+                model: this.super.model,
+                controllerSignal: this.controller.signal,
+                content: output,
+              });
```
timothycarambat, line 290: We should be careful with this. Super large docs do **not** summarize well. This LC wrapper for summarization absolutely sucks and tbh we should avoid it at all costs. 

Instead, just truncate content if it it too long.

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,574 @@
...
+                .map((p) => p.trim())
+                .filter(Boolean)
+                .map((rel) => path.resolve(searchRoot, rel))
+            );
+          },
+
+          /**
+           * Run all search strategies (filename, path, content) and rank by relevance.
+           * Scoring weights: filename match +2, path match +1, content match +3, recent (<7d) +1.
+             Files must match at least one strategy to appear in results.
```
timothycarambat, line 422: good idea

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,574 @@
...
+              if (!content) return null;
+
+              // Report citation for this file
+              this.super.addCitation?.({
+                id: file.relativePath,
+                title: file.name,
+                text: content,
+                chunkSource: `file://${file.relativePath}`,
+                score: null,
+              });
```
timothycarambat, line 538: We only need to report it as a citation if it is in the final set and actually read, it could be pruned out of the top 10, no?

server/utils/agents/aibitat/plugins/file-search.js
```diff
@@ -0,0 +1,574 @@
...
+          /**
+           * Format a file size in bytes to a human-readable string.
+           * @param {number} bytes
+           * @returns {string}
+           */
+          formatSize: function (bytes) {
+            if (bytes < 1024) return `${bytes} B`;
+            if (bytes < 1024 * 1024) return `${(bytes / 1024).toFixed(1)} KB`;
+            return `${(bytes / (1024 * 1024)).toFixed(1)} MB`;
+          },
```
timothycarambat, line 565: humanFileSize() is a function we have that can probably replace this

PR #5241, opened 2026-03-20
ffeat: add SSE support for ordinary conversation API endpoints

### LINKED ISSUE #5240: [FEAT]: Provide SSE support for ordinary conversation requests

Currently, when @agent is not used, ordinary conversation requests seem to be just a regular POST request, merely waiting for a one-time response content. However, most current models support the thinking mode. The thinking stage may last for a long time, but the conversation box merely waits without showing the thinking content. Moreover, after a long period of thinking, an error prompt may be displayed because the POST request has timed out and the reply data has not been received.

nehaaprasad:
### Pull Request Type

[x] ✨ feat

### Relevant Issues

resolves #5240

### Description

- Add optional stream: true to ordinary API chat routes.
- Stream responses via SSE for workspace and thread chat endpoints.
- Keep default non-stream JSON behavior unchanged.

### Visuals (if applicable)
N/A

### DIFF
--- a/server/endpoints/api/workspaceThread/index.js
+++ b/server/endpoints/api/workspaceThread/index.js
@@ -389,27 +389,42 @@ function apiWorkspaceThreadEndpoints(app) {
           userId,
           attachments = [],
           reset = false,
+          stream = false,
         } = reqBody(request);
         const workspace = await Workspace.get({ slug });
-        const thread = await WorkspaceThread.get({
-          slug: threadSlug,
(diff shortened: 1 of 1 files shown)


timothycarambat:
I appreciate the effort - 

There is a very good reason for this arch and this change would break several things and installations globally - this should be restricted to a core team change as it will incur a ton of side effects a non-core member would not be aware of!

PR #5254, opened 2026-03-23
feat: Document Embedding Status Events | Refactor Document Embedding to Job Queue and Forked Process

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Description

Moves native embedding out of the main server process and into an isolated child process with a FIFO job queue, preventing OOM crashes from large document batches. Adds real-time per-file embedding progress in the UI via SSE.

#### Isolated embedding worker (`server/jobs/embedding-worker.js`)
- Embedding now runs in a forked child process spawned through `BackgroundService/Bree` — the same infrastructure used for all other background jobs (cleanup, sync).
- The worker stays alive between jobs to keep the ML model loaded in memory. After a configurable TTL (default 5 min, via `NATIVE_EMBEDDING_WORKER_TTL`), the worker is killed to free resources.
- If the worker OOMs on a large document, only the child process dies — the main server stays up.

#### WorkerQueue (`server/utils/WorkerQueue/`)
- `WorkerQueue.js` — generic serial FIFO job queue. Manages the worker lifecycle: spawns via Bree, waits for a ready handshake, sends jobs one at a time via IPC, handles results/errors, and manages the TTL idle timer.
- `EmbeddingProgressBus.js` — singleton event emitter that buffers progress events per-workspace so late-joining SSE clients can replay history.
- `index.js` — creates the embedding queue instance, wires the chunk progress callback to the bus, and registers graceful shutdown handlers.

#### NativeEmbedder changes (`server/utils/EmbeddingEngines/native/index.js`)
- `embedChunks()` now routes through the worker queue instead of running in-process.
- The original in-process logic is renamed to `embedChunksInProcess()` — only called by the worker.
- `embedTextInput()` runs directly in-process (single query embeddings don't need isolation).
- `embedChunksInProcess()` accepts an `onProgress` callback for chunk-level progress reporting.

#### Document model progress events (`server/models/documents.js`)
- `addDocuments()` emits progress events on the bus: `batch_starting`, `doc_starting`, `doc_complete`, `doc_failed`, `chunk_progress`, and `all_complete`.

#### SSE endpoint (`server/endpoints/workspaces.js`)
- New `GET /workspace/:slug/embed-progress` endpoint. Subscribes to the progress bus filtered by workspace and user, replays buffered history for reconnecting clients, and streams live events.

#### Vector DB providers
- All 8 providers (Astra, Chroma, Lance, Milvus, PGVector, Pinecone, Qdrant, Weaviate) + the base class pass an `embeddingContext` parameter through `addDocumentToNamespace()` → `embedChunks()` so the worker queue can attach document context to progress events.

#### Frontend progress UI
- `EmbeddingProgressContext` — React context that manages SSE connections via `@microsoft/fetch-event-source`. Tracks per-workspace, per-file status (`pending` → `embedding` → `complete`/`failed`) with chunk-level progress percentages. Auto-clears 5s after completion.
- `WorkspaceDirectory` — replaces the generic loading spinner with per-file progress rows showing status icons and progress bars during embedding.
- `DocumentSettings` — connects SSE on mount (catches in-progress jobs via history replay), triggers `startEmbedding()` before the API call, and auto-refreshes file lists when embedding completes.

### Additional Information

- Worker TTL is configurable via `NATIVE_EMBEDDING_WORKER_TTL` env var (seconds, default 300).
- This PR only covers embedding. Reranking is unchanged from master.

### Developer Validations


server/jobs/embedding-worker.js
```diff
@@ -0,0 +1,78 @@
...
+      // This callback converts in-process progress into IPC messages that the
+      // parent's WorkerQueue receives and forwards to the EmbeddingProgressBus.
+      const result = await embedder.embedChunksInProcess(
+        textChunks,
+        (progress) => {
+          process.send({
+            type: "progress",
+            jobId: msg.jobId,
+            chunksProcessed: progress.chunksProcessed,
+            totalChunks: progress.totalChunks,
```
shatfield4: Let's add a `message:` to all the `process.send()` calls in the `embedding-worker.js`.

The server logs get spammed currently with hard to read logs since our `onWorkerMessageHandler` inside `server/utils/BackgroundWorkers/index.js` logs on every worker event that is received. It logs out the `.message` property on the job event which does not exist on all the events sent by `embedding-worker.js`.

server/jobs/embedding-worker.js
```diff
@@ -0,0 +1,78 @@
...
+          });
+        }
+      );
+
+      process.send({
+        type: "result",
+        jobId: msg.jobId,
+        result: { vectors: result },
```
shatfield4: Add message field here too.

server/jobs/embedding-worker.js
```diff
@@ -0,0 +1,78 @@
...
+      });
+    } catch (err) {
+      process.send({
+        type: "error",
+        jobId: msg.jobId,
+        error: err.message || String(err),
+      });
```
shatfield4: Add message field.

server/jobs/embedding-worker.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  if (msg.type === "shutdown") {
+    process.exit(0);
+  }
+});
+
+// Signal that the worker is ready
+process.send({ type: "ready" });
```
shatfield4: Add message field.

frontend/src/EmbeddingProgressContext.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+const EmbeddingProgressContext = createContext();
+
+export function useEmbeddingProgress() {
+  const ctx = useContext(EmbeddingProgressContext);
+  if (!ctx)
+    throw new Error(
+      "useEmbeddingProgress must be used within EmbeddingProgressProvider"
+    );
+  return ctx;
```
shatfield4: Change this to just

```
return useContext(EmbeddingProgressContext);
```

This will cause the frontend to crash if we ever run into an instance where the context is undefined and we would rather this pass through instead of crashing the entire frontend. All other contexts do not throw errors in the hook function.

server/utils/WorkerQueue/index.js
```diff
@@ -0,0 +1,60 @@
...
+  const { result } = await embeddingQueue.enqueue({ payload, context });
+  return result.vectors;
+}
+
+// Graceful shutdown: kill forked workers so they don't become orphaned.
+for (const signal of ["SIGTERM", "SIGINT"]) {
+  process.on(signal, () => {
+    embeddingQueue.killWorker();
+  });
+}
```
shatfield4: We can remove this now since bree uses @ladjs/graceful under the hood and automatically handles cleaning up processes gracefully.

server/utils/WorkerQueue/WorkerQueue.js
```diff
@@ -0,0 +1,292 @@
...
+    await bg.bree.add({
+      name: this.#breeJobId,
+      path: this.workerScript,
+    });
+
+    await bg.bree.run(this.#breeJobId);
+    this.#worker = bg.bree.workers.get(this.#breeJobId);
+
+    if (!this.#worker) {
+      throw new Error("Failed to get worker reference from Bree");
```
shatfield4: Can we reuse the logic from `BackgroundService.runJob()` instead of handling all the bree calls in here? The `runJob` method basically does the same thing just without TTL, so you can add the TTL logic as an option to `runJob` and have the BackgroundService handle everything instead to keep this simpler and more maintainable/reusable.

server/utils/WorkerQueue/WorkerQueue.js
```diff
@@ -0,0 +1,292 @@
...
+        worker.kill("SIGKILL");
+      } catch {
+        /* already dead */
+      }
+    }, 5_000);
+
+    this.#cleanupBreeJob(breeJobId);
+  }
+
+  // -- internal ---------------------------------------------------------------
```
shatfield4: Remove comment, # prefixes are self explanatory.

server/utils/EmbeddingEngines/native/index.js
```diff
@@ -223,27 +223,65 @@ class NativeEmbedder {
    */
   async embedTextInput(textInput) {
     textInput = this.#applyQueryPrefix(textInput);
-    const result = await this.embedChunks(
+    const result = await this.embedChunksInProcess(
```
shatfield4: Is there a reason for not just using embedChunks? Switching to embedChunksInProcess goes around the Bree worker and does this in the main thread when we can use the bree worker version. I'm not sure if I'm missing anything here but it seems like we can use the worker version in case someone decides to send a huge piece of text as context which could cause OOM.

server/utils/WorkerQueue/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Queue an embedding job for the native embedder worker.
+ * @param {{ textChunks: string[] }} payload
+ * @param {{ workspaceSlug: string, filename: string, userId: number|null }|null} context - Document context for progress reporting
+ * @returns {Promise<Array<number[]>>} The embedding vectors
+ */
+async function queueEmbedding(payload, context = null) {
+  embeddingQueue.ttl =
+    envTTLSec("NATIVE_EMBEDDING_WORKER_TTL", DEFAULT_EMBEDDING_TTL_SEC) * 1000;
```
shatfield4: We can remove this since it is already set by the constructor on startup to prevent rereading the env and  reassigning it every time we queue an embedding. If it is meant to be adjustable at runtime though then we can leave it, I just think there is no benefit to adjusting TTL timings at runtime.

shatfield4 (changes requested):
A few refactor changes.

server/jobs/embedding-worker.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/b4166c31ba725e39025c77538475d007e0984dfe

server/jobs/embedding-worker.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/b4166c31ba725e39025c77538475d007e0984dfe

server/jobs/embedding-worker.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/b4166c31ba725e39025c77538475d007e0984dfe

server/jobs/embedding-worker.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/b4166c31ba725e39025c77538475d007e0984dfe

frontend/src/EmbeddingProgressContext.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/395cb50b33571081f16d0f18c94312b3958d0a41

server/utils/WorkerQueue/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/518019d7aeabc4cd01de65b5eaf10a69a95c2436

server/utils/WorkerQueue/WorkerQueue.js
angelplusultra: I see what you mean about the duplication. I extracted the shared Bree boilerplate into two helpers on `BackgroundService`. `spawnWorker(scriptPath)` and `removeJob(jobId)` and refactored both `runJob` and `WorkerQueue` to use them. So all Bree interaction is centralized now.

As for making runJob itself work for `WorkerQueue`'s use cas, the gap is bigger than just TTL. `runJob` is fire and-forget (spawn, send, resolve on exit), but `WorkerQueue` needs to keep the worker alive across many jobs so the ML model stays loaded, plus a ready handshake, serial job queue, and structured IPC routing. Adding all of that to runJob would essentially mean two completely different code paths in one method.

The shared helpers should address the maintainability concern though. `WorkerQueue` no longer touches Bree directly.
  
https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/9b8a7f1770e5b4a3d685862859a2b9afaf9f6d23

server/utils/WorkerQueue/WorkerQueue.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/8ce9ea51522630c08e93709f050ba059b7ecdf15

server/utils/EmbeddingEngines/native/index.js
angelplusultra: According to the PRD for the feature, it states for prompt embedding to remain in the main process.

https://www.notion.so/mintplex-labs/Threaded-Embedder-31ed47b4f77480acb5ebef3237b1502d?source=copy_link#31ed47b4f77480e595b9c69d55495b4c

server/utils/WorkerQueue/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5254/commits/bbe1a06c1813158140742f35e2400492bea028d4

server/utils/WorkerQueue/WorkerQueue.js
shatfield4: Got it, this is better and cleaner so I'll resolve this.

angelplusultra:
> Check `embedding-worker.js` there are still `process.send()` calls that do not have the `message` property.
> 
> Make sure every time we call `process.send()` we have the message property for a clean log output in the server console.
> 
> Other than that, LGTM.

Were they not all handled in https://github.com/Mintplex-Labs/anything-llm/commit/b4166c31ba725e39025c77538475d007e0984dfe? I noticed you didn't click resolve for any of the `process.send()` fixes.

shatfield4 (approved):
LGTM

PR #5269, opened 2026-03-26
feat: Memories/Personalization

### LINKED ISSUE #5276: [FEAT]: Memories/Personalization

- Implement the memories/personalization feature per internal spec
- Uses bg worker to processes chats and extract memories to be used in future conversations
- Reranks memories based on query to provide useful context to LLM model
- Support for regular chats, agent mode

### LINKED ISSUE #5389: [FEAT]: Implement new memories UI

- Implement new memories feature UI per internal designs

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5276 
resolves #5389

Translations PR: https://github.com/Mintplex-Labs/anything-llm/pull/5661
Docs PR: https://github.com/Mintplex-Labs/anythingllm-docs/pull/235

### Description

- Added a new memories database table with Prisma schema and migration for storing user memories scoped to global or per-workspace                                                                           
- Added a Memory server model with full CRUD, promote-to-global, replace workspace memories, and multi-user migration support   
- Added memoryProcessed flag to workspace_chats table to track which chats have been processed for memory extraction            
- Added server API endpoints for memories (list, create, update, delete, promote to global, clear all, run extraction)          
- Added a memoryEnabled middleware guard on all memory endpoints so they 403 when the feature is off                            
- Added ownership checks on memory update/delete/promote so users can only modify their own memories                            
- Added a background worker job (extract-memories) that runs every 15 minutes, uses the workspace LLM to extract memories from  unprocessed chats                                                                                                               
- Background extraction skips running if the user has been active in the last 20 minutes to avoid processing mid-conversation   
- Added memory injection into chat system prompts (streaming, sync, and API chat handlers) via a shared promptWithMemories util 
- Memories are reranked against the current prompt + recent history using the native embedding reranker when more than 5 workspace memories exist                                                                                                        
- Added global memory limit of 5 and workspace memory limit of 20, enforced on create and promote                               
- Added admin Personalization settings page with toggle to enable/disable the feature, run extraction manually, clear all memories, and manage global memories                                                                                            
- Added workspace-level Personalization settings tab where users can view, add, edit, delete, and promote workspace memories to global                                                                                                                          
- Added Personalization link to the admin settings sidebar                                                                    
- Added Personalization tab to workspace settings, visible only when the feature is enabled                                     
- Default role users can now access workspace settings (only the Personalization tab) when memory is enabled, with proper redirects for all other tabs                                                                                                    
- Added frontend Memory API model for all memory endpoints                                                                      
- Added shared MemoryForm and MemoryItem components used across both admin and workspace personalization pages                  
- Added memory_enabled to system settings supported/public fields and exposed MemoryEnabled in the system keys response         
- Added single-user to multi-user migration for both memories and workspace chats (assigns orphaned records to the admin user)

### Visuals (if applicable)

Screenshot 2026-04-10 at 4 56 50 PM

Screenshot 2026-04-10 at 4 56 26 PM

Screenshot 2026-04-10 at 4 56 36 PM

### Additional Information

- Test CRUD for all areas where we can add/promote to global memories
- Throughly test RBAC ensuring memories are paired to user accounts in multi-user mode
- Ensure pending memories are migrated to the new admin account created when enabling multi-user mode and there are no orphaned memories

### Developer Validations


frontend/src/components/Memories/MemoryForm/index.jsx
angelplusultra, line 1: Since users can manually create their own memories, there should be some guidance explaining what a “Memory” actually is. The LLM is instructed in the system prompt that each memory should be a single, concise statement (one sentence maximum). It would make sense to surface similar guidance in the UI, so users don’t enter overly detailed entries like, “My name is Marcello, I am a full-stack developer who prefers code snippets in TypeScript.” Currently, there’s nothing in the interface that helps steer users toward the intended format.

frontend/src/pages/Admin/Personalization/index.jsx
```diff
@@ -0,0 +1,268 @@
...
+            onUpdateMemory={handleUpdateMemory}
+            onAddGlobal={handleAddGlobal}
+          />
+        </div>
+      </div>
+    </div>
+  );
+}
+
+function PersonalizationContent({
```
angelplusultra: Is a separate component necessary? It seems like an unnecessary layer of indirection and isn’t reused anywhere.

frontend/src/pages/WorkspaceSettings/index.jsx
```diff
@@ -67,13 +71,33 @@ function ShowWorkspaceChat() {
...
+  }
+
+  // Default users can only access the personalization tab
+  if (isDefaultUser && tab !== "personalization") {
+    if (memoryEnabled) {
+      return <Navigate to={paths.workspace.settings.personalization(slug)} />;
+    }
+    return <Navigate to={paths.workspace.chat(slug)} />;
+  }
+
```
angelplusultra: Nit: This could probably be a react router route protection middlewaree  with `<Outlet />`instead of directly in the file.

frontend/src/main.jsx
```diff
@@ -40,7 +40,7 @@ const router = createBrowserRouter([
           const { default: WorkspaceSettings } = await import(
             "@/pages/WorkspaceSettings"
           );
-          return { element: <ManagerRoute Component={WorkspaceSettings} /> };
+          return { element: <PrivateRoute Component={WorkspaceSettings} /> };
```
angelplusultra: This is where you could add an extra route protection middleware for if memory is enabled and role type

server/utils/BackgroundWorkers/index.js
```diff
@@ -15,6 +15,11 @@ class BackgroundService {
       timeout: "1m",
       interval: "12hr",
     },
+    {
+      name: "extract-memories",
+      interval: "15m",
```
angelplusultra: Consideration: 
I think it'd be handy for this to be shorter in development.

for example:
```js
interval: process.env.NODE_ENV === "development" ? "15s"  : "15m"
```

Or even better yet, maybe configurable via environment variable?

```js
interval: process.env.MEMORY_EXTRACTION_JOB_INTERVAL || "15m"
```

server/jobs/extract-memories.js
```diff
@@ -0,0 +1,159 @@
...
+
+        log(
+          `Extracted ${memories.length} memories for user ${userId} in workspace "${workspace.name}". Processed ${chatIds.length} chats.`
+        );
+      } catch (pairError) {
+        log(
+          `Error processing user ${userId}, workspace ${workspaceId}: ${pairError.message}`
+        );
+      }
+    }
```
angelplusultra: Nit: We're hitting the database twice for the same data here — once to get the distinct pairs, then again inside the loop to fetch each pair's chats. We can simplify this by fetching all
  unprocessed chats in a single query and grouping them ourselves:
```js
  const allUnprocessed = await WorkspaceChats.where(
    { memoryProcessed: false },
    null,
    { createdAt: "asc" }
  );
  if (allUnprocessed.length === 0) {
    log("No unprocessed chats found. Exiting.");
    return;
  }

  const groups = new Map();
  for (const chat of allUnprocessed) {
    const key = `${chat.user_id}:${chat.workspaceId}`;
    if (!groups.has(key)) groups.set(key, []);
    groups.get(key).push(chat);
  }

  log(
    `Found ${groups.size} user/workspace pair(s) with unprocessed chats.`
  );

  for (const [, chats] of groups) {
    const { user_id: userId, workspaceId } = chats[0];
    try {
      const workspace = await Workspace.get({ id: workspaceId });
      if (!workspace) {
        log(`Workspace ${workspaceId} not found. Skipping.`);
        continue;
      }

      const currentMemories = await Memory.forUserWorkspace(
        userId,
        workspaceId
      );

      const userMessage = buildExtractionUserMessage(
        currentMemories,
        chats
      );

      const LLMConnector = getLLMProvider({
        provider: workspace.chatProvider,
        model: workspace.chatModel,
      });

      const { textResponse } = await LLMConnector.getChatCompletion(
        [
          { role: "system", content: EXTRACTION_SYSTEM_PROMPT },
          { role: "user", content: userMessage },
        ],
        { temperature: 0.2 }
      );

      const memories = parseMemoriesResponse(textResponse);
      if (!memories) {
        log(
          `Failed to parse LLM response for user ${userId}, workspace ${workspaceId}. Skipping.`
        );
        continue;
      }

      await Memory.replaceWorkspaceMemories(userId, workspaceId, memories);

      const chatIds = chats.map((c) => c.id);
      await WorkspaceChats.markMemoryProcessed(chatIds);

      log(
        `Extracted ${memories.length} memories for user ${userId} in workspace "${workspace.name}". Processed ${chatIds.length} chats.`
      );
    } catch (pairError) {
      log(
        `Error processing user ${userId}, workspace ${workspaceId}: ${pairError.message}`
      );
    }
  }
```

  This also lets you drop `unprocessedMemoryGroups()` from WorkspaceChats since it's no longer needed.

server/endpoints/memory.js
```diff
@@ -0,0 +1,226 @@
...
+        response.status(200).json({ memories });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
+      }
+    }
+  );
+
+  app.get(
+    "/memories/:workspaceId",
```
angelplusultra: These share the same path pattern as `PUT /memories/:memoryId` and `DELETE /memories/:memoryId` — different HTTP methods so no runtime error today, but the param means two different things
  depending on the verb, which is fragile. Shouldn't the REST pattern be `GET /workspaces/:workspaceId/memories` and `POST /workspaces/:workspaceId/memories`?

angelplusultra (changes requested):
Mostly nits and refactor suggestions

frontend/src/pages/WorkspaceSettings/index.jsx
shatfield4: Going to skip this since it would require a large rewrite of how our protected routes will work since we do not use the outlet pattern yet.

frontend/src/components/Memories/MemoryForm/index.jsx
```diff
@@ -39,6 +39,9 @@ export default function MemoryForm({
         rows={2}
         className="w-full border-none bg-theme-bg-secondary placeholder:text-theme-settings-input-placeholder text-white text-sm rounded-lg focus:outline-primary-button active:outline-primary-button outline-none block p-2.5 resize-none"
       />
+      <p className="text-xs text-theme-text-secondary">
+        {t("personalization.form.hint")}
+      </p>
```
angelplusultra: 👍️

frontend/src/pages/Admin/Personalization/index.jsx
```diff
@@ -139,6 +139,78 @@ export default function Personalization() {
 
   const globalMemories = memories.filter((m) => m.scope === "global");
 
+  function renderContent() {
```
angelplusultra: My bad maybe I wasn't clear. Just have it be part of the `<Personalization />` component. My original point was just that moving all this code into a separate component added a needless layer of indirection. and this function doesn't seem to do anything but just return jsx, so why not just have it in `<Personalization />`?

server/endpoints/memory.js
angelplusultra, line 1: 👍️

server/utils/BackgroundWorkers/index.js
angelplusultra, line 1: 👍️

server/.env.example
angelplusultra, line 1: 👍️

angelplusultra (changes requested):
Just one clarification  but besides that it looks good.

frontend/src/pages/Admin/Personalization/index.jsx
shatfield4: We do need to extract this to a subcomponent/layout since there are multiple views that use the same layout with sidebar in the settings. Refactored this so it is clearer what is happening by adding a layout and wrapping all early return jsx fragments in this layout.

Reason for doing this is, reading early return jsx fragments is much easier to follow when reading through code vs conditional rendering many pieces of the UI inline.

angelplusultra (approved):
LGTM

shatfield4:
Added implementation of new UI designs.

resolves #5389

timothycarambat:
@shatfield4 This PR needs Documentation https://github.com/Mintplex-Labs/anythingllm-docs/issues/234

server/jobs/extract-memories.js
```diff
@@ -0,0 +1,160 @@
...
+  Number(process.env.MEMORY_IDLE_THRESHOLD_MS) || 20 * 60 * 1000; // 20 minutes
+
+const EXTRACTION_SYSTEM_PROMPT = `You are a memory extraction system for a personalized AI assistant. You manage a list of memories about the user that will be injected into future conversations to provide personalized responses.
+
+Instructions:
+- Review the conversations and extract any useful facts, preferences, or context about the user
+- Return an updated list of memories (max 20) that combines existing memories with any new information
+- If new information contradicts an old memory, keep only the updated version
+- If an existing memory is still relevant, keep it
+- Each memory should be a single, concise statement (1 sentence max)
```
angelplusultra: Consider surfacing similar guidance in the UI for manually created memories. Perhaps even a placeholder text with an example?

Examples fed to LLM: 
```
["User's name is Sarah", "User prefers Python over JavaScript", "User works on the Phoenix project"]`;
```
Image

server/jobs/extract-memories.js
```diff
@@ -0,0 +1,160 @@
...
+
+    const lastChat = await WorkspaceChats.get({}, null, { createdAt: "desc" });
+    if (lastChat) {
+      const msSinceLastChat =
+        Date.now() - new Date(lastChat.createdAt).getTime();
+      if (msSinceLastChat < IDLE_THRESHOLD_MS) {
+        log("User is still active. Skipping extraction.");
+        return;
+      }
+    }
```
angelplusultra: The idle threshold check looks at the most recent chat across all users and workspaces. If User A is actively chatting, User B's idle workspace (which has unprocessed chats from an hour ago) will never get processed until User A also goes idle. The intent seems to be "don't extract while the user is mid-conversation," but the check is system-wide.

  Consider checking idle per user/workspace group instead, or moving the idle check inside the for (const [, chats] of groups) loop using the last chat timestamp from each group.

shatfield4:
Docs PR: https://github.com/Mintplex-Labs/anythingllm-docs/pull/235

server/models/memory.js
```diff
@@ -0,0 +1,210 @@
...
+      await prisma.$transaction(async (tx) => {
+        await tx.memories.deleteMany({
+          where: { userId, workspaceId, scope: "workspace" },
+        });
+
+        for (const content of memories.slice(0, WORKSPACE_LIMIT)) {
+          await tx.memories.create({
+            data: { userId, workspaceId, scope: "workspace", content },
+          });
+        }
```
angelplusultra: Nit: This loops up to 20 sequential await tx.memories.create() calls inside the transaction. Since each one is its own round-trip to the database, this could be simplified to a single tx.memories.createMany() call which batches all the inserts into one query. Not a blocker but it's a free perf win.
```js

  // Current: 20 sequential INSERTs
  for (const content of memories.slice(0, WORKSPACE_LIMIT)) {
    await tx.memories.create({
      data: { userId, workspaceId, scope: "workspace", content },
    });
  }

  // Better: 1 INSERT with 20 rows
  await tx.memories.createMany({
    data: memories.slice(0, WORKSPACE_LIMIT).map((content) => ({
      userId, workspaceId, scope: "workspace", content,
    })),
  });
```

server/endpoints/memory.js
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        response.status(200).json({ memory });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
+      }
+    }
+  );
+}
```
angelplusultra, line 194: Your checking that the memory belongs to the user but you're not checking if the user has access to the workspace. A user could _technically_ demote a global memory into a workspace they don't have access to.


```js
  app.post("/memories/:memoryId/demote", [...], async (request, response) => {
    const user = await userFromSession(request, response);
    const memoryId = Number(request.params.memoryId);
    const { workspaceId } = reqBody(request);  // comes from the request body — user-controlled

    // ✅ Checks the user owns the memory
    const existing = await Memory.get({ id: memoryId });
    if (!existing || !ownerMatch(existing, user)) {
      response.status(404).json({ error: "Memory not found." });
      return;
    }

    // ❌ No check that the user belongs to this workspace
    const { memory, message } = await Memory.demoteToWorkspace(
      memoryId,
      Number(workspaceId)  // user can pass any workspace ID here
    );
  });
```

frontend/src/components/WorkspaceChat/ChatContainer/ChatSettingsMenu/index.jsx
```diff
@@ -0,0 +1,180 @@
...
+          <div
+            key={key}
+            onClick={() => onSizeChange(key)}
+            className={`px-2 py-1 rounded cursor-pointer text-sm font-normal text-white light:text-slate-800 ${
+              selectedSize === key
+                ? "bg-zinc-700 light:bg-slate-200"
+                : "hover:bg-zinc-700/50 light:hover:bg-slate-100"
+            }`}
+          >
+            {t(labelKey)}
```
angelplusultra: I would suggest not using dyamic keys for i18n. It doesnt work our static analysis for finding orphaned keys and could result in keys being deleted from the locales object if not inserted in the `DYNAMIC_KEY_ALLOWLIST`in `/frontend/locales/findUnusuedTranslationKeys.js`. The simplest fix is just use hardcoded strings for `t()`. Or add all the keys used in this in this `t()` call to the `DYNAMIC_KEY_ALLOWLIST`

frontend/src/components/WorkspaceChat/ChatContainer/MemoriesSidebar/index.jsx
```diff
@@ -0,0 +1,189 @@
...
+      </button>
+    </div>
+  );
+}
+
+function EmptyState({ onCreateClick }) {
+  return (
+    <p className="text-sm leading-5 text-zinc-400 light:text-slate-600 text-center">
+      <Trans
+        i18nKey="chat_window.memories.empty"
```
angelplusultra: be aware of i18n keys that could be pruned if not detected directly in a t() call or in the allowlist

angelplusultra (changes requested):
Some refactor suggestions and nits

frontend/src/components/WorkspaceChat/ChatContainer/MemoriesSidebar/index.jsx
timothycarambat: fwiw I added this because I noticed it a long time ago:
https://github.com/Mintplex-Labs/anything-llm/blob/master/frontend/src/locales/findUnusedTranslations.mjs#L86-L87

server/models/memory.js
timothycarambat: +1

server/endpoints/memory.js
timothycarambat: This "bug" happens a lot and everyone needs to be more mindful of it. The generic middleware prevents unauthorized requests but also check the userID for the request if the entity has a user ownership relationship.

Its exhausting fielding the reports from script kiddie 'researchers' reporting things like this for such a low-iq attack surface we can easily defend against and should do.

server/utils/agents/defaults.js
angelplusultra, line 1: Workspace memory reranking doesn't work on the agent chat path. `getDefinition` doesn't receive the user's prompt, so `promptWithMemories` is called without `prompt` or `rawHistory`, which causes it to skip reranking and just grab the 5 most recent workspace memories. The user's prompt is available upstream on `this.invocation.prompt` in `AgentHandler` — it would just need to be plumbed through to `getDefinition` so the reranker can pick the most relevant memories instead of the most recent ones

frontend/src/components/WorkspaceChat/ChatContainer/ChatSettingsMenu/index.jsx
timothycarambat: +1, lets keep moving to the factory pattern where was have like a `getTextSizes(t)` function and pass in the translator function so we dont have to worry about accidental pruning.

angelplusultra (changes requested):
Note about agent chat path missing memory reranking

server/models/memory.js
shatfield4: Resolving this since the current version of Prisma we're using does not support `createMany` with SQLite. It was added in a later version and we can bump the Prisma version to get this functionality in the future but will need to be done in another PR.

Note: Prisma released `createMany` support for SQLite in version 5.12.0 officially https://www.prisma.io/changelog/2024-04-03#implementing-createmany-for-sqlite

server/utils/memories/index.js
```diff
@@ -0,0 +1,133 @@
...
+      workspaceId
+    );
+
+    if (globalMemories.length === 0 && workspaceMemories.length === 0)
+      return "";
+
+    let selectedWorkspace = workspaceMemories;
+    if (workspaceMemories.length > INJECTED_WORKSPACE_LIMIT) {
+      // Skip reranking when there's no query context (agent calls)
+      const hasContext = prompt?.trim() || rawHistory?.length > 0;
```
angelplusultra: This comment is stale, delete it.

frontend/src/components/WorkspaceChat/ChatContainer/ChatSettingsMenu/index.jsx
```diff
@@ -6,11 +6,13 @@ import System from "@/models/system";
...
-  { key: "small", labelKey: "chat_window.small" },
-  { key: "normal", labelKey: "chat_window.normal" },
-  { key: "large", labelKey: "chat_window.large" },
-];
+function getTextSizes(t) {
+  return [
+    { key: "small", label: t("chat_window.small") },
+    { key: "normal", label: t("chat_window.normal") },
+    { key: "large", label: t("chat_window.large") },
+  ];
```
angelplusultra: 👍️

server/jobs/extract-memories.js
```diff
@@ -101,6 +91,17 @@ function parseMemoriesResponse(text) {
...
+        // Check idle per-group so one active user doesn't block extraction for others.
+        // Chats are sorted asc, so the last element is the most recent in this group.
+        const lastChatInGroup = chats[chats.length - 1];
+        const msSinceLastChat =
+          Date.now() - new Date(lastChatInGroup.createdAt).getTime();
+        if (msSinceLastChat < IDLE_THRESHOLD_MS) {
+          log(
+            `User ${userId} in workspace ${workspaceId} is still active. Skipping.`
+          );
+          continue;
```
angelplusultra: 👍️

angelplusultra (approved):
LGTM. Just delete that stale comment.

timothycarambat:
Screenshot 2026-04-22 at 7 23 41 PM

If the workspace name is long it overflows the sidebar and is side scrollable, it should truncate to some specific length to prevent this.

timothycarambat:
Screenshot 2026-04-22 at 7 25 21 PM

This can be wider as to not breakline

frontend/src/models/memory.js
```diff
@@ -0,0 +1,82 @@
...
+      headers: baseHeaders(),
+    })
+      .then((res) => res.json())
+      .catch((e) => {
+        console.error(e);
+        return { memory: null, error: e.message };
+      });
+  },
+
+  demoteToWorkspace: async function (memoryId, workspaceId) {
```
timothycarambat: JSDoc all these so we have type hints in UI

frontend/src/models/memory.js
```diff
@@ -0,0 +1,82 @@
...
+    return await fetch(`${API_BASE}/memories/${memoryId}/demote`, {
+      method: "POST",
+      headers: baseHeaders(),
+      body: JSON.stringify({ workspaceId }),
+    })
+      .then((res) => res.json())
+      .catch((e) => {
+        console.error(e);
```
timothycarambat: We dont need these error logs since we return error anyway

frontend/src/main.jsx
timothycarambat: Why did we change scope of this route?

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+async function memoryEnabled(_req, response, next) {
+  const enabled =
+    (await SystemSettings.getValueOrFallback(
+      { label: "memory_enabled" },
+      "off"
+    )) === "on";
+  if (!enabled)
+    return response.status(403).json({ error: "Personalization is disabled." });
+  next();
+}
```
timothycarambat: Rename this `memoryFeatureEnabled` and once you migrate the check for this to a method in SystemSettings like I mentioned in parent review comment you can then just call `await SystemSettings.memoriesEnabled()` and throw if not.

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+const {
+  flexUserRoleValid,
+  ROLES,
+} = require("../utils/middleware/multiUserProtected");
+
+function ownerMatch(memory, user) {
+  const memUserId = memory.userId ?? null;
+  const reqUserId = user?.id ?? null;
+  return memUserId === reqUserId;
+}
```
timothycarambat: This should _also_ just be a middleware - it should just be skipped if the system is not in MuM.

Easy to check since the first middleware is `validatedRequest` which assigns
[response.locals.multiUserMode = multiUserMode;
](https://github.com/Mintplex-Labs/anything-llm/blob/feat/memory/server/utils/middleware/validatedRequest.js#L9)

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+      try {
+        const user = await userFromSession(request, response);
+        const workspaceId = Number(request.params.workspaceId);
+
+        const workspace = multiUserMode(response)
+          ? await Workspace.getWithUser(user, { id: workspaceId })
+          : await Workspace.get({ id: workspaceId });
+        if (!workspace) {
+          response.status(403).json({ error: "Invalid workspace." });
+          return;
```
timothycarambat: `return response.status(403).json({ error: "Invalid workspace." });`

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+        }
+
+        const globalMemories = await Memory.globalForUser(user?.id);
+        const workspaceMemories = await Memory.forUserWorkspace(
+          user?.id,
+          workspaceId
+        );
```
timothycarambat: Promise.all() these to run concurrent **OR** make a single method on memory if we do this a lot.

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+    async (request, response) => {
+      try {
+        const user = await userFromSession(request, response);
+        const workspaceId = Number(request.params.workspaceId);
+        const { content, scope = "workspace" } = reqBody(request);
+
+        if (!content || !content.trim()) {
+          response.status(400).json({ error: "Content is required." });
+          return;
+        }
```
timothycarambat: This should happen after perms check

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+
+        const workspace = multiUserMode(response)
+          ? await Workspace.getWithUser(user, { id: workspaceId })
+          : await Workspace.get({ id: workspaceId });
+        if (!workspace) {
+          response.status(403).json({ error: "Invalid workspace." });
+          return;
+        }
```
timothycarambat: I notice we are doing this a lot - if we can make the input body param or URL query param `:slug` you can use this middleware: https://github.com/Mintplex-Labs/anything-llm/blob/feat/memory/server/utils/middleware/validWorkspace.js#L6-L20

Would prevent a lot of code dupe

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+        const { memory, message } = await Memory.create({
+          userId: user?.id,
+          workspaceId: scope === "global" ? null : workspaceId,
+          scope,
+          content: content.trim(),
+        });
+
+        if (!memory) {
+          response.status(400).json({ error: message });
+          return;
```
timothycarambat: `return response.status(400).json({ error: message });`

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+    async (request, response) => {
+      try {
+        const user = await userFromSession(request, response);
+        const memoryId = Number(request.params.memoryId);
+        const { content } = reqBody(request);
+
+        if (!content || !content.trim()) {
+          response.status(400).json({ error: "Content is required." });
+          return;
+        }
```
timothycarambat: Val logic after perms

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+
+        const existing = await Memory.get({ id: memoryId });
+        if (!existing || !ownerMatch(existing, user)) {
+          response.status(404).json({ error: "Memory not found." });
+          return;
+        }
```
timothycarambat: Should be its own middleware if we do this check a lot (already mentioned)

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+
+  app.delete(
+    "/memories/:memoryId",
+    [validatedRequest, flexUserRoleValid([ROLES.all]), memoryEnabled],
+    async (request, response) => {
+      try {
+        const user = await userFromSession(request, response);
+        const memoryId = Number(request.params.memoryId);
+
+        const existing = await Memory.get({ id: memoryId });
```
timothycarambat: Shouldnt this just be able to take in a userid so we can query the DB directly with either `id` or `id + userId` if user.id is not invalid?

server/endpoints/memory.js
```diff
@@ -0,0 +1,238 @@
...
+          return;
+        }
+
+        const targetWorkspace = multiUserMode(response)
+          ? await Workspace.getWithUser(user, { id: Number(workspaceId) })
+          : await Workspace.get({ id: Number(workspaceId) });
+        if (!targetWorkspace) {
+          response.status(403).json({ error: "Invalid workspace." });
+          return;
+        }
```
timothycarambat: Good defense in depth check here.

server/jobs/extract-memories.js
```diff
@@ -0,0 +1,161 @@
+const { log, conclude } = require("./helpers/index.js");
```
timothycarambat, line 1: I really truly believe here we should try to have this be an actual tool call instead of a free form response output we parse.

We can use EphemeralAgent here with a local custom plugin/tool that way we can use a proper tool call when the model can call tools and still failover to `Untooled` if not.

This would almost guarantee not wasting tokens with the model blabbering on and on before if gives us content. This tool can just return an array of strings that can be memories based on a chat convo.

server/jobs/extract-memories.js
```diff
@@ -0,0 +1,161 @@
...
+    if (enabled !== "on") {
+      log("Memory extraction is disabled. Exiting.");
+      return;
+    }
+
+    const allUnprocessed = await WorkspaceChats.where(
+      { memoryProcessed: false },
+      null,
+      { createdAt: "asc" }
+    );
```
timothycarambat, line 38: We need safety limits on this. Like if a user has 20 chats - fine sure. But if it has been a long time we should only grab the last 20 chats. Otherwise we risk doing a review on like 500 chats.

We should make everything seen even if it is not explicitly summarized just to not watch it.

## 🚨This is a big problem!!!🚨
Also this is still including chats which have been deleted via `/reset` - this should not process `include: false` chats!!!

Lastly, carefully consider the fact that you could possibly be loading a **lot** of chats even if we move to just summarizing directly that last 20 - but since we cannot determine that at query time (does `groupBy` exist? We have `userId` on workspace chats) we should instead query the chats _at the time we are about to summarize_ so that we have absolutely the chats reflective at that time and the latest info.

Also - check the workspace chat has a response - it could have died midway through and got an empty response or still be in a pending state in real time.

server/jobs/extract-memories.js
```diff
@@ -0,0 +1,161 @@
...
+          provider: workspace.chatProvider,
+          model: workspace.chatModel,
+        });
+
+        const { textResponse } = await LLMConnector.getChatCompletion(
+          [
+            { role: "system", content: EXTRACTION_SYSTEM_PROMPT },
+            { role: "user", content: userMessage },
+          ],
+          { temperature: 0.2 }
```
timothycarambat: ? Why this value (also would cause issues for some provider/models who hate temp!)

server/jobs/extract-memories.js
```diff
@@ -0,0 +1,161 @@
...
+          log(
+            `Failed to parse LLM response for user ${userId}, workspace ${workspaceId}. Skipping.`
+          );
+          continue;
+        }
+
+        await Memory.replaceWorkspaceMemories(userId, workspaceId, memories);
+
+        const chatIds = chats.map((c) => c.id);
+        await WorkspaceChats.markMemoryProcessed(chatIds);
```
timothycarambat: If a summary workflow fails, assume it to be marked processed always - otherwise we can end up with a growing list that contains some problematic edge case.

Also, how does this handle **super long** prompts and responses? We should be truncating long user prompts or responses

server/models/memory.js
```diff
@@ -0,0 +1,210 @@
+const prisma = require("../utils/prisma");
+
+const GLOBAL_LIMIT = 5;
+const WORKSPACE_LIMIT = 20;
+
+const Memory = {
+  forUserWorkspace: async function (userId, workspaceId) {
+    try {
+      const memories = await prisma.memories.findMany({
+        where: { userId, workspaceId, scope: "workspace" },
```
timothycarambat: Not just here, but all through this model you need type safety for `userId` or other fields you just do `{ field }` for since they are not done across the board and its better to have defense in depth for these things to prevent a bad implementation in the future from allowing injection at field value level

server/models/memory.js
```diff
@@ -0,0 +1,210 @@
...
+        orderBy: { createdAt: "desc" },
+      });
+      return memories;
+    } catch (error) {
+      console.error(error.message);
+      return [];
+    }
+  },
+
+  globalForUser: async function (userId) {
```
timothycarambat, line 73: JS Doc these methods

server/models/systemSettings.js
```diff
@@ -407,6 +411,9 @@ const SystemSettings = {
       JWTSecret: !!process.env.JWT_SECRET,
       StorageDir: process.env.STORAGE_DIR,
       MultiUserMode: await this.isMultiUserMode(),
+      MemoryEnabled:
+        (await this.getValueOrFallback({ label: "memory_enabled" }, "off")) ===
+        "on",
```
timothycarambat: Can use the class method once it exists like `MultiUserMode` above it for readability

server/models/workspaceChats.js
```diff
@@ -298,6 +298,31 @@ const WorkspaceChats = {
       return false;
     }
   },
+  markMemoryProcessed: async function (ids = []) {
+    if (!ids.length) return;
```
timothycarambat: Should also validate is array and they are Number()

server/prisma/migrations/20260423005842_init/migration.sql
```diff
@@ -0,0 +1,43 @@
...
+    "api_session_id" TEXT,
+    "createdAt" DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "lastUpdatedAt" DATETIME NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    "feedbackScore" BOOLEAN,
+    "memory_processed" BOOLEAN NOT NULL DEFAULT false,
+    CONSTRAINT "workspace_chats_user_id_fkey" FOREIGN KEY ("user_id") REFERENCES "users" ("id") ON DELETE CASCADE ON UPDATE CASCADE
+);
+INSERT INTO "new_workspace_chats" ("api_session_id", "createdAt", "feedbackScore", "id", "include", "lastUpdatedAt", "prompt", "response", "thread_id", "user_id", "workspaceId") SELECT "api_session_id", "createdAt", "feedbackScore", "id", "include", "lastUpdatedAt", "prompt", "response", "thread_id", "user_id", "workspaceId" FROM "workspace_chats";
+DROP TABLE "workspace_chats";
+ALTER TABLE "new_workspace_chats" RENAME TO "workspace_chats";
```
timothycarambat: I know this works but i _REALLY_ do not ever want to trust SQLite to do a full drop-and-replace of the most important table in the entire DB - you can avoid this but not adding a default field to `memory_processed`

server/prisma/schema.prisma
```diff
@@ -395,3 +398,21 @@ model external_communication_connectors {
...
   lastUpdatedAt DateTime @default(now())
 }
+
+model memories {
+  id             Int         @id @default(autoincrement())
+  userId         Int?        @map("user_id")
+  workspaceId    Int?        @map("workspace_id")
+  scope          String      @default("workspace")
+  content        String
+  sourceThreadId Int?        @map("source_thread_id")
```
timothycarambat: I looked over the code and done see really any place this field even is used in a way that matters? Where is this coming from and/or how is it used?

server/utils/agents/defaults.js
```diff
@@ -57,17 +58,29 @@ const WORKSPACE_AGENT = {
...
+    provider = null,
+    workspace = null,
+    user = null,
+    prompt = ""
+  ) => {
     const basePrompt = await Provider.systemPrompt({
       provider,
       workspace,
       user,
     });
```
timothycarambat: Can we not just augment the prompt directly with memories in this method since we already have the user and workspace potentially in there as well?

server/utils/chats/apiChatHandler.js
```diff
@@ -388,9 +389,16 @@ async function chatSync({
 
   // Compress & Assemble message to ensure prompt passes token limit with room for response
   // and build system messages based on inputs and history.
+  const systemPrompt = await promptWithMemories({
+    systemPrompt: await chatPrompt(workspace, user),
```
timothycarambat: Same here - can we augment the chat in this already async method?

server/utils/memories/index.js
```diff
@@ -0,0 +1,132 @@
+const { Memory } = require("../../models/memory");
+const { SystemSettings } = require("../../models/systemSettings");
+
+const INJECTED_WORKSPACE_LIMIT = 5;
```
timothycarambat: This needs a description - I did not realize what it did for a while

timothycarambat (changes requested):
## Reminder!
This needs a documentation PR to inform users how this works, what it does, how to manage memories - basically a short tutorial on our features so that they can be informed about this new feature to take advantage of!

Notes:
- When clicking edit on a memory the cursor goes to the start of the input and either should go to the end
- If i have 5/5 Global memories and I have the global tab focused in sidebar and I click to add another the modal closes like it was successful but nothing happens.
- the system settings check value for memories needs to be moved from `"on"|"off"` to `"true"|"false"` and reading the value should work like other fields. I know that value is a string column so it is storing that as a string but just make a helper in SystemSettings for this like the other boolean/string fields to compensate for it like - https://github.com/Mintplex-Labs/anything-llm/blob/feat/memory/server/models/systemSettings.js#L624-L632 so everything is normalized across all system settings.
- See line level comments for more

Look above for QA misses in the UI i noticed


Things I did:
- Refactored how Chat menu was broken up
- Refactored the sidebar to prevent prop drilling and cleanup DOM and nested component rendering waterfalls

server/jobs/extract-memories.js
shatfield4: 175468edd

server/jobs/extract-memories.js
shatfield4: 175468edd

server/jobs/extract-memories.js
shatfield4: 175468edd for failed processed messages

d79a92a9b for long prompts, uses truncate with max message length for each message being processed

server/models/memory.js
shatfield4: a2af10d12

server/models/systemSettings.js
shatfield4: 5ac888991

server/models/workspaceChats.js
shatfield4: 4fa63894a

server/utils/agents/defaults.js
shatfield4: 5c96c9798

server/utils/chats/apiChatHandler.js
shatfield4: 5c96c9798

server/prisma/migrations/20260423005842_init/migration.sql
shatfield4: c3fbd53c6
Made additive by just making `memoriesProcessed` optional on the `workspace_chats` table.

server/prisma/schema.prisma
shatfield4: Dropped unused `sourceThreadId` field c3fbd53c6

shatfield4:
> ## Reminder!
> This needs a documentation PR to inform users how this works, what it does, how to manage memories - basically a short tutorial on our features so that they can be informed about this new feature to take advantage of!
> 
> Notes:
> 
> * When clicking edit on a memory the cursor goes to the start of the input and either should go to the end
> * If i have 5/5 Global memories and I have the global tab focused in sidebar and I click to add another the modal closes like it was successful but nothing happens.
> * the system settings check value for memories needs to be moved from `"on"|"off"` to `"true"|"false"` and reading the value should work like other fields. I know that value is a string column so it is storing that as a string but just make a helper in SystemSettings for this like the other boolean/string fields to compensate for it like - https://github.com/Mintplex-Labs/anything-llm/blob/feat/memory/server/models/systemSettings.js#L624-L632 so everything is normalized across all system settings.
> * See line level comments for more
> 
> Look above for QA misses in the UI i noticed
> 
> Things I did:
> 
> * Refactored how Chat menu was broken up
> * Refactored the sidebar to prevent prop drilling and cleanup DOM and nested component rendering waterfalls

Docs were already done just got lost in the commit history messages: https://github.com/Mintplex-Labs/anything-llm/pull/5269#issuecomment-4238646555

Everything should be resolved now.

PR #5295, opened 2026-03-28
feat: add custom base URL support for Anthropic provider

### LINKED ISSUE #5234: [FEAT]: Add custom base URL for Anthropic provider and optional model discovery for Generic OpenAI

I would like to request two related improvements for self-hosted / compatible API backends:

1. **Add custom base URL / custom host support to the Anthropic provider**
2. **Add optional model discovery / model selection support to the Generic OpenAI provider**

### Why is this needed?

There are now local/self-hosted backends that expose APIs compatible with OpenAI and/or Anthropic.

For example, some backends support:

- OpenAI-compatible `/v1/chat/completions`
- OpenAI-compatible `/v1/models`
- Anthropic-compatible `/v1/messages`
- streaming
- reasoning / thinking
- tool calling

A good example is **oMLX**, which advertises itself as a drop-in replacement for both OpenAI and Anthropic APIs.

#### 1. Anthropic provider
The Anthropic connector appears to be tied to Anthropic-hosted endpoints only, with no way to set a custom host / base URL.

That makes it impossible to use Anthropic-compatible self-hosted backends directly.

#### 2. Generic OpenAI provider
The Generic OpenAI connector works with custom hosts, but it appears to require a manually entered model name and does not seem to expose the same model selection UX as more specific providers.

This means that even when a backend supports `GET /v1/models`, users cannot conveniently pick from available models in the UI.

#### Anthropic provider
Please allow:
- API Key
- Model
- **Base URL / Custom Host**

Example:
- `https://api.anthropic.com`
- `http://localhost:8080`
- `https://my-self-hosted-backend.example.com`

#### Generic OpenAI provider
Please optionally support:
- calling `GET /v1/models`
- populating the model dropdown from that endpoint
- falling back to manual model entry when model discovery is unavailable

### Why this would help

This would make AnythingLLM much more flexible for:

- self-hosted backends
- OpenAI-compatible servers
- Anthropic-compatible servers
- private/local inference infrastructure
- advanced users who do not want to rely only on hosted APIs

It would also reduce friction when testing backends that already expose standards-compatible endpoints.

### Additional context

Right now, Generic OpenAI is often the best workaround for self-hosted backends because it supports custom hosts, but it may miss provider-specific UX features such as model selection.

At the same time, the Anthropic provider could be a better fit for Anthropic-compatible servers, but only if custom base URLs are supported.

satishkc7:
Closes #5234

## What

Adds an optional `ANTHROPIC_BASE_URL` setting that lets users route the Anthropic provider to any Anthropic-compatible backend instead of the default `api.anthropic.com`.

## Why

Self-hosted inference servers (oMLX, Azure AI Foundry, local proxies) expose Anthropic-compatible `/v1/messages` APIs but have no way to be used with the Anthropic provider today. Users are forced to use Generic OpenAI as a workaround, losing Anthropic-specific UX (model picker, prompt caching, etc.).

## Changes

| File | Change |
|------|--------|
| `server/utils/AiProviders/anthropic/index.js` | Pass `baseURL` to `AnthropicAI` constructor and `fetchModelMaxTokens` when `ANTHROPIC_BASE_URL` is set |
| `server/utils/helpers/updateENV.js` | Register `AnthropicBaseURL` → `ANTHROPIC_BASE_URL` with `isValidURL` check; relax `sk-ant-` key format check when a custom base URL is present |
| `frontend/src/components/LLMSelection/AnthropicAiOptions/index.jsx` | Add "Custom Base URL" input in the existing advanced settings panel |
| `server/.env.example` / `docker/.env.example` | Document `ANTHROPIC_BASE_URL` |

## Review notes

- No changes to existing behavior when `ANTHROPIC_BASE_URL` is unset — the SDK uses its default endpoint
- The API key format check (`sk-ant-`) is bypassed only when a custom base URL is configured, since third-party backends accept arbitrary key strings
- The UI field is placed alongside the existing Prompt Caching control in the advanced settings section to keep the primary settings clean

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -28,6 +28,7 @@ GID='1000'
 # ANTHROPIC_API_KEY=sk-ant-xxxx
 # ANTHROPIC_MODEL_PREF='claude-2'
 # ANTHROPIC_CACHE_CONTROL="5m" # Enable prompt caching (5m=5min cache, 1h=1hour cache). Reduces costs and improves speed by caching system prompts.
+# ANTHROPIC_BASE_URL= # Optional: override the Anthropic API base URL for self-hosted or compatible backends (e.g. http://localhost:8080)
 
 # LLM_PROVIDER='lmstudio'
 # LMSTUDIO_BASE_PATH='http://your-server:1234/v1'
--- a/frontend/src/components/LLMSelection/AnthropicAiOptions/index.jsx
+++ b/frontend/src/components/LLMSelection/AnthropicAiOptions/index.jsx
@@ -84,6 +84,27 @@ export default function AnthropicAiOptions({ settings }) {
               </option>
             </select>
           </div>
+          <div className="flex flex-col w-60">
+            <label className="text-white text-sm font-semibold block mb-3">
+              Custom Base URL{" "}
+              <span className="font-normal text-theme-text-secondary">
+                (optional)
+              </span>
+            </label>
+            <input
+              type="url"
+              name="AnthropicBaseURL"
+              className="border-none bg-theme-settings-input-bg text-white placeholder:text-theme-settings-input-placeholder text-sm rounded-lg focus:outline-primary-button active:outline-primary-button outline-none block w-full p-2.5"
+              placeholder="https://api.anthropic.com"
+              defaultValue={settings?.AnthropicBaseURL ?? ""}
+              autoComplete="off"
+              spellCheck={false}
+            />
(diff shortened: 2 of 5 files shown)


timothycarambat:
The Anthropic provider is **only** for Anthropic intentionally. If we want to support this it has to be through a new Generic like OpenAI - or move Generic OpenAi to just Generic and then have the endpoint be anthropic, openai, or even ollama compatible

PR #5296, opened 2026-03-29
fix: ensure text wraps properly in narrow windows

### LINKED ISSUE #5291: [BUG]: ALLM window cuts off text instead of wrapping it.

Desktop App

### What happened?

I have upgraded to ALLM v1.11.2. I tend to make my windows smaller on my desktop in Windows 11. 

When I make the window small enough to see past chats in a workspace the text is cut off, it does not wrap. Both my question and AI answers should wrap to the window size. 

Screen shot: 

1. Start ALLM.
2. Open a workspace with existing questions and answers.
3. Make the window skinnier. The text will not wrap when the ALLM main window gets too skinny.

cyphercodes:
## Problem
When the ALLM desktop window is made narrow, chat message text gets cut off instead of wrapping properly (Issue #5291).

## Solution
Added `overflow-wrap: break-word` and `word-break: break-word` CSS properties to the `.markdown` class to ensure text wraps correctly at any window width.

## Changes
- Updated `frontend/src/index.css` to add text wrapping properties

## Testing
- Text now wraps properly when window is resized to narrow widths
- Both user messages and AI responses display correctly without being cut off

Fixes #5291

### DIFF
--- a/frontend/src/index.css
+++ b/frontend/src/index.css
@@ -856,6 +856,8 @@ dialog::backdrop {
 
 .markdown {
   text-wrap: wrap;
+  overflow-wrap: break-word;
+  word-break: break-word;
 }
 
 .markdown pre {


timothycarambat:
This does not solve the issue. If you look at the issue this is not an overflow issue, but a breakpoint one.

PR #5310, opened 2026-03-31
fix: coerce env values to string before .includes() check in updateENV

xandr0s:
## Summary
- Fix crash (500) in `POST /api/v1/system/update-env` when numeric values are passed
- Root cause: `.includes("******")` called on Number type — Number has no `.includes()` method
- Fix: wrap with `String()` before calling `.includes()`

## Steps to reproduce
```bash
curl -X POST /api/v1/system/update-env   -d '{"GenericOpenAiEmbeddingMaxConcurrentChunks": 1}'
# => 500 Internal Server Error
# => TypeError: newENVs[key].includes is not a function
```

## Change
```diff
- (key) => validKeys.includes(key) && !newENVs[key].includes("******")
+ (key) => validKeys.includes(key) && !String(newENVs[key]).includes("******")
```

## Test plan
- [ ] POST numeric value to `/api/v1/system/update-env` — no crash, value saved
- [ ] POST string with `******` — still filtered correctly
- [ ] POST normal string — saved as before

### DIFF
--- a/server/utils/helpers/updateENV.js
+++ b/server/utils/helpers/updateENV.js
@@ -1196,7 +1196,7 @@ async function updateENV(newENVs = {}, force = false, userId = null) {
   const runAfterAll = [];
   const validKeys = Object.keys(KEY_MAPPING);
   const ENV_KEYS = Object.keys(newENVs).filter(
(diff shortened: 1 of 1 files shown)


timothycarambat:
You shouldn't be using the frontend API directly outside the UI because of things like this - what is the use case you are working for where you need to manipulate the update-env directly?

PR #5322, opened 2026-04-01
feat: Scheduled Jobs

### LINKED ISSUE #5349: [FEAT]: Scheduled Jobs — recurring AI tasks on a cron schedule
## What would you like to see?

**Scheduled Jobs** — the ability to create recurring AI tasks that automatically run prompts with optional agent tools on a cron schedule.

### Use cases
- Automated daily summaries or reports generated by an agent
- Periodic data checks or monitoring tasks using agent tools (web scraping, API calls, etc.)
- Scheduled content generation or analysis workflows
- Any repetitive agent task a user currently has to trigger manually

### Proposed capabilities
- **CRUD management** — create, edit, delete, enable/disable jobs with cron scheduling
- **Per-job tool selection** — choose which agent skills (built-in, MCP servers, agent flows, imported plugins) each job can use, or default to all enabled skills
- **Execution trace capture** — full visibility into each run: agent thinking, tool calls with results, final response, and usage metrics
- **Run history** — view past runs with read/unread tracking
- **Manual trigger** — "Run Now" button for immediate execution regardless of enabled state
- **Continue in thread** — fork a completed run into a workspace thread for follow-up conversation
- **Push notifications** — notify users when a job completes via web push
- **Concurrency control** — configurable limit on how many jobs can run in parallel (`SCHEDULED_JOB_MAX_CONCURRENT`)
- **Timeout** — configurable max execution time per job (`SCHEDULED_JOB_TIMEOUT_MS`, default 5 min)
- **Cold startup recovery** — orphaned runs marked as failed, stale schedules recomputed on server restart

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)

### Relevant Issues

resolves #5349
Docs PR: https://github.com/Mintplex-Labs/anythingllm-docs/pull/239
Translations PR: #5482 

### Description

Adds **Scheduled Jobs** — recurring AI tasks that run prompts with optional agent tools on a cron schedule (single-user mode only).

Users can create jobs with a name, prompt, cron schedule (visual builder or raw expression), and optional tool selection from all configured agent skills (built-in, MCP servers, agent flows, imported plugins). Jobs run automatically, and each execution captures a full trace — agent thinking, tool calls with results, generated files, final response, and token metrics.

**Key capabilities:**
- Create / edit / delete / enable / disable jobs
- Per-job tool selection from all configured agent skills
- Run history (last 50 per job) with read/unread tracking and detail view
- "Run Now" for immediate execution; "Continue in Thread" to fork a completed run into a workspace thread
- Web push notification when a job completes

**Architecture highlights:**
- Cron timers run in-process via `later.setInterval` (not Bree's built-in scheduling) so executions can be routed through a `p-queue` for concurrency control before spawning workers via `runJob()`. Bree couples scheduling with worker spawning, leaving no hook to intercept.
- Worker uses `EphemeralAgentHandler` (same runtime as the Telegram bot) with a new `toolOverrides` arg for per-job tool restriction.
- New `onToolCallResult` hook on AIbitat captures tool outputs into the execution trace.
- On boot: orphaned "running" runs are marked failed and `nextRunAt` is recomputed for all enabled jobs.

**Config:**
- `SCHEDULED_JOB_MAX_CONCURRENT` (default 1)
- `SCHEDULED_JOB_TIMEOUT_MS` (default 300000 / 5 min)

**Route protection:** New `SingleUserRoute` wrapper in `PrivateRoute/index.jsx` mirroring `AdminRoute`/`ManagerRoute`. Server-side endpoints are guarded by the existing `isSingleUserMode` middleware.

**Database:** Two new tables — `scheduled_jobs` and `scheduled_job_runs` (FK + cascade delete). Migration under `server/prisma/migrations/`.

**i18n:** Fully localized under `scheduledJobs.*` in `frontend/src/locales/en/common.js`. Run `node frontend/src/locales/normalizeEn.mjs` after merge to propagate.

### Additional Information

- Named "Scheduled Jobs" rather than "Background Jobs" (as in the PRD) to avoid confusion with the existing `BackgroundService`/Bree infrastructure.
- Single-user mode only for v1. No multi-user, no retry mechanism.
- Run history is capped at 50 most recent runs per job; pagination can be added later if needed.

### Demo of Job Creation, Job Run Inspection and Forwarding Job Run to New Thread

https://github.com/user-attachments/assets/b70834c9-f8be-487c-bf35-c60e48436903

### Demo of Job Completion Push Notification
https://github.com/user-attachments/assets/b9a74074-292f-4d0e-9647-399a632c700a

### Development
This PR contains new DB migrations, make sure to run \`yarn prisma migrate dev\` before testing the feature.

### Developer Validations

- [x] I ran \`yarn lint\` from the root of the repo & committed changes


frontend/src/pages/GeneralSettings/ScheduledJobs/index.jsx
```diff
@@ -0,0 +1,159 @@
...
+
+  useEffect(() => {
+    fetchJobs();
+  }, []);
+
+  // Poll every 5s while tab is visible so status badges and run timestamps stay in sync.
+  usePolling(fetchJobs, 5000);
+
+  const handleDelete = async (id) => {
+    if (!window.confirm(t("scheduledJobs.confirmDelete"))) return;
```
angelplusultra: should this be an application modal?

mekler22:
Is this already integrated into the latest docker image? As I cannot seem to run a scheduled or recurring task. Also, doesn't it need a system clock for this which is missing in anythingllm, or maybe in vllm?

timothycarambat:
@mekler22 you are looking at PR, so that would not appear in latest. Latest only is what is in `master` - not every single PR. Also Node uses the system clock and every OS has a system clock. I am not sure what you mean

frontend/src/pages/GeneralSettings/ScheduledJobs/RunDetailPage.jsx
```diff
@@ -0,0 +1,331 @@
...
+
+                {result.toolCalls?.length && (
+                  <ToolCallsSection t={t} result={result} />
+                )}
+
+                {result.generatedFiles?.length && (
+                  <GeneratedFilesSection t={t} result={result} />
+                )}
+
+                {result.text && <FinalResponseSection t={t} result={result} />}
```
shatfield4: This part of the frontend is rendering a 0 in black text when viewing the run details.

Image

server/models/scheduledJob.js
```diff
@@ -0,0 +1,226 @@
+const prisma = require("../utils/prisma");
+
+const ScheduledJob = {
+  writable: ["name", "prompt", "tools", "schedule", "enabled"],
+
+  // Maximum number of scheduled jobs that can be enabled at once.
+  // null = no limit. Set to a positive integer to cap concurrent active jobs;
+  // attempting to enable a job past the cap will be rejected at the API layer.
+  MAX_ACTIVE: null,
+
```
shatfield4, line 23: I believe this is dead code and not used anymore so lets clean up the logic here for the MAX_ACTIVE variable on the ScheduledJob object. We can remove MAX_ACTIVE, countActive, canActivate, and all three canActivate callsites in the endpoints file.

I suggest wiring this up to the SCHEDULED_JOB_MAX_ACTIVE variable and document this in the .env.example

server/jobs/run-scheduled-job.js
```diff
@@ -0,0 +1,221 @@
...
+    const TIMEOUT =
+      Number(process.env.SCHEDULED_JOB_TIMEOUT_MS) || 5 * 60 * 1000;
+    const startTime = Date.now();
+
+    await Promise.race([
+      agentHandler.startAgentCluster(),
+      new Promise((_, reject) =>
+        setTimeout(() => reject(new Error("SCHEDULED_JOB_TIMEOUT")), TIMEOUT)
+      ),
+    ]);
```
shatfield4: If the agent finishes first here, the setTimeout is still armed and keeps event loop alive until it fires. This means the event loop is still alive until the 5 minute timeout fires keeping these loops alive for 5 minutes after this is called. We should take the timer id and clearTimeout in a .finally to make sure this is cleaned up properly.

server/models/scheduledJob.js
```diff
@@ -0,0 +1,226 @@
...
+   * Compute the next run time from a cron expression.
+   * Uses @breejs/later which is already available via Bree.
+   * @param {string} cronExpression
+   * @returns {Date|null}
+   */
+  computeNextRunAt: function (cronExpression) {
+    try {
+      const later = require("@breejs/later");
+      const sched = later.parse.cron(cronExpression);
+      const next = later.schedule(sched).next(1);
```
shatfield4: Can we move these imports to the top level to avoid importing it on each call?

server/models/scheduledJob.js
```diff
@@ -0,0 +1,226 @@
...
+
+  /**
+   * Validate a cron expression.
+   * Uses cron-validate which is already available via Bree.
+   * @param {string} cronExpression
+   * @returns {boolean}
+   */
+  isValidCron: function (cronExpression) {
+    try {
+      const cron = require("cron-validate").default;
```
shatfield4: Can we move this import to top level as well?

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,462 @@
...
+          tools: tools || null,
+          schedule: schedule.trim(),
+        });
+
+        if (error) {
+          return response.status(400).json({ job: null, error });
+        }
+
+        const { BackgroundService } = require("../utils/BackgroundWorkers");
+        new BackgroundService().addScheduledJob(job);
```
shatfield4: Instead of creating a new BackgroundService like this in every single endpoint, can we instatiate it at the top level and add a function to the BackgroundService to .getInstance() ?

This would allow us always reuse the same object instead of creating a new instance in every single endpoint.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,462 @@
...
+          Number(request.params.id),
+          updates
+        );
+
+        if (error) {
+          return response.status(400).json({ job: null, error });
+        }
+
+        const { BackgroundService } = require("../utils/BackgroundWorkers");
+        await new BackgroundService().syncScheduledJob(job.id);
```
shatfield4: Same here if we have a .getInstance() function we can grab the existing instance of BackgroundService.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,462 @@
...
+  );
+
+  // Delete a scheduled job
+  app.delete(
+    "/scheduled-jobs/:id",
+    [validatedRequest, isSingleUserMode],
+    async (request, response) => {
+      try {
+        const { BackgroundService } = require("../utils/BackgroundWorkers");
+        new BackgroundService().removeScheduledJob(Number(request.params.id));
```
shatfield4: Same here if we have a .getInstance() function we can grab the existing instance of BackgroundService.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,462 @@
...
+            });
+          }
+        }
+
+        const { job: updated } = await ScheduledJob.update(job.id, {
+          enabled: !job.enabled,
+        });
+
+        const { BackgroundService } = require("../utils/BackgroundWorkers");
+        await new BackgroundService().syncScheduledJob(job.id);
```
shatfield4: Same here if we have a .getInstance() function we can grab the existing instance of BackgroundService.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,462 @@
...
+      try {
+        const job = await ScheduledJob.get({
+          id: Number(request.params.id),
+        });
+        if (!job) {
+          return response.status(404).json({ error: "Job not found" });
+        }
+
+        const { BackgroundService } = require("../utils/BackgroundWorkers");
+        new BackgroundService().enqueueScheduledJob(job.id);
```
shatfield4: Same here if we have a .getInstance() function we can grab the existing instance of BackgroundService.

server/jobs/run-scheduled-job.js
```diff
@@ -0,0 +1,221 @@
...
+
+    // Wait for handler close (agent termination)
+    await Promise.race([
+      handlerPromise,
+      new Promise((resolve) => setTimeout(resolve, 5000)),
```
shatfield4: Do we need to keep this 5 second timeout? I think we can remove this because we already await the handlerPromise and it will continue after it completes.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
```diff
@@ -0,0 +1,104 @@
...
+      <td className="px-6 py-4 text-theme-text-secondary text-xs">
+        {job.enabled && job.nextRunAt
+          ? new Date(job.nextRunAt).toLocaleString()
+          : "—"}
+      </td>
+      <td className="px-6 py-4">
+        <div className="flex items-center justify-end gap-2">
+          <button
+            type="button"
+            onClick={() => navigate(`/settings/scheduled-jobs/${job.id}/runs`)}
```
shatfield4: Add this as a function to `paths.js` instead of doing hardcoded paths for navigation.

frontend/src/pages/GeneralSettings/ScheduledJobs/RunDetailPage.jsx
```diff
@@ -0,0 +1,331 @@
...
+            </p>
+          ) : (
+            <>
+              <RunHeader
+                t={t}
+                job={job}
+                run={run}
+                result={result}
+                continuing={continuing}
+                onBack={() => navigate(`/settings/scheduled-jobs/${id}/runs`)}
```
shatfield4: Add this to `paths.js` instead of hardcoding the path.

shatfield4 (changes requested):
A few places where we can clean up logic and remove dead code.

server/models/scheduledJob.js
angelplusultra: Good catch, but in this case this is not dead code. This will be used later when this feature gets ported to Desktop. Tim told me to put this in here now, when its ported to desktop it will hold a hardcoded value like 3. It is intended to be static and non configurable.

server/endpoints/scheduledJobs.js
angelplusultra: `new BackgroundService()` does not create a whole new object in memory , if you look at the constructor for the class, its using the singleton pattern.
 
 ```js
 constructor() {
    if (BackgroundService._instance) {
      this.#log("SINGLETON LOCK: Using existing BackgroundService.");
      return BackgroundService._instance;
    }

    this.logger = setLogger();
    BackgroundService._instance = this;
  }
  ```
 
  I agree that it might be misleading and a `getInstance()` method would be  more clear that its a singleton rather than overwriting how the `new` keyword works , but I didn't write the `BackgroundService` class, just extending it here for this feature. Applying that fix here would mean refactoring every call site of `new BackgroundService()` which would be out of scope for this feature for now.

frontend/src/pages/GeneralSettings/ScheduledJobs/RunDetailPage.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/4cfea8f9299149938e1b86e248e88c411273b4af

server/jobs/run-scheduled-job.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/b1d915db2696642e6751e4bede0e6a860a861dcb

server/jobs/run-scheduled-job.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/b36facffef0de3fece9c82330792ba3764aaed7b

frontend/src/pages/GeneralSettings/ScheduledJobs/RunDetailPage.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/5a9635912d66c9f89c205a9cb3a9f72661be2599

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/5a9635912d66c9f89c205a9cb3a9f72661be2599

server/models/scheduledJob.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/33c057583ebd258945a92f3ed8f563deeebb5cc0

server/models/scheduledJob.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/33c057583ebd258945a92f3ed8f563deeebb5cc0

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
```diff
@@ -0,0 +1,257 @@
...
+        <div className="relative">
+          <input
+            type="text"
+            value={search}
+            onChange={(e) => setSearch(e.target.value)}
+            onFocus={() => setOpen(true)}
+            onClick={() => setOpen(true)}
+            placeholder={t("scheduledJobs.modal.toolsSearch", "Search")}
+            className="border border-transparent light:border-slate-300 bg-zinc-800 light:bg-white text-zinc-300 light:text-slate-700 placeholder:text-zinc-400 light:placeholder:text-slate-500 text-sm rounded-lg focus:outline-sky-500 outline-none block w-full px-3.5 py-2.5 pr-9"
+          />
```
shatfield4, line 190: UI bug here for when on smaller screens, we should make the tools selector popup scrollable so users can select all tools they need. 

Image

frontend/src/locales/en/common.js
```diff
@@ -1401,6 +1402,166 @@ const TRANSLATIONS = {
...
+        fri: "Fri",
+        sat: "Sat",
+      },
+    },
+    runHistory: {
+      back: "Back to jobs",
+      title: "Run History: {{name}}",
+      schedule: "Schedule:",
+      emptyTitle: "No runs yet for this job",
+      emptySubtitle: "Run the job now and view it's results.",
```
shatfield4: Typo, should be "its".

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,462 @@
...
+        new BackgroundService().removeScheduledJob(Number(request.params.id));
+
+        const success = await ScheduledJob.delete(Number(request.params.id));
+        return response.status(200).json({ success });
+      } catch (e) {
+        console.error(e.message, e);
+        response.sendStatus(500);
+      }
+    }
+  );
```
shatfield4, line 289: Found a bug here when deleting a scheduled job that is in progress. It looks like `removeScheduledJob()` is only clearing the cron from running again on the specified interval.

When the in progress job completes, the cascading delete has already deleted the job that the worker is targeting to update which throws a prisma error. Lets make sure to kill any in progress jobs for the scheduled job we are deleting.

server/models/scheduledJob.js
```diff
@@ -0,0 +1,226 @@
+const prisma = require("../utils/prisma");
+const later = require("@breejs/later");
+const cronValidate = require("cron-validate").default;
```
shatfield4, line 3: For these imports, I don't see them directly added to the package.json. Should we make them dependencies that are explicitly defined in case the package that adds them as a dependency is removed in the future?

server/utils/BackgroundWorkers/index.js
```diff
@@ -1,13 +1,19 @@
 const path = require("path");
 const Graceful = require("@ladjs/graceful");
 const Bree = require("@mintplex-labs/bree");
+const later = require("@breejs/later");
+const PQueue = require("p-queue").default;
```
shatfield4, line 5: Looks like this is also not explicitly defined in the package.json either.

server/utils/BackgroundWorkers/index.js
```diff
@@ -163,6 +188,135 @@ class BackgroundService {
...
+   * @param {object} job - scheduled_jobs DB record
+   */
+  addScheduledJob(job) {
+    this.removeScheduledJob(job.id);
+    const sched = later.parse.cron(job.schedule);
+    const timer = later.setInterval(() => {
+      this.enqueueScheduledJob(job.id);
+    }, sched);
+    this.#scheduledJobTimers.set(job.id, timer);
+  }
```
shatfield4, line 260: Does `later.setInterval` bypass `SCHEDULED_JOB_MAX_CONCURRENT`? It looks like all cron timers fire independently and hit the database regardless of the concurrency limit.

server/endpoints/scheduledJobs.js
shatfield4: I saw your note on the previous review comment about adding a `getInstance` function so that completely makes sense why you wouldn't want to do that here. Can we at least move the imports for each of the `new BackgroundService` to the top instead of importing it in every single endpoint? If I am missing something here please correct me but I saw that `BackgroundService` is already a singleton class so when calling `new BackgroundService` it is actually already grabbing the existing instance to prevent multiple instantiations.

server/endpoints/scheduledJobs.js
shatfield4: Move import to top of file.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,462 @@
...
+              error: `Cannot enable: maximum of ${activation.limit} active scheduled jobs reached. Disable another job first.`,
+            });
+          }
+        }
+
+        const { job: updated } = await ScheduledJob.update(job.id, {
+          enabled: !job.enabled,
+        });
+
+        const { BackgroundService } = require("../utils/BackgroundWorkers");
```
shatfield4: Move import to top of file.

server/endpoints/scheduledJobs.js
shatfield4: Move import to top of file.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,462 @@
...
+          const wsResult = await Workspace.new("Scheduled Jobs");
+          workspace = wsResult.workspace;
+          if (!workspace) {
+            return response
+              .status(500)
+              .json({ error: "Failed to create workspace" });
+          }
+        }
+
+        const threadName = `${job.name} - ${new Date(run.startedAt).toLocaleString()}`;
```
shatfield4: Since this runs inside the worker env, if a user was self hosting this on a remote server it may give them a different timezone than they are expecting. Can we use `.toISOString()` instead so that when we display it to the user we can display it in the timezone they are expecting? Very minor thing but could cause some confusing for the user.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/CollapsibleSection.jsx
```diff
@@ -0,0 +1,37 @@
...
+  children,
+  defaultOpen = false,
+}) {
+  const [open, setOpen] = useState(defaultOpen);
+  return (
+    <div className="border border-zinc-700 light:border-slate-400 rounded-lg overflow-hidden">
+      <button
+        type="button"
+        onClick={() => setOpen(!open)}
+        className="w-full h-12 flex items-center gap-2 px-[18px] hover:bg-white/5 light:hover:bg-slate-100 transition-colors text-left"
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/GeneratedFileCard.jsx
```diff
@@ -0,0 +1,139 @@
...
+          </p>
+        </div>
+      </div>
+      <button
+        type="button"
+        onClick={handleDownload}
+        disabled={downloading}
+        title={t("scheduledJobs.file.download")}
+        aria-label={t("scheduledJobs.file.download")}
+        className="text-zinc-400 light:text-slate-600 hover:text-zinc-50 light:hover:text-slate-950 transition-colors shrink-0 ml-4 disabled:opacity-50"
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
```diff
@@ -0,0 +1,107 @@
...
+        {job.enabled && job.nextRunAt
+          ? new Date(job.nextRunAt).toLocaleString()
+          : "—"}
+      </span>
+      <div className="w-[140px] flex items-center justify-end gap-3.5">
+        <button
+          type="button"
+          onClick={stop(() => onDelete(job.id))}
+          className="text-zinc-400 light:text-slate-950 hover:text-red-400 light:hover:text-red-600 transition-colors"
+          title={t("scheduledJobs.row.delete")}
```
shatfield4, line 63: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
```diff
@@ -0,0 +1,107 @@
...
+        >
+          <X className="h-4 w-4" />
+        </button>
+        <button
+          type="button"
+          onClick={stop(() => onEdit(job))}
+          className="text-zinc-400 light:text-slate-950 hover:text-white light:hover:text-slate-700 transition-colors"
+          title={t("scheduledJobs.row.edit")}
```
shatfield4, line 71: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
```diff
@@ -0,0 +1,107 @@
...
+        >
+          <PencilSimple className="h-4 w-4" />
+        </button>
+        <button
+          type="button"
+          onClick={stop(() => onTrigger(job.id))}
+          disabled={inFlight}
+          className="text-zinc-400 light:text-slate-950 hover:text-white light:hover:text-slate-700 transition-colors disabled:opacity-40 disabled:cursor-not-allowed"
+          title={t("scheduledJobs.row.runNow")}
+        >
```
shatfield4, line 81: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
```diff
@@ -0,0 +1,107 @@
...
+          onClick={stop(() => onToggle(job.id))}
+          title={
+            job.enabled
+              ? t("scheduledJobs.row.disable")
+              : t("scheduledJobs.row.enable")
+          }
+          className={`relative h-[15px] w-7 rounded-full p-0.5 transition-colors ${
+            job.enabled ? "bg-green-400" : "bg-zinc-600 light:bg-slate-300"
+          }`}
+        >
```
shatfield4, line 97: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/RunRow.jsx
```diff
@@ -0,0 +1,58 @@
...
+    !run.readAt && run.status !== "running" && run.status !== "queued";
+
+  return (
+    <button
+      type="button"
+      onClick={() =>
+        navigate(paths.settings.scheduledJobRunDetail(jobId, run.id))
+      }
+      className="flex items-center px-4 h-14 hover:bg-white/5 light:hover:bg-slate-200 transition-colors text-left w-full"
+    >
```
shatfield4, line 63: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/components/ToolCallCard.jsx
```diff
@@ -0,0 +1,115 @@
...
+        </div>
+      )}
+
+      {resultText && (
+        <div>
+          <button
+            type="button"
+            onClick={() => setShowResult(!showResult)}
+            className="text-xs text-blue-400 hover:text-blue-300 transition-colors"
+          >
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/CronBuilder.jsx
```diff
@@ -0,0 +1,199 @@
...
+                      ? state.weekdays.filter((d) => d !== day.value)
+                      : [...state.weekdays, day.value];
+                    update({ weekdays: next.length ? next : [day.value] });
+                  }}
+                  className={`px-3 py-1 text-xs rounded-full border transition-colors ${
+                    selected
+                      ? "bg-primary-button text-white border-primary-button"
+                      : "bg-transparent text-theme-text-secondary border-white/10 hover:border-white/30"
+                  }`}
+                >
```
shatfield4, line 164: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+              <p className="text-xs leading-4 text-zinc-400 light:text-slate-600 max-w-[700px]">
+                {t("scheduledJobs.description")}
+              </p>
+            </div>
+            {!loading && jobs.length > 0 && (
+              <button
+                type="button"
+                onClick={handleCreate}
+                className="h-9 px-5 rounded-lg bg-zinc-50 text-zinc-950 light:bg-slate-900 light:text-white text-sm font-medium hover:bg-zinc-200 light:hover:bg-slate-800 transition-colors shrink-0"
+              >
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/index.jsx
```diff
@@ -0,0 +1,175 @@
...
+                    </p>
+                    <p className="text-sm font-medium text-zinc-400 light:text-slate-600">
+                      {t("scheduledJobs.emptySubtitle")}
+                    </p>
+                  </div>
+                  <button
+                    type="button"
+                    onClick={handleCreate}
+                    className="h-9 px-5 rounded-lg bg-zinc-50 text-zinc-950 light:bg-slate-900 light:text-white text-sm font-medium hover:bg-zinc-200 light:hover:bg-slate-800 transition-colors"
+                  >
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/NewJobModal.jsx
```diff
@@ -0,0 +1,283 @@
...
+          <div className="flex items-start justify-between">
+            <h3 className="text-xl font-semibold text-theme-text-primary">
+              {isEditing
+                ? t("scheduledJobs.modal.titleEdit")
+                : t("scheduledJobs.modal.titleNew")}
+            </h3>
+            <button
+              onClick={onClose}
+              type="button"
+              className="transition-all duration-300 text-gray-400 bg-transparent hover:border-white/60 rounded-lg text-sm p-1.5 ml-auto inline-flex items-center"
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/NewJobModal.jsx
```diff
@@ -0,0 +1,283 @@
...
+              ].map((tab) => (
+                <button
+                  key={tab.value}
+                  type="button"
+                  onClick={() => handleModeChange(tab.value)}
+                  className={`px-3 py-1 text-xs rounded-md transition-colors ${
+                    form.scheduleMode === tab.value
+                      ? "bg-zinc-50 text-zinc-950 light:bg-zinc-950 light:text-white"
+                      : "text-theme-text-secondary hover:text-theme-text-primary"
+                  }`}
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/NewJobModal.jsx
```diff
@@ -0,0 +1,283 @@
...
+              onChange={setSelectedTools}
+            />
+          )}
+
+          <div className="flex items-center justify-between pt-4">
+            <button
+              type="button"
+              onClick={onClose}
+              className="h-[34px] px-3.5 text-sm font-medium text-zinc-50 light:text-slate-900 border border-zinc-700 light:border-slate-600 rounded-lg hover:bg-zinc-800 light:hover:bg-slate-100 transition-colors"
+            >
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/NewJobModal.jsx
```diff
@@ -0,0 +1,283 @@
...
+              {t("scheduledJobs.modal.cancel")}
+            </button>
+            <button
+              type="submit"
+              disabled={saving}
+              className="h-[34px] px-3.5 text-sm font-medium text-zinc-950 light:text-white bg-zinc-50 light:bg-slate-900 hover:bg-zinc-200 light:hover:bg-slate-800 rounded-lg transition-colors disabled:opacity-50"
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/RunDetailPage.jsx
```diff
@@ -0,0 +1,328 @@
...
+  };
+  const statusLabel =
+    statusLabels[run.status] || run.status?.replace("_", " ") || "";
+  return (
+    <div className="w-full flex items-end justify-between gap-x-4 pb-6 border-white/10 light:border-zinc-300 border-b-2">
+      <div className="flex flex-col gap-y-2">
+        <button
+          type="button"
+          onClick={onBack}
+          className="flex items-center gap-2 text-zinc-400 light:text-slate-600 hover:text-zinc-50 light:hover:text-slate-950 text-sm transition-colors w-fit"
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/RunDetailPage.jsx
```diff
@@ -0,0 +1,328 @@
...
+            </span>
+          )}
+        </div>
+      </div>
+      {run.status === "completed" && (
+        <button
+          type="button"
+          onClick={onContinueInThread}
+          disabled={continuing}
+          className="h-9 px-5 rounded-lg bg-zinc-50 text-zinc-950 light:bg-slate-900 light:text-white text-sm font-medium hover:bg-zinc-200 light:hover:bg-slate-800 transition-colors disabled:opacity-50 shrink-0"
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/RunHistoryPage.jsx
```diff
@@ -0,0 +1,132 @@
...
+      <div
+        style={{ height: isMobile ? "100%" : "calc(100% - 32px)" }}
+        className="relative md:ml-[2px] md:mr-[16px] md:my-[16px] md:rounded-[16px] bg-theme-bg-secondary w-full h-full overflow-y-scroll p-4 md:p-0"
+      >
+        <div className="flex flex-col w-full px-1 md:pl-6 md:pr-[50px] md:py-6 py-16">
+          <div className="w-full flex flex-col gap-y-2 pb-6 border-white/10 light:border-slate-300 border-b-2">
+            <button
+              type="button"
+              onClick={() => navigate(paths.settings.scheduledJobs())}
+              className="flex items-center gap-2 text-zinc-400 light:text-slate-600 hover:text-zinc-50 light:hover:text-slate-950 text-sm transition-colors w-fit"
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/RunHistoryPage.jsx
```diff
@@ -0,0 +1,132 @@
...
+                    <p className="text-sm font-medium text-zinc-400 light:text-slate-600">
+                      {t("scheduledJobs.runHistory.emptySubtitle")}
+                    </p>
+                  </div>
+                  <button
+                    type="button"
+                    onClick={handleRunNow}
+                    disabled={triggering}
+                    className="h-9 px-5 rounded-lg bg-zinc-50 text-zinc-950 light:bg-slate-900 light:text-white text-sm font-medium hover:bg-zinc-200 light:hover:bg-slate-800 transition-colors disabled:opacity-50"
+                  >
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
```diff
@@ -0,0 +1,257 @@
...
+                <div key={group.rootId}>
+                  <button
+                    type="button"
+                    onClick={() => toggleGroup(group)}
+                    className={`flex items-center gap-2 w-full px-2 py-1.5 rounded-md text-sm text-zinc-50 light:text-slate-700 hover:bg-zinc-700/60 light:hover:bg-slate-100 transition-colors ${
+                      !group.standalone
+                        ? "bg-zinc-700 light:bg-slate-200 font-medium"
+                        : ""
+                    }`}
+                  >
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
```diff
@@ -0,0 +1,257 @@
...
+                  </button>
+
+                  {!group.standalone &&
+                    group.subTools.map((sub) => (
+                      <button
+                        key={sub.id}
+                        type="button"
+                        onClick={() => toggleSubTool(sub.id)}
+                        className="flex items-center gap-2 w-full pl-5 pr-2 py-1.5 rounded-md text-sm text-zinc-300 light:text-slate-600 hover:bg-zinc-700/60 light:hover:bg-slate-100 transition-colors"
+                      >
```
shatfield4: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
```diff
@@ -0,0 +1,257 @@
...
+              key={id}
+              className="bg-zinc-800 light:bg-slate-200 flex gap-1.5 h-[26px] items-center justify-center px-3.5 py-0.5 rounded-full text-sm text-zinc-300 light:text-slate-700"
+            >
+              <span className="whitespace-nowrap">{labelFor(id)}</span>
+              <button
+                type="button"
+                onClick={() => removeTool(id)}
+                aria-label={`Remove ${labelFor(id)}`}
+                className="text-zinc-400 light:text-slate-500 hover:text-zinc-50 light:hover:text-slate-900 transition-colors"
+              >
```
shatfield4, line 343: Every button needs `border-none` class for desktop.

frontend/src/pages/GeneralSettings/ScheduledJobs/CronBuilder.jsx
```diff
@@ -0,0 +1,199 @@
...
+    { value: 4, label: t("scheduledJobs.builder.weekdays.thu") },
+    { value: 5, label: t("scheduledJobs.builder.weekdays.fri") },
+    { value: 6, label: t("scheduledJobs.builder.weekdays.sat") },
+  ];
+  const [state, setState] = useState(
+    () => parseCronToBuilderState(value).state
+  );
+  const [wasFallback, setWasFallback] = useState(
+    () => parseCronToBuilderState(value).wasFallback
+  );
```
shatfield4: Let's do this instead to avoid calling `parseCronToBuilderState()` twice:

```
const initial = parseCronToBuilderState(value);
const [state, setState] = useState(initial.state);
const [wasFallback, setWasFallback] = useState(initial.wasFallback);
```

shatfield4 (changes requested):
Found a couple of bugs, 1 UI (tools menu is not scrollable when creating new job) and 1 UX (not handling in progress jobs when user deletes job from UI). The rest are just small nits/clarifications.

Just be sure that whenever adding new `<button>` elements we always use `border-none` class on it unless the designs specify adding a border. We need to do this for merging with desktop or else the default electron styles will add borders to any button element.

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
angelplusultra: It is scrollable. Can you elaborate?

angelplusultra: https://github.com/user-attachments/assets/b40c3b63-c731-49ec-a33e-a591a835d594

server/utils/BackgroundWorkers/index.js
angelplusultra: When a cron timer fires, it doesnt spawn the worker and execute the job. It *enqueues* the job with `this.enqueueScheduledJob(job.id)`. Heres what that method looks like under the hood.




```js
 /**
   * Enqueue a scheduled job for execution. Called by both the cron timer
   * (in addScheduledJob) and the manual trigger endpoint. ScheduledJobRun.start()
   * atomically rejects the call if the job already has a run in flight.
   *
   * @param {number} jobId - scheduled_jobs.id
   */
  async enqueueScheduledJob(jobId) {
    const { ScheduledJobRun } = require("../../models/scheduledJobRun");

    const run = await ScheduledJobRun.start(jobId);
    // if start returns null, skip enqueuing, schueduled job already has a run in flight
    if (!run) return;

    this.#scheduledJobQueue.add(() =>
      this.#runScheduledJobWorker(jobId, run.id).catch(async (err) => {
        this.#log(`Scheduled job ${jobId} failed: ${err.message}`);
        await ScheduledJobRun.failIfNotTerminal(run.id, err.message);
      })
    );
  }
```


We are using the `p-queue` library for the job queue in ` this.#scheduledJobQueue` which internally handles concurrency. so the `add()` method adds the job to the queue and `p-queue` processes them by the specified concurrency limit.

You can see it instantiated with concurrency limits here:
https://github.com/Mintplex-Labs/anything-llm/pull/5322/changes/BASE..d235251d68bd0dc92beb5581252c17934e60716f#diff-6af4d297ce89c04a9bb5e261f531da36704479dbc097b4429050c80e81506fe4R14-R16

frontend/src/locales/en/common.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/7af970d2e7f49566a583f2e56ff6486e5e6c32bc

server/endpoints/scheduledJobs.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/5439cda75b267a470841a917328f34004e42d209

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/RunHistoryPage.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/NewJobModal.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/NewJobModal.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/NewJobModal.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/index.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/index.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/CronBuilder.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/components/ToolCallCard.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/components/RunRow.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/components/JobRow.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/components/GeneratedFileCard.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/components/CollapsibleSection.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/NewJobModal.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/RunDetailPage.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/RunDetailPage.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

frontend/src/pages/GeneralSettings/ScheduledJobs/RunHistoryPage.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/9d4724bb10799909c9dc684ceee89a093826b1cf

server/endpoints/scheduledJobs.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/124bd2ab5d9594e3da6ab1a46f4faab63f266b5f

server/endpoints/scheduledJobs.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/d26c5d8b2ccdf8e16a2cbf7c79b95fb6fd29f10c

server/endpoints/scheduledJobs.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/d26c5d8b2ccdf8e16a2cbf7c79b95fb6fd29f10c

server/endpoints/scheduledJobs.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/d26c5d8b2ccdf8e16a2cbf7c79b95fb6fd29f10c

server/endpoints/scheduledJobs.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/d26c5d8b2ccdf8e16a2cbf7c79b95fb6fd29f10c

server/utils/BackgroundWorkers/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/486702f4c5a0477f203a18268638bbdf6dfc9c4f

server/models/scheduledJob.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/486702f4c5a0477f203a18268638bbdf6dfc9c4f

frontend/src/pages/GeneralSettings/ScheduledJobs/CronBuilder.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/c1d9d86c223a914fc45b6e6d7836ee4987099ae1

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
shatfield4: https://github.com/user-attachments/assets/cb13d5d2-c350-4e32-bdae-a64dedf2b2f5

Here's what I am running into. Still happening in the latest commit. Just happens when I am creating a new scheduled job and try to scroll down to all the tools in the menu.

shatfield4 (changes requested):
Added video for UI bug I was running into when scrolling the tools modal here: https://github.com/Mintplex-Labs/anything-llm/pull/5322#discussion_r3120934024


Once that is fixed, LGTM.

frontend/src/pages/GeneralSettings/ScheduledJobs/JobFormModal/ToolsSelector.jsx
angelplusultra: Hmmm I'm not entirely sure whats causing it not to scroll for you, I think maybe you don't have enough tools in the tools popover to cause a vertical overflow. I tried reproducing it but It still scrolls fine for me.

https://github.com/user-attachments/assets/959bcccd-9bf5-436e-9fd4-55039a587022

However, I do think that when the popover clips the screen height, the popover should probably vertically flip. 

See it here:
https://github.com/Mintplex-Labs/anything-llm/pull/5322/commits/184cb02204a57c71cba6b72b28ce6c394d4f2cd2

shatfield4 (approved):
LGTM

frontend/src/models/scheduledJobs.js
```diff
@@ -0,0 +1,147 @@
...
+  },
+
+  delete: async function (id) {
+    return await fetch(`${API_BASE}/scheduled-jobs/${id}`, {
+      method: "DELETE",
+      headers: baseHeaders(),
+    })
+      .then((res) => res.json())
+      .catch((e) => {
+        console.error(e);
```
timothycarambat: We can remove the console logs for error here - will see in JS console anyway

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,465 @@
...
+// the codebase returns the same instance that `server/index.js` booted. We
+// grab that reference once and reuse it across handlers.
+const backgroundService = new BackgroundService();
+
+function scheduledJobEndpoints(app) {
+  if (!app) return;
+
+  // List available tools for job configuration
+  app.get(
+    "/scheduled-jobs/available-tools",
```
timothycarambat, line 19: The logic in this endpoint for loading tools should be a method on the ScheduledJobs model.

Additionally, we should show ALL TOOLS even if they are not currently enabled. As a user I found it **very** annoying that i had to go back and forth to agent skills to enable a tool just to pick it.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,465 @@
...
+      } catch (e) {
+        console.error(e.message, e);
+        response.sendStatus(500);
+      }
+    }
+  );
+
+  // Mark a run as read
+  app.post(
+    "/scheduled-jobs/runs/:runId/read",
```
timothycarambat: You can combined `read` and `continue` into a single endpoint with `:action` and then just do the logic in the single route and throw on invalid action.

To keep it simple `continue` logic should be moved mostly to a ScheduledJobRun method for keeping logic out of the route.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,465 @@
...
+          return response.status(404).json({ error: "Job not found" });
+        }
+
+        const result = safeJsonParse(run.result, {});
+        const responseText = result?.text || "No response was generated.";
+
+        // Get or create the "Scheduled Jobs" workspace
+        let workspace = await Workspace.get({ slug: "scheduled-jobs" });
+        if (!workspace) {
+          const wsResult = await Workspace.new("Scheduled Jobs");
```
timothycarambat: Lets make a method on workspace for `_upsert`ing a new workspace - about time we need it anyway.

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,465 @@
...
+          workspace = wsResult.workspace;
+          if (!workspace) {
+            return response
+              .status(500)
+              .json({ error: "Failed to create workspace" });
+          }
+        }
+
+        const threadName = `${job.name} - ${new Date(run.startedAt).toISOString()}`;
```
timothycarambat: I think just making this had the default name is fine - this is not very "human" readable anyway!

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,465 @@
...
+              { jobId: job.id },
+              1,
+              { startedAt: "desc" }
+            );
+            return {
+              ...job,
+              latestRun: latestRun || null,
+            };
+          })
+        );
```
timothycarambat: This def works but just...feels unoptimized and/or wrong? Not sure about this. It does not have any major issues or anything, just doesnt sit right

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,465 @@
...
+          prompt: prompt.trim(),
+          tools: tools || null,
+          schedule: schedule.trim(),
+        });
+
+        if (error) {
+          return response.status(400).json({ job: null, error });
+        }
+
+        backgroundService.addScheduledJob(job);
```
timothycarambat, line 186: nice

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,465 @@
...
+        const { job, error } = await ScheduledJob.update(
+          Number(request.params.id),
+          updates
+        );
+
+        if (error) {
+          return response.status(400).json({ job: null, error });
+        }
+
+        await backgroundService.syncScheduledJob(job.id);
```
timothycarambat, line 264: 👍

server/endpoints/scheduledJobs.js
```diff
@@ -0,0 +1,465 @@
...
+
+  // List runs for a job
+  app.get(
+    "/scheduled-jobs/:id/runs",
+    [validatedRequest, isSingleUserMode],
+    async (request, response) => {
+      try {
+        const runs = await ScheduledJobRun.where(
+          { jobId: Number(request.params.id) },
+          50,
```
timothycarambat, line 364: Sensible, I wonder if we should support pagination?

server/jobs/run-scheduled-job.js
```diff
@@ -0,0 +1,212 @@
...
+    if (!transitioned) {
+      log(`Scheduled job run ${runId} was no longer queued, skipping`);
+      conclude();
+      return;
+    }
+
+    log(`Starting scheduled job: "${job.name}" (id=${job.id})`);
+
+    // Update timestamps now so nextRunAt advances even if the job fails
+    await ScheduledJob.updateRunTimestamps(job.id);
```
timothycarambat, line 58: You can move this to `finally` block so it always executes? Obviously will need some var outside the scope to make sure job actually existed and wasnt deleted.

server/jobs/run-scheduled-job.js
```diff
@@ -0,0 +1,212 @@
...
+    const { Workspace } = require("../models/workspace.js");
+    let workspace = await Workspace.get({ slug: "scheduled-jobs" });
+    if (!workspace) {
+      const result = await Workspace.new("Scheduled Jobs");
+      workspace = result.workspace;
+      if (!workspace) {
+        throw new Error("Failed to create Scheduled Jobs workspace");
+      }
+      log("Created dedicated 'Scheduled Jobs' workspace");
+    }
```
timothycarambat: ? The runner needs a workspace?

server/jobs/run-scheduled-job.js
```diff
@@ -0,0 +1,212 @@
...
+
+    // Build trace-capturing handler
+    const thoughts = [];
+    const toolCalls = [];
+    const generatedFiles = [];
+    let textResponse = "";
+    let metrics = {};
+
+    const handler = {
```
timothycarambat: Very smart, should be move to a local util or some job-level util for `scheduled-job-helpers` or something

server/jobs/run-scheduled-job.js
```diff
@@ -0,0 +1,212 @@
...
+      prompt: job.prompt,
+      userId: null,
+      threadId: null,
+    }).init();
+
+    // Tool overrides control which tools the agent can use:
+    // - Array with items: only those specific tools are loaded
+    // - Empty array: no tools are loaded
+    // - null: all enabled agent skills are loaded (default behavior)
+    const toolOverrides = safeJsonParse(job.tools, null);
```
timothycarambat: is the `null` case able to be set on frontend? If not - then we should just assume failover state is no-tools

server/jobs/run-scheduled-job.js
```diff
@@ -0,0 +1,212 @@
...
+    await agentHandler.createAIbitat({
+      handler,
+      toolOverrides,
+    });
+
+    // Auto-approve all tool invocations
+    agentHandler.aibitat.requestToolApproval = async () => ({
+      approved: true,
+      message: "Auto-approved by scheduled job runner.",
+    });
```
timothycarambat: :+1

server/jobs/run-scheduled-job.js
```diff
@@ -0,0 +1,212 @@
...
+        duration,
+      },
+    });
+
+    log(
+      `Job "${job.name}" completed in ${duration}ms (${textResponse.length} chars)`
+    );
+
+    // Send push notification — must load subscriptions since this is a child process
+    try {
```
timothycarambat: Love this, but move to util.

Also, WDYT about having the model also generate the notification title/body?

server/models/scheduledJob.js
timothycarambat: Correct, I did say that

server/utils/BackgroundWorkers/index.js
```diff
@@ -90,15 +112,23 @@ class BackgroundService {
...
+    await this.#bootScheduledJobs();
   }
 
   async stop() {
     this.#log("Stopping...");
+    for (const [id, timer] of this.#scheduledJobTimers) {
+      timer.clear();
+      this.#scheduledJobTimers.delete(id);
+    }
+    this.#scheduledJobQueue.clear();
```
timothycarambat, line 137: lets move this to private method on BGworker class

timothycarambat:
Note: There is a know edge case bug where if a job has NO TOOLS and a completion is run - no metrics are reported at all.

PR #5324, opened 2026-04-02
Model router

### LINKED ISSUE #5313: [FEAT]: Model router

- Build an intelligent prompt routing layer that intercepts user prompts before they reach LLM to dynamically select the model that would provide the best result for the costs
- Main idea is to use calculated (deterministic) rules such as: keyword triggers, token count, message counts, date/time periods (some models have peak pricing times) etc
- See if we can get semantic routing working by allowing users to configure a specific model used for routing chats inside admin settings for model router and using that model to do a tool call to route to the correct model each time
- Follow backend design per internal spec on Notion

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5313
Translations PR: #5666
Docs PR: https://github.com/Mintplex-Labs/anythingllm-docs/pull/245

### Description

- Add model router as a new provider type that routes messages to different LLM providers/models based on configurable rules               
- Support two rule types: calculated (deterministic conditions) and LLM-classified (semantic matching via fallback model)
- Route based on prompt content, conversation token count, message count, time of day, or image attachments                                
- CSV support for keyword matching ("python,javascript,coding" matches any)                                                           
- Configurable cooldown to cache routing decisions and avoid reevaluation on every message (decided to keep this so that users can dial it in to work for their specific workflows)
- Full settings UI for creating/editing routers and rules with drag-and-drop priority ordering                                             
- Inline provider credential setup when configuring routing targets                                                                        
- Works in both regular chat and agent mode, including per-turn re-routing in agent mode                                                   
- Ephemeral routing notification shown in chat during streaming with animated border, disappears on refresh                                
- Router selectable at system level or per-workspace via the model picker  

#### Model routing animation
https://github.com/user-attachments/assets/d3b0a25b-d0cb-4802-9615-593e3eb6e917

#### Chat UI with model router switching to different models
Screenshot 2026-04-08 at 4 06 39 PM

#### Agent mode chat UI with router switching to different models
Screenshot 2026-04-08 at 4 11 37 PM

#### Model router settings page
Screenshot 2026-04-08 at 4 07 00 PM

#### Editing router
Screenshot 2026-04-08 at 4 07 28 PM

#### Router rules configuration
Screenshot 2026-04-08 at 4 07 17 PM

### Developer Validations


frontend/src/pages/GeneralSettings/LLMPreference/index.jsx
```diff
@@ -85,6 +86,15 @@ import { CaretUpDown, MagnifyingGlass, X } from "@phosphor-icons/react";
 import CTAButton from "@/components/lib/CTAButton";
 
 export const AVAILABLE_LLM_PROVIDERS = [
+  {
+    name: "Model Router",
+    value: "anythingllm-router",
+    logo: AnythingLLMIcon,
+    options: (settings) => <ModelRouterOptions settings={settings} />,
+    description:
+      "Route messages to different LLM providers based on rules you define.",
```
angelplusultra, line 97: i18n

frontend/src/pages/GeneralSettings/LLMPreference/index.jsx
```diff
@@ -85,6 +86,15 @@ import { CaretUpDown, MagnifyingGlass, X } from "@phosphor-icons/react";
 import CTAButton from "@/components/lib/CTAButton";
 
 export const AVAILABLE_LLM_PROVIDERS = [
+  {
+    name: "Model Router",
```
angelplusultra, line 92: i18n?

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+            workspaceCount,
+          });
+        }
+        response.status(200).json({ routers: results });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
+      }
+    }
+  );
```
angelplusultra, line 27: Right now this code hits the database once to get all the routers, then loops through each one and hits the database twice more, once for rules and once for workspace count. You can get all of that in a single query using Prisma's `_coun`t:

You could put this in a model method.

```js
try {
    const routers = await prisma.model_routers.findMany({
      orderBy: { createdAt: "asc" },
      include: {
        _count: {
          select: {
            rules: true,
            workspaces: true,
          },
        },
      },
    });

    const results = routers.map(({ _count, ...router }) => ({
      ...router,
      ruleCount: _count.rules,
      workspaceCount: _count.workspaces,
    }));

    response.status(200).json({ routers: results });
  } catch (e) {
    console.error(e);
    response.sendStatus(500).end();
  }
```

server/prisma/schema.prisma
```diff
@@ -394,3 +396,40 @@ model external_communication_connectors {
...
 }
+
+model model_routers {
+  id                Int                  @id @default(autoincrement())
+  name              String               @unique
+  description       String?
+  fallback_provider String
+  fallback_model    String
+  cooldown_seconds  Int                  @default(30)
+  created_by        Int?
```
angelplusultra, line 455: is this a relation to `user` ? Could we call it user and add a relationship to user?

server/models/modelRouter.js
```diff
@@ -0,0 +1,201 @@
...
+        },
+      });
+      return { router, message: null };
+    } catch (error) {
+      console.error(error.message);
+      if (error.message.includes("Unique constraint"))
+        return {
+          router: null,
+          message: "A router with that name already exists.",
+        };
```
angelplusultra: Checking a string match for this contraint error is brittle, prisma offers ways to check for this:

```js
const { Prisma } = require("@prisma/client");

// ...
} catch (error) {
  console.error(error.message);
  if (
    error instanceof Prisma.PrismaClientKnownRequestError &&
    error.code === "P2002"
  ) {
    return {
      router: null,
      message: "A router with that name already exists.",
    };
  }
  return { router: null, message: error.message };
}
```

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+        const { id } = request.params;
+        const router = await ModelRouter.getWithRules({ id: Number(id) });
+        if (!router) {
+          response
+            .status(404)
+            .json({ router: null, error: "Router not found." });
+          return;
+        }
+
+        const workspaceCount = await ModelRouter.workspaceCount(router.id);
```
angelplusultra: These 2 calls get paired a lot. You could execute both operations in one db call with a clever prisma query and just put it inside one model method

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+        response.status(200).json({ router: { ...router, workspaceCount } });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
```
angelplusultra: no need to call `end()` after `sendStatus()`. `sendStatus` ends the response.

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+          results.push({
+            ...router,
+            ruleCount: rules.length,
+            workspaceCount,
+          });
+        }
+        response.status(200).json({ routers: results });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
```
angelplusultra: no need to call `end()` after `sendStatus()`. `sendStatus` ends the response.

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+        );
+        response.status(200).json({ router, error: message });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
+      }
+    }
+  );
+
+  app.post(
```
angelplusultra: update operation would be a `put` or `patch`

server/models/modelRouter.js
```diff
@@ -0,0 +1,201 @@
...
+        ...(orderBy !== null ? { orderBy } : { orderBy: { createdAt: "asc" } }),
+      });
+      return results;
+    } catch (error) {
+      console.error(error.message);
+      return [];
+    }
+  },
+
+  update: async function (id = null, data = {}) {
```
angelplusultra, line 174: can the error field on the return object be `error` instead of `message` to be more clear?

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+    "/admin/model-routers/:id",
+    [validatedRequest, flexUserRoleValid([ROLES.admin])],
+    async (request, response) => {
+      try {
+        const { id } = request.params;
+        const data = reqBody(request);
+        const { router, message } = await ModelRouter.update(Number(id), data);
+        response.status(200).json({ router, error: message });
```
angelplusultra: This should prob be broken into two responses. If an error happened while trying to update it shouldn't return a 200 response.

something like
```js
 const { router, error } = await ModelRouter.update(Number(id), data);

if(error) {
    response.status(400).json({ router, error });
}

File: server/endpoints/modelRouter.js [hunk 12]
```diff
@@ -0,0 +1,193 @@
...
+    "/admin/model-routers/:id",
+    [validatedRequest, flexUserRoleValid([ROLES.admin])],
+    async (request, response) => {
+      try {
+        const { id } = request.params;
+        const success = await ModelRouter.delete(Number(id));
+        response.status(200).json({
+          success,
+          error: success ? null : "Failed to delete router.",
+        });
```
angelplusultra: Same thing here, errors shouldnt respond with a 200. break out into two response paths.

server/models/modelRouter.js
```diff
@@ -0,0 +1,201 @@
...
+
+  delete: async function (id = null) {
+    if (!id) return false;
+
+    try {
+      // Null out any workspace references before deleting
+      await prisma.workspaces.updateMany({
+        where: { router_id: Number(id) },
+        data: { router_id: null },
+      });
```
angelplusultra: pretty sure this is redundant and can be removed, schema already handles this when you delete a model router a workspace has a relation to.

server/endpoints/modelRouter.js
angelplusultra, line 1: Question:

I notice all routes are prefixed with `/admin` and the file that mounts all `/admin` endpoints is in `endpoints/admin.js` Should this be moved there for consistency? Maybe these ednpoints shouldn't even be prefixed with this parent route at all?

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+        const data = reqBody(request);
+        const { rule, message } = await ModelRouterRule.create(
+          Number(id),
+          data,
+          user?.id || null
+        );
+        response.status(200).json({ rule, error: message });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
```
angelplusultra: no need to call end() after sendStatus(). sendStatus ends the response.

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+      try {
+        const { id } = request.params;
+        const user = await userFromSession(request, response);
+        const data = reqBody(request);
+        const { rule, message } = await ModelRouterRule.create(
+          Number(id),
+          data,
+          user?.id || null
+        );
+        response.status(200).json({ rule, error: message });
```
angelplusultra: Same thing here, errors shouldnt respond with a 200. break out into two response paths.

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+  app.post(
+    "/admin/model-routers/:id/rules/reorder",
+    [validatedRequest, flexUserRoleValid([ROLES.admin])],
+    async (request, response) => {
+      try {
+        const { ruleUpdates } = reqBody(request);
+        if (!Array.isArray(ruleUpdates)) {
+          response
+            .status(200)
+            .json({ success: false, error: "ruleUpdates must be an array." });
```
angelplusultra, line 148: this should be an error code

server/models/modelRouterRule.js
```diff
@@ -0,0 +1,279 @@
...
+  "between",
+];
+const TITLE_REGEX = /^[a-z0-9_]+$/;
+
+const ModelRouterRule = {
+  VALID_TYPES,
+  VALID_PROPERTIES,
+  VALID_COMPARATORS,
+
+  create: async function (routerId, data = {}, creatorId = null) {
```
angelplusultra, line 33: can the error field on the return object be `error` instead of `message` to be more clear?

server/models/modelRouterRule.js
```diff
@@ -0,0 +1,279 @@
...
+            message: "A rule with that title already exists on this router.",
+          };
+        if (error.message.includes("priority"))
+          return {
+            rule: null,
+            message: "A rule with that priority already exists on this router.",
+          };
+        return { rule: null, message: "Duplicate rule constraint violated." };
+      }
+      return { rule: null, message: error.message };
```
angelplusultra: There's more Prisma-y ways to catch these specific error types that are better than string matching. You can look them up.
e.g

```js
const { PrismaClientKnownRequestError } = require("@prisma/client/runtime/library");

// a unique constraint was violated
if (e instanceof PrismaClientKnownRequestError && e.code === "P2002") {
  const field = e.meta?.target; // e.g. ["title"] or ["priority"]

  if (field?.includes("title"))
    return { rule: null, message: "A rule with that title already exists on this router." };
  if (field?.includes("priority"))
    return { rule: null, message: "A rule with that priority already exists on this router." };

  return { rule: null, message: "Duplicate rule constraint violated." };
}
```

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+    async (request, response) => {
+      try {
+        const { ruleId } = request.params;
+        const success = await ModelRouterRule.delete(Number(ruleId));
+        response
+          .status(200)
+          .json({ success, error: success ? null : "Failed to delete rule." });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
```
angelplusultra: no need to call end() after sendStatus(). sendStatus ends the response.

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+        const { ruleId } = request.params;
+        const data = reqBody(request);
+        const { rule, message } = await ModelRouterRule.update(
+          Number(ruleId),
+          data
+        );
+        response.status(200).json({ rule, error: message });
+      } catch (e) {
+        console.error(e);
+        response.sendStatus(500).end();
```
angelplusultra: no need to call end() after sendStatus(). sendStatus ends the response.

server/endpoints/modelRouter.js
```diff
@@ -0,0 +1,193 @@
...
+        const { ruleUpdates } = reqBody(request);
+        if (!Array.isArray(ruleUpdates)) {
+          response
+            .status(200)
+            .json({ success: false, error: "ruleUpdates must be an array." });
+          return;
+        }
+        const { success, message } =
+          await ModelRouterRule.reorderRules(ruleUpdates);
+        response.status(200).json({ success, error: message });
```
angelplusultra: Same thing here, errors shouldnt respond with a 200. break out into two response paths.

server/models/modelRouterRule.js
```diff
@@ -0,0 +1,279 @@
...
+    }
+  },
```
angelplusultra, line 208: Same here, could check for official prisma constraint errors.

server/models/modelRouterRule.js
```diff
@@ -0,0 +1,279 @@
...
+
+  /**
+   * Bulk update priorities for rules within a router.
+   * @param {Array<{id: number, priority: number}>} ruleUpdates
+   */
+  reorderRules: async function (ruleUpdates = []) {
+    try {
+      await prisma.$transaction(
```
angelplusultra, line 229: good idea with the transaction here 👍

server/models/modelRouterRule.js
```diff
@@ -0,0 +1,279 @@
...
+        orderBy: { priority: "asc" },
+      });
+      return rules;
+    } catch (error) {
+      console.error(error.message);
+      return [];
+    }
+  },
+
+  update: async function (id = null, data = {}) {
```
angelplusultra, line 142: this update handler is pretty busy, I see a lot of repetition that could be refactored to be less loc.

An example:

```js
const VALID_PROPERTIES = [
  "promptContent",
  "conversationTokenCount",
  "conversationMessageCount",
  "currentHour",
  "hasImageAttachment",
];
const VALID_COMPARATORS = [
  "contains",
  "gt",
  "gte",
  "lt",
  "lte",
  "eq",
  "neq",
  "between",
];
const TITLE_REGEX = /^[a-z0-9_]+$/;

function assignEnum(updates, data, key, validList, label) {
  if (data[key] === undefined) return null;
  if (!validList.includes(data[key]))
    return {
      rule: null,
      message: `${label} must be one of: ${validList.join(", ")}`,
    };
  updates[key] = data[key];
  return null;
}

function assignEnumOrNull(updates, data, key, validList, label) {
  if (data[key] === undefined) return null;
  const v = data[key];
  if (v && !validList.includes(v))
    return {
      rule: null,
      message: `${label} must be one of: ${validList.join(", ")}`,
    };
  updates[key] = v || null;
  return null;
}
```


```js
  update: async function (id = null, data = {}) {
    if (!id) throw new Error("No rule id provided for update");

    const updates = {};

    if (data.title !== undefined) {
      const title = this._validateTitle(data.title);
      if (!title)
        return {
          rule: null,
          message: "Title must be lowercase with underscores only.",
        };
      updates.title = title;
    }

    const simpleFields = [
      ["enabled", (v) => Boolean(v)],
      ["priority", (v) => Number(v)],
      ["description", (v) => v || null],
      ["value", (v) => (v != null ? String(v) : null)],
      ["route_provider", (v) => String(v)],
      ["route_model", (v) => String(v)],
    ];
    for (const [key, map] of simpleFields) {
      if (data[key] !== undefined) updates[key] = map(data[key]);
    }

    const enumErr =
      assignEnum(updates, data, "type", VALID_TYPES, "Type") ||
      assignEnumOrNull(
        updates,
        data,
        "property",
        VALID_PROPERTIES,
        "Property"
      ) ||
      assignEnumOrNull(
        updates,
        data,
        "comparator",
        VALID_COMPARATORS,
        "Comparator"
      );
    if (enumErr) return enumErr;

    if (Object.keys(updates).length === 0)
      return { rule: { id }, message: "No valid fields to update." };

    try {
      const rule = await prisma.model_router_rules.update({
        where: { id: Number(id) },
        data: updates,
      });
      return { rule, message: null };
    } catch (error) {
      console.error(error.message);
      if (error.message.includes("Unique constraint")) {
        if (error.message.includes("title"))
          return {
            rule: null,
            message: "A rule with that title already exists on this router.",
          };
        if (error.message.includes("priority"))
          return {
            rule: null,
            message: "A rule with that priority already exists on this router.",
          };
        return { rule: null, message: "Duplicate rule constraint violated." };
      }
      return { rule: null, message: error.message };
    }
  },
```

server/utils/chats/stream.js
```diff
@@ -51,10 +51,75 @@ async function streamChatWithWorkspace(
...
-    provider: workspace?.chatProvider,
-    model: workspace?.chatModel,
-  });
+  let LLMConnector;
+  let routingMetadata = null;
+
+  // Determine if we should use the model router. This applies when:
+  // 1. Workspace explicitly sets chatProvider to "anythingllm-router", OR
+  // 2. Workspace uses system default and system LLM is "anythingllm-router"
+  const effectiveProvider = workspace?.chatProvider || process.env.LLM_PROVIDER;
```
angelplusultra: consider `resolvedProvider`. I think that's a clearer name and consistent wording across codebase.

server/utils/AiProviders/modelRouter/index.js
```diff
@@ -0,0 +1,123 @@
...
+    return this.delegateProvider.embedTextInput(textInput);
+  }
+
+  async embedChunks(textChunks) {
+    return this.delegateProvider.embedChunks(textChunks);
+  }
+
+  async compressMessages(promptArgs, rawHistory) {
+    return this.delegateProvider.compressMessages(promptArgs, rawHistory);
+  }
```
angelplusultra, line 167: Are any of these methods ever invoked? In `stream.js`, `LLMConnector` gets assigned to `router.delegateProvider` which is the actual provider instance. These methods are never invoked anywhere? 

Unless im unaware of something, i dont think the `delegateProvider` ref is necesarry. You could just return the provider directly in `resolve()` and adjust `stream.js` to assign `LLMConnector` from there.

server/utils/router/llmClassifier.js
```diff
@@ -0,0 +1,78 @@
...
+  const systemPrompt = CLASSIFICATION_PROMPT.replace(
+    "{categories}",
+    categories
+  );
+
+  try {
+    const provider = getLLMProvider({
+      provider: router.fallback_provider,
+      model: router.fallback_model,
+    });
```
angelplusultra: So the model that actually performs the route evaluation is the routers fallback model. I feel like this is crucial information about this feature that should be surfaced in the UI. When selecting a fallback model, you're also selecting what model is used for route evaluation. As of now, I don't think anything in the UI is making that clear.

server/utils/router/llmClassifier.js
```diff
@@ -0,0 +1,78 @@
...
+
+    const messages = await provider.constructPrompt({
+      systemPrompt,
+      contextTexts: [],
+      chatHistory: [],
+      userPrompt: prompt,
+      attachments: [],
+    });
+
+    const fullText = await collectStreamResponse(provider, messages);
```
angelplusultra: Why does this need to be a stream? Couldn’t this just be a standard completion call?

Streaming feels more like a UX concern for displaying responses incrementally, rather than something we need for an internal step like route classification. Wouldn’t keeping this as a simple completion make the implementation a bit cleaner?

frontend/src/pages/GeneralSettings/ModelRouters/LLMProviderModelPicker/index.jsx
angelplusultra, line 1: As I mentioned in my server-side comment on the LLM classification file, since the fallback provider is being used for route classification, it would be helpful to surface that information to the user.

angelplusultra (changes requested):
Nice work, everything seems to work smoothly.


I dont see this env var updated in the `.env.example`. It should be grouped 

```
# LLM_PROVIDER="anythingllm-router"
# MODEL_ROUTER_ID=1
```

server/utils/router/llmClassifier.js
shatfield4: Streaming functions are much more maintained in our codebase and we know it is working well in production since most users are using streaming. It does make sense here to use getChatCompletion but from a reliability standpoint it makes sense to use the same functions we're using across all providers for streaming.

server/utils/AiProviders/modelRouter/index.js
shatfield4: These are not directly called but just available to match the structure of every other provider. I think it is best to keep this code so that if we call `getLLMProvider()` in the future and expect to be able to call all functions directly from the model router provider, we would not have to create more conditions to catch edge cases. Although it is not used directly now, it makes sense to keep the function signatures consistent across all providers.

server/prisma/schema.prisma
shatfield4: Skipping this since in our schema, all other `createdBy`/`created_by` fields don't have relations to users either so going to keep this consistent here. The reason for making it optional and no relation to users is because model routers can exist in both single and multi-user modes and would be set to null anyways if there is a direct relation. Keeping it the way we have it just keeps the schema simpler.

server/utils/chats/stream.js
shatfield4: This part of the code is just checking the workspace/global LLM provider so effectiveProvider makes sense for the variable name. Keeping as is.

frontend/src/pages/GeneralSettings/LLMPreference/index.jsx
shatfield4: Leaving as is since no other providers we have use translations for name/descriptions.

frontend/src/pages/GeneralSettings/LLMPreference/index.jsx
shatfield4: Leaving as is since no other providers we have use translations for name/descriptions.

frontend/src/pages/GeneralSettings/ModelRouters/LLMProviderModelPicker/index.jsx
shatfield4: Added this to the fallback model description. The only time the model is being used for routing is when using LLM classified rules so this should be okay to just have it in the description only when they choose the fallback model.

server/models/modelRouter.js
```diff
@@ -96,6 +101,28 @@ const ModelRouter = {
...
+      if (!router) return null;
+      const { _count, ...rest } = router;
+      return { ...rest, workspaceCount: _count.workspaces };
+    } catch (error) {
+      console.error(error.message);
+      return null;
+    }
+  },
+
```
angelplusultra, line 129: 👍️

server/models/modelRouter.js
```diff
@@ -110,13 +137,37 @@ const ModelRouter = {
...
+        ...router,
+        ruleCount: _count.rules,
+        workspaceCount: _count.workspaces,
+      }));
+    } catch (error) {
+      console.error(error.message);
+      return [];
+    }
+  },
+
```
angelplusultra, line 173: 👍️

angelplusultra (approved):
LGTM.

NIT: Not super crucial because I know this is a pre-design implementation but the row on a rule from "{POSITION_NUMBER} {RULE_NAME} {RULE_TYPE}" looks unaligned to me and the rule type is black against dark mode, making it  hard to read. I know it's pre-design but just wanted to point it out.

Image

shatfield4:
- [x] Convert semantic router custom json tool to an aibitat tool

shatfield4:
@angelplusultra just a couple new changes here:
- New webm animation and cleanup dead code for old animation
- Now uses aibitat instead of the custom json tool to classify prompts
- bug fix with the new rule modal not resetting each time we make a new rule.

frontend/src/utils/paths.js
```diff
@@ -148,6 +156,12 @@ export default {
     apiKeys: () => {
       return "/settings/api-keys";
     },
+    modelRouters: () => {
+      return "/settings/model-routers";
+    },
+    modelRouterEdit: (id) => {
```
angelplusultra: This leads to the rules. `modelRouterRules` seems more accurate, editing is now a modal in "/settings/model-routers",

frontend/src/main.jsx
```diff
@@ -255,6 +255,24 @@ const router = createBrowserRouter([
...
+          return { element: <AdminRoute Component={ModelRouters} /> };
+        },
+      },
+      {
+        path: "/settings/model-routers/:id",
+        lazy: async () => {
+          const { default: RouterFormPage } = await import(
+            "@/pages/GeneralSettings/ModelRouters/RouterForm"
+          );
+          return { element: <AdminRoute Component={RouterFormPage} /> };
```
angelplusultra: `RouterFormPage` should be renamed to `RouterRulesPage` or something like that.

frontend/src/pages/GeneralSettings/ModelRouters/RouterRulesPage/index.jsx
angelplusultra, line 1: Should be renamed to `RouterRulesPage/index.jsx` or something to be more clear, this name was from a pre-design implementation.

angelplusultra (changes requested):
Just a few minor nits and one UI suggestion:

I’m not sure having an AND/OR button for every individual condition is the best fit here. Since all conditions are evaluated together using a single logical operator, it feels a bit redundant in this flat structure. I think this pattern would make more sense if we were supporting nested/grouped conditions, but that doesn’t seem to be the case right now. As it stands, this feels like a step back from the previous implementation.

Also, I noticed that the alignment starts to shift when selecting values that include caption text, which makes the layout feel a bit uneven.

Some possible ideas to explore:

* Moving the caption/help text into a tooltip so the row layout stays consistent
* Aligning the AND/OR control more closely with the input row to prevent visual drift
Image

shatfield4:
> Just a few minor nits and one UI suggestion:
> 
> I’m not sure having an AND/OR button for every individual condition is the best fit here. Since all conditions are evaluated together using a single logical operator, it feels a bit redundant in this flat structure. I think this pattern would make more sense if we were supporting nested/grouped conditions, but that doesn’t seem to be the case right now. As it stands, this feels like a step back from the previous implementation.
> 
> Also, I noticed that the alignment starts to shift when selecting values that include caption text, which makes the layout feel a bit uneven.
> 
> Some possible ideas to explore:
> 
> * Moving the caption/help text into a tooltip so the row layout stays consistent
> * Aligning the AND/OR control more closely with the input row to prevent visual drift
> 
> Image

- Keeping and/or since this is how it was designed and it is very clear to the user which operator it is using between rules
- Fixed positioning of and/or block
- Added telemetry to creation of new model routers for tracking

angelplusultra (approved):
Looks much better, last nit for me would just the delete xs to align with the row as well, when there's helper text it gets pushed down.

After that LGTM

Image

angelplusultra (approved):
LGTM

frontend/src/components/SettingsSidebar/MenuOption/index.jsx
```diff
@@ -28,9 +29,11 @@ export default function MenuOption({
 
   const isActive = hasChildren
     ? (!isExpanded &&
-        childOptions.some((child) => child.href === location.pathname)) ||
+        childOptions.some((child) =>
+          isPathMatch(child.href, location.pathname)
+        )) ||
       location.pathname === href
-    : location.pathname === href;
+    : isPathMatch(href, location.pathname);
```
timothycarambat, line 37: What was the reasoning for the modification of this change? Did this sub-menu option highlight system not work prior?

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/LLMSelector/RouterPickerSelection/index.jsx
```diff
@@ -0,0 +1,64 @@
...
+    );
+  }
+
+  if (routers.length === 0) {
+    return (
+      <p className="text-xs text-zinc-400 light:text-slate-500">
+        {t("model-router.router-selection.no-routers-chat")}
+      </p>
+    );
+  }
```
timothycarambat, line 40: Have this link to the settings page in the prompt model dropdown like we do in the workspace chat selection UI as well

PR #5460, opened 2026-04-17
fix(collector): always close Puppeteer browser (Chromium zombie leak)

Alminc91:
## Summary

Close Puppeteer browser via `try/finally` so any error in `page.goto`, `evaluate`, or timeout no longer leaks the Chromium process.

## Context

Observed on 2026-04-17: `intern` container accumulated **14 zombie Chromium processes (~2.5 GiB)** in 8.4 GiB total RAM after an OpenRouter-auth-cascade. `docker restart intern` freed 8 GiB and confirmed every zombie was an orphaned Puppeteer Chromium.

Root cause: `loader.scrape` relied on reaching `await browser.close()` at the end of the function — but any throw in between (page.goto timeout, waitForFunction failure, evaluate exception) would skip the close.

## Changes

- Wrap `scrape()` body in `try/finally`
- Close browser in `finally` with `.catch()` so a close-error doesn't mask the original
- Remove inner `browser.close()` from the options.evaluate callback (the outer finally handles it; the inner call was redundant and masked the leak when it threw)

## Test plan

- [x] Syntax check: `node --check` passes
- [ ] Build AnythingLLM docker image
- [ ] Deploy to intern container first, monitor for 24h — verify no zombies accumulate
- [ ] Roll out to 92 customer containers

## Deployment

**Important**: Fix only takes effect after image rebuild + container rollout. Until then, current containers continue to leak on errors.


Alminc91:
Mispushed — should target our kufer fork, not upstream. Reopening there.

PR #5492, opened 2026-04-23
fix(container+prompt): privilege-drop entrypoint + sharper queryRewriter heuristic

Alminc91:
## Summary
- **Container privilege handling**: Install `gosu`, fix `/app/server/storage` ownership as root, then drop to `anythingllm` user
- **queryRewriter prompt**: Switch from syntactic ("contains pronouns") to semantic ("would a stranger understand?") self-contained check, with "when in doubt, reformulate" tiebreaker

### 1. Container permissions
Old entrypoint only fixed the DB file post-hoc with a writability check, leaving WAL/SHM and other storage files with whatever UID `docker cp` produced. New flow:
1. Container starts as root
2. `chown -R anythingllm:anythingllm $STORAGE_DIR` — fixes everything
3. `gosu anythingllm node ...` — drops to unprivileged user for runtime

Backwards-compatible: legacy non-root containers still hit the old DB writability check path.

### 2. queryRewriter heuristic
Old rule was syntactic and missed many follow-ups that semantically needed context (e.g. self-contained-looking questions in the middle of a thread). New rule asks the practical question: "could a stranger understand this without the chat history?" Adds explicit tiebreaker: favour reformulation over context-loss. Matches the RAG-quality work documented in `docs/ROADMAP.md` (March 2026 entry).

## Test plan
- [ ] Build container, verify `/app/server/storage` ownership is `anythingllm:anythingllm` after start
- [ ] Verify `node` and `collector` processes both run as `anythingllm` (`ps`)
- [ ] Smoke-test chatbot with self-contained question → no reformulation
- [ ] Smoke-test follow-up like "und davon?" → reformulation includes prior subject

## Note
The `anythingllm-chat-widget.min.js` / `.min.css` changes that appeared on the previous local branch are intentionally **not** included — those were a stale local rebuild, the authoritative version is on `master` (deployed directly).


Alminc91:
Wrong base repo — closing, will re-open against our fork (Alminc91/anything-llm-kufer)

PR #5507, opened 2026-04-24
fix: gate generic OpenAI tool stream usage reporting

haimingZZ:
## Summary
- let `tooledStream` omit `stream_options` when a provider explicitly disables them
- make the Generic OpenAI native tool-calling stream path honor `GENERIC_OPEN_AI_REPORT_USAGE` before sending `include_usage`
- add a focused regression test for default usage reporting and provider-disabled stream options

## Tests
- `node -c server/utils/agents/aibitat/providers/helpers/tooled.js`
- `node -c server/utils/agents/aibitat/providers/genericOpenAi.js`
- `npx jest server/__tests__/utils/agents/aibitat/providers/helpers/tooled.test.js --runInBand`
- `git diff --check`

Note: targeted `npx eslint ...` could not run in this checkout because the local install is missing ESLint config dependencies (`@eslint/js`).


timothycarambat:
Are there any downsides to sending `stream_options` in this aspect? Is there a provider which throws when given this param?

PR #5528, opened 2026-04-25
fix: SPA nav for thread/workspace switching

### LINKED ISSUE #5512: [BUG]: Whole page "blinks" when switching between threads

Docker (local)

### What happened?

Whenever I switch between threads, the whole page turns white for a split second. This seems to be because it loads in a new page each time, ie. between `/my-workspace/t/<chat id 1>` and `/my-workspace/t/<chat id 2>`

Open up AnythingLLM in a firefox browser. Click between different threads.
(may yield same results on chrome)

_No response_

_No response_

shatfield4:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5512 

### Description

- Converts from `<a>` to `<Link>` for SPA nav preventing screen flashes from entire page reloads
- Nested thread route under workspace to keep parent route (default/workspace) mounted when accessing thread routes
- `key={threadSlug}` on `WorkspaceChatContainer`, ChatContainer holds chat history in local state, so we add the key to force a fresh mount
- `key={slug}` on `ShowWorkspaceChat` to also force fresh mount
- Move `Sidebar` up to `WorkspaceChat` to prevent the `key={slug}` from forcing a fresh mount on the sidebar since this state only updates when creating new threads/workspaces (which already do a full page refresh)

### Developer Validations


frontend/src/components/Sidebar/ActiveWorkspaces/index.jsx
```diff
@@ -117,12 +117,15 @@ export default function ActiveWorkspaces() {
...
                       <div className="flex gap-x-2 items-center justify-between">
-                        <a
-                          href={
+                        <Link
+                          to={
                             isActive
-                              ? null
+                              ? "#"
                               : paths.workspace.chat(workspace.slug)
                           }
```
angelplusultra: Pretty sure this ternary is not needed, React Router will not reload the page if the current route is active. Could just be 
```js
 to={paths.workspace.chat(workspace.slug)}

```

frontend/src/components/Sidebar/ActiveWorkspaces/index.jsx
```diff
@@ -117,12 +117,15 @@ export default function ActiveWorkspaces() {
...
+                          onClick={(e) => {
+                            if (isActive) e.preventDefault();
+                          }}
```
angelplusultra: Don't think this is needed, I removed it and everything works the same.

frontend/src/components/Sidebar/ActiveWorkspaces/ThreadContainer/ThreadItem/index.jsx
```diff
@@ -96,11 +96,13 @@ export default function ThreadItem({
...
+          <Link
             ref={ref}
-            href={
-              window.location.pathname === linkTo || ctrlPressed ? "#" : linkTo
-            }
+            to={linkTo}
+            onClick={(e) => {
+              if (ctrlPressed || window.location.pathname === linkTo)
+                e.preventDefault();
+            }}
```
angelplusultra: pretty sure this onClick can be removed

angelplusultra (changes requested):
Navigating between threads is definitely snappier now that these links use SPA navigation. That said, the experience still feels quite flickery when switching between threads, which seems to be the core UX issue. In my opinion, the loading state during thread transitions also needs improvement to make navigation feel smoother and more seamless.

https://github.com/user-attachments/assets/24c841a2-b724-4df4-bfcd-efd59e310926

frontend/src/components/Sidebar/ActiveWorkspaces/index.jsx
shatfield4: 1659e9735

frontend/src/components/Sidebar/ActiveWorkspaces/ThreadContainer/ThreadItem/index.jsx
shatfield4: 1659e9735

frontend/src/components/Sidebar/ActiveWorkspaces/index.jsx
shatfield4: https://github.com/Mintplex-Labs/anything-llm/commit/1659e973508d614a5125ed789000871447d74a6b

shatfield4:
@angelplusultra Now stores currently loaded state in a `useState` to avoid passing it as props which trigger the entire rerender of the component. Doing it this way prevents the UI from showing the loading state which is what was causing the very quick screen flash.

angelplusultra (approved):
This feels great. much better UX, in my opinion.

The only hazard I can think of is that UI actions like message submission does not appear to be disabled during thread transitions. 

That creates a potential race condition: if you navigate to another thread, trigger the silent load, and send a message before the remount completes, the message could be submitted against the previously mounted thread and then the UI would switch to the newly selected one.

This is probably unlikely in practice since the transitions are fast, but it still seems worth noting.

If you see this as an issue you can go ahead and submit a fix otherwise LGTM

shatfield4:
Fixed race condition @angelplusultra

Decide to not abort the chat being processed in bg since most platforms like ChatGPT or Claude will continue processing the chat in the bg even if we leave the page.

angelplusultra (approved):
LGTM

PR #5540, opened 2026-04-27
fix: keep chat lookups scoped for null users

haimingZZ:
## Summary
- add a shared helper for workspace chat ownership filters so null-user sessions are scoped as `user_id: null`
- use that helper for workspace and thread chat delete/edit/fork/feedback/TTS lookup paths
- add focused Jest coverage to prevent `undefined` from dropping the user ownership filter, including thread and fork filters

## Why
Some workspace chat routes built Prisma `where` clauses with `user_id: user?.id`. When `user` is missing, that value becomes `undefined`, which can be omitted from the query filter instead of matching only null-user chats. This keeps those routes scoped to the intended null-user chat records.

## Tests
- `yarn test server/__tests__/utils/helpers/chat/userScopedChatClause.test.js --runInBand`
- `node -c server/endpoints/workspaces.js && node -c server/endpoints/workspaceThreads.js && node -c server/utils/helpers/chat/userScopedChatClause.js && node -c server/__tests__/utils/helpers/chat/userScopedChatClause.test.js`
- `git diff --check origin/master...HEAD`

## Local notes
- `yarn --cwd server lint:check --quiet` could not run in this checkout because `server/node_modules` is absent and `eslint` is not installed locally.

### DIFF
--- a/server/__tests__/utils/helpers/chat/userScopedChatClause.test.js
+++ b/server/__tests__/utils/helpers/chat/userScopedChatClause.test.js
@@ -1,0 +1,50 @@
+const {
+  userScopedChatClause,
+} = require("../../../../utils/helpers/chat/userScopedChatClause");
+
+describe("userScopedChatClause", () => {
+  it("uses the signed-in user id when present", () => {
+    expect(userScopedChatClause({ workspaceId: 1 }, { id: 7 })).toEqual({
+      workspaceId: 1,
+      user_id: 7,
+    });
+  });
+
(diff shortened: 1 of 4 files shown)


timothycarambat:
Please stop creating PRs without issues describing the issue or improvement to be done. Some of these things dont need fixing and I dont want you wasting time. Also you **do not** need a test for every single PR, which you seem to generate while leaving other atomic tests out. A file is either fully tested, or not.

PR #5549, opened 2026-04-28
fix: tools popover overflowing screen on small viewports

### LINKED ISSUE #5530: [BUG]: Cannot Scroll to Bottom of Fully Populated Slash Command Entries

Suggest the ability to scroll to the bottom of the fully populated Slash Command Entries, to add them commands via the Add button that used to be there but is no longer visible in my OS. For example, a scroll bar either on the Slash Commands section of Tools to allow users to view all commands or slash command completion when typing "/command" so that possible commands that exist pop up. On my machine the only way to add the Slash Command is via the **_Import Item_** button.

Image

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5530

### Description

The Tools popover (Slash Commands / Agent Skills) used a static viewport-based max-height (`max-h-[min(360px, calc(100dvh - 25rem))]` in centered mode, `... - 11rem` otherwise). On shorter windows — or when the centered home view's heading pushes the input lower — the calculation didn't reflect the popover's actual position, so the popover bottom would render below the viewport. Combined with the inner container's `no-scroll` class hiding the scrollbar, users had no way to reach the bottom entries.

This replaces the static class with a `useLayoutEffect` that measures the popover's `getBoundingClientRect()` and computes the real available space, then applies it via inline `style={{ maxHeight }}`. It recomputes on `resize` and is clamped to `[160px, 360px]`.

- Centered mode (popover opens below input): `window.innerHeight - rect.top - 16`
- Non-centered mode (popover opens above input): `rect.bottom - 16`

The `no-scroll` class is preserved, so the scrollbar stays hidden — users still scroll via mouse wheel / trackpad / arrow keys, but the popover now never extends past the viewport edge.

### Additional Information

No behavior change in normal-sized windows where the static 360px cap was already fitting; the dynamic calc only kicks in when available space is less than 360px.

### Developer Validations


frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/ToolsMenu/index.jsx
```diff
@@ -62,6 +71,24 @@ export default function ToolsMenu({
...
+      const rect = el.getBoundingClientRect();
+      const available = centered
+        ? window.innerHeight - rect.top - 16
+        : rect.bottom - 16;
+      setMaxHeight(Math.max(160, Math.min(360, available)));
+    };
+    update();
+    window.addEventListener("resize", update);
+    return () => window.removeEventListener("resize", update);
+  }, [showing, centered, activeTab]);
```
shatfield4: Nit: remove `activeTab` from dep array. Not needed here.

shatfield4 (changes requested):
Small nit here but if that is fixed then LGTM. I tried to do this using a CSS only approach but was unable to get it to work so this way using `useLayoutEffect` seems to be the only way to get this to resize properly.

Only other note is that the min height of 160 can cause the menu to not resize on really small screen but realistically no user has screen that small.

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/ToolsMenu/index.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5549/commits/9dbb10a1d9646b32a98ee09a955443de7392c92c

angelplusultra:
> Small nit here but if that is fixed then LGTM. I tried to do this using a CSS only approach but was unable to get it to work so this way using `useLayoutEffect` seems to be the only way to get this to resize properly.
> 
> Only other note is that the min height of 160 can cause the menu to not resize on really small screen but realistically no user has screen that small.

https://github.com/Mintplex-Labs/anything-llm/pull/5549/commits/9dbb10a1d9646b32a98ee09a955443de7392c92c

shatfield4 (approved):
LGTM

PR #5584, opened 2026-05-06
fix: parse reasoning output from LM Studio reasoning models

### LINKED ISSUE #5583: [BUG]: LM Studio Provider Does Not Present Reasoning Output

Docker (local)

### What happened?

The LM Studio provider doesn't present a reasoning models thoughts when they are outputted. Models such as `deepseek-r1` and `qwen3` output reasoning content.

1. Set up LM Studio as provider and choose a reasoning model that outputs it's reasoning. 

2. Send a basic prompt (not "@agent")

3. Notice thoughts are not being presented as it would with another provider such as Ollama.

_No response_

_No response_

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5583

### Description

The LM Studio provider was using the generic `handleDefaultStreamResponseV2` stream handler, which only reads `choices[0].delta.content`. LM Studio's OpenAI-compatible endpoint emits reasoning tokens for reasoning models in a separate `reasoning_content` field (same shape as DeepSeek), so those tokens were silently dropped on both the streaming and non-streaming paths and the UI never showed the model's thinking output.

Streaming (`handleStream`) — replaced the default handler with a bespoke implementation modeled after the existing DeepSeek/Foundry handlers:

- Reads both `delta.content` and `delta.reasoning_content` per chunk.
- On the first reasoning chunk, prepends a `<think>` tag and streams subsequent reasoning tokens through verbatim.
- On the first content token after a reasoning run, emits the closing `</think>` tag, flushes the buffered reasoning into `fullText`, and resumes normal content streaming.
- Preserves the existing usage-metric handling, abort handling, and `finish_reason` termination logic from the default handler.

Non-streaming (`getChatCompletion`) — added a `#parseReasoningFromResponse` helper that, when `message.reasoning_content` is present, prepends `<think>{reasoning}</think>` to the response content before returning it as `textResponse`. Same shape the streaming path produces, so downstream rendering is consistent across both modes.

### Developer Validations

### DIFF
--- a/server/utils/AiProviders/lmStudio/index.js
+++ b/server/utils/AiProviders/lmStudio/index.js
@@ -214,6 +214,21 @@ class LMStudioLLM {
     ];
   }
 
+  /**
+   * Parses and prepends reasoning from the response and returns the full text response.
+   * @param {Object} response
+   * @returns {string}
+   */
+  #parseReasoningFromResponse({ message }) {
+    let textResponse = message?.content;
+    if (
+      !!message?.reasoning_content &&
+      message.reasoning_content.trim().length > 0
+    )
+      textResponse = `<think>${message.reasoning_content}</think>${textResponse}`;
+    return textResponse;
+  }
+
   async getChatCompletion(messages = null, { temperature = 0.7 }) {
     if (!this.model)
       throw new Error(
@@ -235,7 +250,7 @@ class LMStudioLLM {
       return null;
 
     return {
-      textResponse: result.output.choices[0].message.content,
+      textResponse: this.#parseReasoningFromResponse(result.output.choices[0]),
       metrics: {
         prompt_tokens: result.output.usage?.prompt_tokens || 0,
         completion_tokens: result.output.usage?.completion_tokens || 0,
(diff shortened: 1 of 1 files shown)


shatfield4 (approved):
Tested using `deepseek/deepseek-r1-0528-qwen3-8b` and everything worked as it should. LGTM.

RailKill:
should the `reasoning_content` implementations be moved to `handleDefaultStreamResponseV2` so that it covers other providers? i think it should be fine since there's if-checks and guards. the GenericOpenAI provider is still using the old code and has TODO comments to use a shared function. that's the point of `handleDefaultStreamResponseV2` right?

i'm experiencing this problem with the LocalAI provider, and i'm sure Ollama and other providers have the same problem because they all use OpenAI-compatible endpoints, requiring the same fix. i think this would be a good opportunity to do that

angelplusultra:
> should the `reasoning_content` implementations be moved to `handleDefaultStreamResponseV2` so that it covers other providers? i think it should be fine since there's if-checks and guards. the GenericOpenAI provider is still using the old code and has TODO comments to use a shared function. that's the point of `handleDefaultStreamResponseV2` right?
> 
> i'm experiencing this problem with the LocalAI provider, and i'm sure Ollama and other providers have the same problem because they all use OpenAI-compatible endpoints, requiring the same fix. i think this would be a good opportunity to do that

Yes, there will be some sort of unified refactor for all these identical `handleStream` methods across providers. It most likely will result in a refactor of `handleDefaultStreamResponseV2`.

PR #5609, opened 2026-05-12
Generic OpenAI provider for STT

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)
- [ ] ♻️ refactor (Code refactoring without changing behavior)
- [ ] 💄 style (UI style changes)
- [ ] 🔨 chore (Build, CI, maintenance)
- [ ] 📝 docs (Documentation updates)

### Relevant Issues

resolves #3885

requires #5612
requires #5610

### Description

- Adds Generic OpenAI STT provider

### Developer Validations

- [x] I ran `yarn lint` from the root of the repo & committed changes
- [x] Relevant documentation has been updated (if applicable)
- [x] I have tested my code functionality
- [x] Docker build succeeds locally

PR #5610, opened 2026-05-12
Lemonade provider for STT

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

connect #5609

### Description

- Adds Lemonade STT provider

### Additional Information

- I don't have a machine that supports whisper/TTS models (Lemonade on Apple silicon only support llamacpp models). @angelplusultra please test this to ensure it works properly

### Developer Validations


docker/.env.example
```diff
@@ -333,6 +333,10 @@ GID='1000'
 # STT_PROVIDER="openai"
 # STT_OPEN_AI_MODEL=whisper-1
 
+# STT_PROVIDER="lemonade"
+# STT_LEMONADE_BASE_PATH='http://127.0.0.1:8000'
+# STT_LEMONADE_MODEL_PREF=whisper-1
+
```
angelplusultra, line 348: Might wanna leave a comment about how the API key is reused from the provider set.

server/.env.example
```diff
@@ -338,6 +338,10 @@ STT_PROVIDER="native"
 # STT_PROVIDER="openai"
 # STT_OPEN_AI_MODEL=whisper-1
 
+# STT_PROVIDER="lemonade"
+# STT_LEMONADE_BASE_PATH='http://127.0.0.1:8000'
+# STT_LEMONADE_MODEL_PREF=whisper-1
+
```
angelplusultra, line 356: Might wanna leave a comment about how the API key is reused from the provider set.

angelplusultra (changes requested):
Code looks good to me but I wasn't really able to test it out because Lemonade transcription models (e.g whisper) aren't supported for the machine I'm on right now (macOS).

server/utils/SpeechToText/lemonade/index.js
```diff
@@ -0,0 +1,43 @@
...
+   * @param {string} filename - Original filename, used to hint the audio container/codec to the service.
+   * @returns {Promise<string>} The transcribed text.
+   */
+  async transcribe(audioBuffer, filename = "audio.webm") {
+    const { toFile } = require("openai");
+    const file = await toFile(audioBuffer, filename);
+    const result = await this.openai.audio.transcriptions.create({
+      file,
+      model: this.model,
+    });
```
angelplusultra: Whisper models will reject `webm`. A conversion process to `wav` needs to happen here before we forward it to Lemonade.

angelplusultra (changes requested):
OK so I got to test this on my Linux machine and heres the problem. As far as I'm aware, whisper models do not accept `webm`. It needs to be encoded to `wav`.

The browser's `MediaRecorder` emits `webm/opus`, and we send that straight through. The OpenAI path hides this because OpenAI transcodes server-side, Lemonade's whisper.cpp does not. I confirmed against my running Lemonade server with the same `/api/v1/audio/transcriptions` endpoint the PR hits:
```bash
  # WAV → works
  curl -X POST http://127.0.0.1:13305/api/v1/audio/transcriptions \
      -F "file=@test.wav" -F "model=Whisper-Tiny"
  HTTP/1.1 200 OK
  {"text":" [Bell]\n"}

  # WebM (same bytes the PR sends) → 500
  curl -X POST http://127.0.0.1:13305/api/v1/audio/transcriptions \
      -F "file=@test.webm" -F "model=Whisper-Tiny"
  HTTP/1.1 500 Internal Server Error
  {"error":"failed to read audio data"}
```
So `LemonadeSTT.transcribe()` needs to transcode the audio buffer to WAV before forwarding. We already have an FFMPEG wrapper in the `/collector` (`collector/utils/WhisperProviders/ffmpeg/index.js`) that does exactly this, converts arbitrary audio to WAV for whisper. 

Maybe you can create a new endpoint in the collector for audio encoding conversion e.g. `/utils/convert-audio`

pseudo code
```js
 app.post(
    "/util/convert-audio",
    [verifyPayloadIntegrity],
    async function (request, response) {
      const { audioBase64, targetFormat = "wav" } = reqBody(request);
      try {
        const input = Buffer.from(audioBase64, "base64");
        const output = await FFMPEGWrapper.convertBuffer(input, targetFormat);
        response.status(200).json({
          success: true,
          audioBase64: output.toString("base64"),
        });
      } catch (e) {
        response.status(200).json({ success: false, reason: e.message });
      }
    }
  );
```

Keep in mind that this transcoding step will also need to apply to Generic OpenAI STT as well as other providers where the downstream server doesn't transcode the audio for us.

server/utils/SpeechToText/lemonade/index.js
shatfield4: Added new collector endpoint to convert audio file to wav. Please test on your Linux machine if you are able to since I do not have a compatible machine for this.

server/utils/SpeechToText/lemonade/index.js
```diff
@@ -23,15 +28,41 @@ class LemonadeSTT {
...
+      throw new Error(result?.reason || "Audio conversion failed.");
+
+    const wavPath = path.resolve(hotdirPath, result.wavFilename);
+    try {
+      return fs.readFileSync(wavPath);
+    } finally {
+      if (fs.existsSync(wavPath)) fs.unlinkSync(wavPath);
+    }
+  }
+
```
angelplusultra: I could see other local providers that support transcription models needing this in the future. Consider extracting this into a reusable helper function that isn’t tied to a specific class (e.g. SpeechToText/helpers.js). That way, we won’t have to rewrite the same logic for each provider we add later.

server/utils/SpeechToText/lemonade/index.js
```diff
@@ -23,15 +28,41 @@ class LemonadeSTT {
...
+    const sourcePath = path.resolve(hotdirPath, sourceFilename);
+    fs.writeFileSync(sourcePath, audioBuffer);
+
+    const result = await new CollectorApi().convertAudioToWav(sourceFilename);
+    if (!result?.success || !result?.wavFilename)
+      throw new Error(result?.reason || "Audio conversion failed.");
+
+    const wavPath = path.resolve(hotdirPath, result.wavFilename);
+    try {
+      return fs.readFileSync(wavPath);
```
angelplusultra: Might wanna do async readFile from 'fs/promises'

server/utils/SpeechToText/lemonade/index.js
```diff
@@ -23,15 +28,41 @@ class LemonadeSTT {
...
+    } finally {
+      if (fs.existsSync(wavPath)) fs.unlinkSync(wavPath);
```
angelplusultra: same here

server/utils/SpeechToText/lemonade/index.js
```diff
@@ -23,15 +28,41 @@ class LemonadeSTT {
     console.log(`\x1b[32m[LemonadeSTT]\x1b[0m ${text}`, ...args);
   }
 
+  async #convertToWav(audioBuffer, extension) {
+    const sourceFilename = `stt-${v4()}${extension}`;
+    const sourcePath = path.resolve(hotdirPath, sourceFilename);
+    fs.writeFileSync(sourcePath, audioBuffer);
```
angelplusultra: same here, use async

server/utils/SpeechToText/lemonade/index.js
```diff
@@ -1,4 +1,9 @@
+const fs = require("fs");
```
angelplusultra: async `fs` functions are available in `fs/promises`. Keeps the event loop unblocked.

server/utils/collectorApi/index.js
```diff
@@ -207,6 +207,39 @@ class CollectorApi {
...
+          new EncryptionManager().xPayload
+        ),
+      },
+      body: data,
+    })
+      .then((res) => {
+        if (!res.ok) throw new Error("Response could not be completed");
+        return res.json();
+      })
+      .then((res) => res)
```
angelplusultra: Rename the callback param to `data` or `json` since its not the raw response anymore, its the decoded json data.

collector/convertAudioToWav/index.js
```diff
@@ -0,0 +1,51 @@
...
+
+  const inputPath = normalizePath(path.resolve(WATCH_DIRECTORY, filename));
+  if (!isWithin(path.resolve(WATCH_DIRECTORY), inputPath))
+    return {
+      success: false,
+      reason: "Filename is outside the hotdir.",
+      wavFilename: null,
+    };
+
+  if (!fs.existsSync(inputPath))
```
angelplusultra: async

server/utils/collectorApi/index.js
```diff
@@ -207,6 +207,39 @@ class CollectorApi {
...
 
+  /**
+   * Convert an audio file in the shared hotdir to WAV via the collector's
+   * FFMPEG wrapper. The source file is trashed by the collector; caller must
+   * read and trash the resulting wav file.
+   * @param {string} filename - The filename of the source audio in the hotdir
+   * @returns {Promise<Object>} - The response from the collector API
+   */
+  async convertAudioToWav(filename = "") {
+    if (!filename) return false;
```
angelplusultra: Inconsistent return contract, have this return a "result" pattern object like the rest of the pathways.

server/utils/SpeechToText/lemonade/index.js
```diff
@@ -23,15 +28,41 @@ class LemonadeSTT {
...
   }
 
+  async #convertToWav(audioBuffer, extension) {
+    const sourceFilename = `stt-${v4()}${extension}`;
+    const sourcePath = path.resolve(hotdirPath, sourceFilename);
+    fs.writeFileSync(sourcePath, audioBuffer);
+
+    const result = await new CollectorApi().convertAudioToWav(sourceFilename);
+    if (!result?.success || !result?.wavFilename)
+      throw new Error(result?.reason || "Audio conversion failed.");
```
angelplusultra: I think this will leave an orphaned temp file since if this throws nothing cleans it up?

angelplusultra (changes requested):
Dope, this works. Overall pretty solid, I recommend converting the file system operations to their async counterparts with `fs/promises`. Left some additional nits.

angelplusultra (approved):
LGTM

PR #5456, opened 2026-04-16
feat: Integrate Deepgram for real-time STT streaming and TTS synthesis

yugborana:
### Pull Request Type

- [yes ] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)
- [ ] ♻️ refactor (Code refactoring without changing behavior)
- [ ] 💄 style (UI style changes)
- [ ] 🔨 chore (Build, CI, maintenance)
- [ ] 📝 docs (Documentation updates)

### Relevant Issues

resolves #4732 

### Description

This PR introduces Deepgram as a native provider option for both Speech-to-Text (STT) and Text-to-Speech (TTS) within AnythingLLM.

### Developer Validations

- [ yes] I ran `yarn lint` from the root of the repo & committed changes
- [ ] Relevant documentation has been updated (if applicable)
- [yes ] I have tested my code functionality
- [ yes] Docker build succeeds locally

PR #5612, opened 2026-05-12
Deepgram provider for STT

### LINKED ISSUE #4732: [FEAT]: Support Deepgram APIs for TTS and STT in AnythingLLM

Hello,

AnythingLLM is some pretty sweet software. I'm just getting started with it, along with Ollama and LM Studio, but so far everything works perfectly!

I noticed that AnythingLLM supports custom speech-to-text (STT) and text-to-speech (TTS) providers, since 2024. https://github.com/Mintplex-Labs/anything-llm/pull/909

It would be pretty cool if Deepgram were supported as a transcription and text-to-speech provider in AnythingLLM. There are WebSocket APIs for live STT and TTS, as well as HTTP endpoints for submitting pre-recorded audio.

Here are the WebSocket APIs that would be more suitable for real-time usage:

https://developers.deepgram.com/reference/speech-to-text/listen-streaming
https://developers.deepgram.com/reference/text-to-speech/speak-streaming

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #4732 
connect #5609
closes #5456

### Description

- Adds Deepgram STT provider

### Additional Information

- You can get a free API key from Deepgram to test this

### Developer Validations


server/utils/helpers/customModels.js
```diff
@@ -956,6 +959,44 @@ async function getLemonadeModels(basePath = null, task = "chat") {
...
+ * Get Deepgram STT models from the Management API.
+ * https://api.deepgram.com/v1/models returns { stt: [...], tts: [...] }.
+ * @param {string} _apiKey - Deepgram API key. Falls back to STT_DEEPGRAM_API_KEY.
+ * @returns {Promise<{models: Array<{id: string, name: string, organization: string}>, error: string | null}>}
+ */
+async function getDeepgramSTTModels(_apiKey = null) {
+  const apiKey =
+    _apiKey === true
+      ? process.env.STT_DEEPGRAM_API_KEY
+      : _apiKey || process.env.STT_DEEPGRAM_API_KEY || null;
```
angelplusultra, line 972: This ternary seems off to me, seems needlessly complex. Basically use argument api key if exists otherwise `process.env.STT_DEEPGRAM_API_KEY` otherwise null? 
```js
const resolvedApiKey = apiKey || process.env.STT_DEEPGRAM_API_KEY || null;
```

server/utils/helpers/customModels.js
```diff
@@ -956,6 +959,44 @@ async function getLemonadeModels(basePath = null, task = "chat") {
...
 
+/**
+ * Get Deepgram STT models from the Management API.
+ * https://api.deepgram.com/v1/models returns { stt: [...], tts: [...] }.
+ * @param {string} _apiKey - Deepgram API key. Falls back to STT_DEEPGRAM_API_KEY.
+ * @returns {Promise<{models: Array<{id: string, name: string, organization: string}>, error: string | null}>}
+ */
+async function getDeepgramSTTModels(_apiKey = null) {
+  const apiKey =
+    _apiKey === true
```
angelplusultra, line 970: Would this ever be `true` when the JSdoc say `_apiKey` is a string? Don't think it ever comes through as a boolean.

server/utils/helpers/customModels.js
```diff
@@ -956,6 +959,44 @@ async function getLemonadeModels(basePath = null, task = "chat") {
...
+    const data = await response.json();
+    const models = (data?.stt ?? [])
+      .filter((m) => m.batch !== false)
+      .map((m) => ({
+        id: m.canonical_name,
+        name: m.name || m.canonical_name,
+        organization: "Deepgram",
+      }));
+
+    if (models.length > 0 && _apiKey) process.env.STT_DEEPGRAM_API_KEY = apiKey;
```
angelplusultra, line 998: What's this for? Why mess with the env vars?

angelplusultra (changes requested):
some nits

server/utils/helpers/customModels.js
shatfield4: Same pattern as other providers, we are saving the key to the env if successful so that the user does not have to type it in again once we know we have a valid response from the API.

server/utils/helpers/customModels.js
shatfield4: This follows the same patterns we use in other providers so we are just matching it here since it functions the same.

server/utils/helpers/customModels.js
shatfield4: Keeping this consistent since it functions the same in all other providers and just matching the same ternary we use in multiple other providers.

embed
angelplusultra, line 1: Was this intentional? Submodule changes?

angelplusultra (approved):
Asked about the incoming submodule changes, otherwise LGTM

frontend/src/media/ttsproviders/deepgram.png
timothycarambat, line 1: is this file 300x300? It is 1.3mb!!

shatfield4: Oops, uploaded the wrong file. Is fixed now to be 300x300.

timothycarambat (dismissed):
Deepgram STT does not take a model as an argument, so why is the model required at all to run STT? If this is the case, we should also drop the model picker since that is only for TTS fields.

Then we can simplify this dramatically since there is no model to even choose and its all just off an API key?

This is the case from what I can tell, but why have the picker and requirement in STT at all?

shatfield4:
> Deepgram STT does not take a model as an argument, so why is the model required at all to run STT? If this is the case, we should also drop the model picker since that is only for TTS fields.
> 
> Then we can simplify this dramatically since there is no model to even choose and its all just off an API key?
> 
> This is the case from what I can tell, but why have the picker and requirement in STT at all?

Deepgram STT does take a model as an argument. Here's a link to [their docs](https://developers.deepgram.com/docs/model) showing all the different STT models they support.

They have different STT models for different use cases like conversations, finance, medical and so on. We do only fetch the STT models as well from their API (using `https://api.deepgram.com/v1/listen` endpoint) when populating the model picker so no TTS models show up here.

timothycarambat:
@shatfield4 Ahhh, my mistake I expected it in a POST body and it wasnt listed here: https://developers.deepgram.com/reference/speech-to-text/listen-pre-recorded. I didnt expect it as a URL param, kinda odd. nvm!

PR #5620, opened 2026-05-13
Fix agent context window overflow and provider tag leak

shatfield4:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #

### Description

- Agents were crashing on follow-up chats with a `Converting circular structure to JSON` error.
- Root cause: `aibitat.provider` was being overwritten from the string `"openai"` to the provider class instance after each turn, so `Provider.contextLimit` silently fell back to an 8k window for any tool that called it.
- Fixed by storing the active provider instance on `aibitat.activeProvider` so the string tag stays intact.
- Added a defensive warning in `Provider.contextLimit` so this kind of regression can't sneak back in silently.
- Added a trim that pops the oldest user/assistant pairs from agent history when the conversation grows past the model's context window — keeps the system prompt and current user message intact.
- Trim reserves 15% of the context window for the model's response and tools.
- Logs a warning and an introspect line whenever trimming happens so we can see it in user reports.

### Additional Information

- Bug introduced in this PR: https://github.com/Mintplex-Labs/anything-llm/pull/5197

### Developer Validations


server/utils/agents/aibitat/utils/messageWindow.js
```diff
@@ -0,0 +1,77 @@
...
+ * in pairs so the alternating user/assistant structure is preserved.
+ *
+ * @param {Object} args
+ * @param {Array<{role: string, content: any}>} args.messages
+ * @param {string} args.provider - Provider tag (e.g., "openai").
+ * @param {string} args.model - Model name.
+ * @param {Function} [args.onTruncate] - Called once with `{ droppedPairs, contextLimit, budget, remainingTokens }` when truncation occurs.
+ * @returns {Array} The trimmed messages (new array; original is not mutated).
+ */
+function trimMessagesToContextWindow({
```
angelplusultra, line 36: What about tools? Don't they need to be accounted for against the models budget?

angelplusultra (changes requested):
`trimMessagesToContextWindow` only sums `message.content`, but the actual request to the model also carries the function schemas. A request can still overflow *after* trimming, because the trimmer doesn't know about them.

shatfield4:
> `trimMessagesToContextWindow` only sums `message.content`, but the actual request to the model also carries the function schemas. A request can still overflow *after* trimming, because the trimmer doesn't know about them.


The 15% buffer is meant to take the tools into consideration and also the models response. The tool call tokens are provider specific so we wouldn't be able to calculate all of these in a simple way without modifying many other files and creating custom tool token count utils for every provider. Users should lean on the ToolReranker (intelligent skill selection) instead of relying on the trimMessagesToContextWindow util.

I did modify the comment to clarify this.

angelplusultra (approved):
LGTM

timothycarambat:
Marking as blocked for now, ran up on some very odd cases and behavior the ultimately lead to the following:

OK — the changes have already been reverted, so there's nothing to roll back. Here's the write-up:

1. **Tokenizer mismatch is unsolvable from our side.** We use `cl100k_base` (GPT tokenizer) as our approximation, but every provider uses its own tokenizer — Qwen, Llama, Gemma, Mistral, etc. all tokenize differently. In testing, I saw 30-40% discrepancies between our estimate and the provider's actual count (LMStudio Qwen3.5 9B). No safety margin we pick will be correct across all models, and being too conservative wastes context while being too loose still causes provider errors. This was the case with LMStudio throwing errors because context sent > context permitted.

2. **Chat template overhead is invisible to us.** Providers wrap messages in chat templates that add special tokens, role markers, BOS/EOS tokens, etc. These are completely opaque to us — we can't know how many tokens `<|im_start|>system\n...<|im_end|>` adds for Qwen vs `[INST]...[/INST]` for Llama. This framing overhead compounds with each message and tool definition. There is just no simple way rn to get accurate tokens without an HTTP request preflight **every** chat.

3. **Reported context windows don't match loaded context windows.** LMStudio (and similar local providers) report `max_context_length` from the model card (e.g., 32K) via their API, but the model may actually be loaded at a much smaller `n_ctx` (e.g., 1024 or 4096). We have no way to know the real loaded context size, so our budget calculation can be completely wrong in either direction.

4. **Content truncation does more harm than good.** When I tried middle-out truncation as a fallback, the mangled content confused models and produced worse results than just letting the provider handle the overflow. Providers that support it (LMStudio, Ollama) have their own sliding window truncation that respects their tokenizer and chat template. Providers that don't (OpenAI, Anthropic) return clear errors the user can act on. Worst case, they /reset

5. **Tool definitions are the biggest wildcard.** Each provider serializes tools differently — OpenAI wraps them in `{"type":"function","function":{...},"strict":true}`, Anthropic uses `input_schema`, Bedrock uses LangChain tool format. The token "cost" varies significantly per provider and we can't accurately estimate it without duplicating each provider's serialization logic. Again, here we woudl need some like near instant cheap tokenizer

**What we should do instead:** The right long-term approach is to let each provider own its own context management (as regular chats already do via `compressMessages()`), since the provider knows its tokenizer, chat template, and actual context limit. The segment-dropping concept (removing old conversation turns) is sound, but it needs to be driven by **accurate** token counts from the provider side, not by our approximation because otherwise we are not hitting the right number to measure against. This matters much more with agents than regualr chats

PR #5630, opened 2026-05-14
fix: prevent gemini agent 400s from parallel tool calls

### LINKED ISSUE #5600: [BUG]: Agent chat fails too fast without trying to recover

Docker (local)

### What happened?

I'm asking Gemini to write an app. At some point it tries to read a file which doesn't exist, and instead of recovering trying a different path it fails the whole process.

It's easy to make it fail, just say write an app abc then modify it to do xyz, and inevitably it will hallucinate but it should have recovered. Gemini CLI for example is able to recover from mistakes and find other paths to success whenever a tool fails:

```
[backend] info: [AgentHandler] filesystem-write-text-file error: Parent directory does not exist: /app/server/storage/anythingllm-fs/go-mcp-cloudrun/terraform
```

Image

Am I missing some way to configure this behavior in AnythingLLM?

Just say to write e.g. a golang app abc then modify it to do xyz, at some point it will try to read a non-existent file and fail instead of trying to recover.

gemini-3-flash-preview

_No response_

shatfield4:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5600

### Description

- This issue was caused by Gemini emitting parallel tool calls in a single response. Our streaming loop only tracks one `functionCall` and was overwriting it with each new chunk, so we executed only one tool but Gemini's API state still expected results for all of them. The mismatched tool call/result count caused Gemini to reject the next request with `400 status code (no body)`, which killed the chat before the agent could recover.
- Set `parallel_tool_calls: false` on Gemini `stream()` and `complete()` requests so the model emits one tool call per turn.
- Added a defensive guard in the streaming loop to keep only the first tool call if Gemini ignores `parallel_tool_calls: false` and emits parallel calls anyway.

### Developer Validations


timothycarambat:
Is there an easy repro for this to test?

PR #5637, opened 2026-05-15
feat: configurable socket-idle timeouts for AI providers

### LINKED ISSUE #5603: [BUG]:  LMStudio Agent loop socket timeout

AnythingLLM desktop app

### What happened?

Image

I am experiencing a recurring issue while using the AnythingLLM Desktop application with Agent Skills enabled. Specifically, I triggered an agent task that required the model to read and process my local files using the file system tool. The dataset contains approximately 1,600 files, so it is expected that the model would take a considerable amount of time to process and analyze the data.

During execution, the model begins processing normally and reaches approximately 11% progress. However, at this stage, a Socket Timeout error occurs in AnythingLLM. When I checked the LM Studio local server logs, I found the message: “client disconnected”.

It appears that AnythingLLM is not maintaining the connection long enough for the local model to complete its processing. As a result, the model is forcibly disconnected before it is able to return a response. This behavior interrupts the execution of long-running agent tasks, even though the model (Qwen 3.5-9B) is still actively processing in LM Studio.

This issue occurs specifically during heavy Agent Skill operations such as large-scale file reading and RAG-based workflows. Normal chat interactions without Agent Skills do not seem to trigger this problem.

I kindly request the developers to investigate this behavior, as it significantly affects the usability of Agent Skills with local models. I hope this issue can be addressed and improved in future updates. Thank you for your work and support.

Image

LM studio / Qwen 3.5-9b

text-embedding-nomic-embed- txt- v1.5

### LINKED ISSUE #5631: [FEAT]: Custom Timeout Values for Socket Connections

https://github.com/Mintplex-Labs/anything-llm/issues/5603

### LINKED ISSUE #5681: OpenAI and Anthropic provider calls should use explicit timeouts and retry/backoff

```md
Support/document OpenAI-compatible gateway base URLs for production reliability
```

**Issue body**

````md
## Question

Would maintainers be open to a small PR that documents or improves support for OpenAI-compatible gateway base URLs?

Some production users route LLM calls through a gateway layer so they can manage:

- retries
- fallback routing
- request tracking
- provider timeout handling
- debugging failed provider calls

For example, an OpenAI-compatible client can usually point at a gateway by changing only the API key and base URL:

```ts
import OpenAI from "openai";

const openai = new OpenAI({
  apiKey: process.env.BADGR_API_KEY,
  baseURL: "https://api.aibadgr.com/v1",
});
````

## Why I’m asking

I noticed a few LLM provider paths where reliability behavior may currently depend on the underlying provider/client defaults:

* `server/utils/AiProviders/openAi/index.js`

  * OpenAI requests appear to be issued directly without an explicit timeout or abort budget
  * completion calls appear to rely on direct provider errors without a shared retry/backoff policy
  * error rethrows may drop useful provider metadata such as status code, headers, request IDs, or retry hints

* `server/utils/AiProviders/anthropic/index.js`

  * Anthropic requests appear to run without an explicit timeout or abort budget
  * transient `429` / `5xx` provider errors may fail immediately
  * one error path appears to return the error object as `textResponse`, which could be treated like a successful model response upstream

* `server/utils/chats/apiChatHandler.js`

  * chat execution appears to use one resolved provider per request without an optional runtime fallback chain if that provider is unavailable

* `server/utils/helpers/chat/LLMPerformanceMonitor.js`

  * performance tracking appears to capture duration/token metrics, but not full upstream failure taxonomy such as provider status, retry count, fallback step, timeout reason, or request correlation ID

These may already be handled elsewhere, so I wanted to ask before opening a larger PR.

## Possible small PR options

Happy to follow whatever maintainers prefer:

1. a small docs PR showing how to configure an OpenAI-compatible gateway base URL
2. a small config improvement for custom OpenAI-compatible base URLs
3. a small timeout/retry improvement in one provider path
4. improved structured error metadata for provider failures
5. no change, if this is outside the project scope

The goal would be optional production reliability support without changing default behavior.

### PR DESCRIPTION
**Pull request by angelplusultra**
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5631
resolves #5603
resolves #5681

### Description

Adds a per-provider, env-configurable socket-idle timeout for every AI provider whose SDK exposes a fetch injection point (the OpenAI SDK, Ollama's client, and the Anthropic SDK), so slow local models stop tripping the SDK's hardcoded 5-minute socket idle timeout and surfacing ERR_SOCKET_TIMEOUT.

**New helper**

server/utils/AiProviders/helpers/index.js — getFetchWithCustomTimeout(rawTimeoutValue, providerSlog, fallbackTimeoutValue?). Given an env value, returns either the global fetch (when no timeout is resolved) or a wrapped fetch whose dispatcher is an undici.Agent with headersTimeout and bodyTimeout set to the resolved value. This bypasses the OpenAI SDK's node-fetch + agentkeepalive stack — that's where the 5-minute idle cap lives — and applies to gaps between streamed chunks as well as the wait for response headers.

Resolution order: valid positive integer → use it; invalid value → log via the provider's slog and fall through; fallback (when provided) → use it; otherwise → unwrapped fetch.

Note on retries: the OpenAI SDK retries failed requests (default maxRetries: 2), so a timeout-aborted attempt can still be retried up to twice before surfacing. Not changed here, but documented in the helper JSDoc for callers who want a hard wall-clock cap.

**Applied to**

The helper is wired into every provider that constructs an OpenAI SDK client, an Ollama client, or an Anthropic SDK client across three layers:

- server/utils/AiProviders/* — LLM providers
- server/utils/EmbeddingEngines/* — embedding engines
- server/utils/agents/aibitat/providers/* — agent providers

Each constructor now passes fetch: getFetchWithCustomTimeout(process.env.<PROVIDER>_RESPONSE_TIMEOUT, <Class>.slog). Several #slog private static loggers were promoted to public slog so the helper could surface its setup/parse logs through the right provider prefix.

**Ollama refactor**

Ollama previously had its own applyOllamaFetch() that only honored timeouts greater than 5 minutes. That bespoke implementation is removed in favor of the shared helper, and a 15-minute fallback (DEFAULT_OLLAMA_SOCKET_TIMEOUT = 900000) is passed in so existing setups keep working without env changes.

**Out of scope**

Cohere and AWS Bedrock are not covered. Their SDKs do not expose a standard fetch injection point compatible with this helper:

- Cohere accepts a fetcher option, but its signature is Fern's (args: Fetcher.Args) => Promise<APIResponse>, not standard fetch — wiring it would require a Cohere-specific adapter.
- Bedrock builds on @smithy/node-http-handler over node:http, not fetch — it would need a Smithy requestHandler with requestTimeout instead.

These can be added in follow-ups; they're intentionally deferred to keep this PR focused on the providers the same helper can cover unmodified.

**Docs**

server/.env.example documents <PROVIDER>_RESPONSE_TIMEOUT for each affected provider. The comment phrasing makes clear this is a socket-idle timeout (waiting for headers, or a gap between streamed chunks) — not a total request timeout, since a long-running streaming response can legitimately exceed this value.

### Visuals (if applicable)

N/A — backend change with no UI surface.

### Additional Information

- Unit tests for the helper live in server/__tests__/utils/AiProviders/helpers/index.test.js and cover the resolution order: undefined/null/empty/non-numeric/zero/negative values, valid integers, and fallback behavior.
- The previous Ollama-specific OLLAMA_RESPONSE_TIMEOUT env var keeps working as before; its .env.example entry was normalized to the same phrasing as the others and the stale "default is 5min" note was corrected to 15min to match the new fallback.

### Developer Validations

- [x] I ran yarn lint from the root of the repo and committed changes

### CODE REVIEW

File: server/utils/agents/aibitat/providers/ollama.js [hunk 1]
```diff
@@ -6,7 +6,11 @@ const { OllamaAILLM } = require("../../../AiProviders/ollama");
 const { Ollama } = require("ollama");
 const { v4 } = require("uuid");
 const { safeJsonParse } = require("../../../http");
+const {
+  getFetchWithCustomTimeout,
+} = require("../../../AiProviders/helpers/index.js");
 
+const DEFAULT_OLLAMA_SOCKET_TIMEOUT = 90000; // 15 min
```
shatfield4: This is 90 seconds, missing a 0 here for 15 mins.

server/utils/AiProviders/helpers/index.js
```diff
@@ -0,0 +1,84 @@
...
+    const moment = require("moment");
+
+    const customFetch = (input, init = {}) => {
+      return fetch(input, {
+        ...init,
+        dispatcher: new Agent({
+          headersTimeout: timeout,
+          bodyTimeout: timeout,
+        }),
+      });
```
shatfield4: This part of the code is instantiating a new agent instance on every single call to this helper. Can we create a cache for the agent instance so that we can continue using that same instance if it exists? By instantiating a new instance every time, we are losing things like keep-alive and the sockets aren't closed so this would create a lot of orphaned instances of agent.

shatfield4 (changes requested):
Typo and suggestion to help avoid orphaned agent instances.

server/utils/agents/aibitat/providers/ollama.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5637/commits/6e9928bc8f289877c91a17cc3db8d889a0ba87e9

server/utils/AiProviders/helpers/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5637/commits/8afad0f7b05f1e868fd674477b84ba68f7e0172f

shatfield4 (approved):
LGTM

shatfield4 (approved):
LGTM

PR #5647, opened 2026-05-17
feat: MCP system prompt variable substitution

### LINKED ISSUE #5646: [FEAT]: MCP system prompt variables

Often, there is information that we want to make passively available to the model, without requiring an agent to explicitly call a tool. Examples might be a server status for an administration workflow, or the user's currently playing music track for a companion or entertainment setup. We can inject these into the system prompt, in the same manner as date and time, with a {mcp.<serverName>.<toolName>} structure.

pageragatz:
## Summary

Adds `{mcp.<server>.<tool>}` variable support to workspace system prompts so any running MCP server tool can be called inline at prompt-expansion time. Closes #5646.

* New resolver in `SystemPromptVariables.expandSystemPromptVariables`: parses the key, looks up the MCP client via the existing `MCPCompatibilityLayer` singleton, calls the tool with no arguments, and substitutes the result string.
* Boot-idempotent — calls `bootMCPServers()` lazily so the variable works on cold start even if no agent has been invoked yet.
* TTL cache (30s) on successful results, negative cache (10s) on failures, and a 3s hard timeout per resolution so a broken or slow MCP server can never block or 500 a chat request — failures fall back to an empty string.
* Frontend hint in **Workspace → Chat Settings** explains the new `{mcp.<server>.<tool>}` syntax next to the existing `{time}`, `{date}`, `{datetime}` examples.

## How it works

Example: a workspace system prompt of

```
You are a helpful assistant. The currently playing track is: {mcp.deezer.lastfm\_get\_now\_playing}
```

…will, on every message, call the `lastfm\_get\_now\_playing` tool on the running `deezer` MCP server (configured in `server/storage/plugins/anythingllm\_mcp\_servers.json`) and inline the result before the prompt is sent to the LLM. Tool names that contain dots (e.g. `mcp.foo.namespace.tool`) work — only the first dot splits server from tool name.

The result string is extracted from the MCP `callTool` response by preferring concatenated `text` content blocks, falling back to `JSON.stringify(result)` if no text blocks are present.

## Design notes

* **Lazy require** of `../utils/MCP` avoids a circular dependency at module load.
* **No tool arguments** in this iteration — variable resolution always calls with `arguments: {}`. The intended use case is a parameterless "fetcher" tool whose output the LLM should see verbatim. Parameterised variables are a natural follow-up but out of scope here.
* **Single-pass expansion** — resolved values are not re-scanned for variables, which prevents recursion attacks via MCP-returned strings.
* **Hard timeout** uses `Promise.race` with the loser-timer cleared in `finally` so each successful resolution doesn't leak a 3s `setTimeout` into the event loop.
* **Fail-soft contract**: any error path (server not running, tool throws, timeout, malformed key) resolves to `""`. The user sees an empty substitution rather than a chat error.

## Test plan

Unit tests in `server/\_\_tests\_\_/models/systemPromptVariables.test.js` cover:

* \[x] Text-content extraction for a normal MCP tool result
* \[x] JSON fallback when no text content blocks are returned
* \[x] Empty-string fallback when the MCP server isn't running
* \[x] Empty-string fallback when the tool call throws
* \[x] Empty-string fallback when the tool call exceeds the 3s timeout (verified via `jest.useFakeTimers`)
* \[x] Malformed `{mcp.foo}` (no tool segment) → empty string
* \[x] Mixed expansion alongside `{workspace.name}` and static variables
* \[x] **Cache verification**: three resolutions of the same variable across two `expandSystemPromptVariables` calls result in exactly one underlying `callTool` (TTL cache works)
* \[x] **Negative cache**: three resolutions of a permanently-failing tool result in exactly one underlying `callTool` (failure path is also cached)
* \[x] **Dotted tool names**: `{mcp.dotted-server.namespace.deep\_tool}` calls `callTool({ name: "namespace.deep\_tool", arguments: {} })`

Manual smoke test:

* \[x] Configured a real `deezer` MCP server and verified that `{mcp.deezer.lastfm\_get\_now\_playing}` expands to a JSON `Current track: {...}` string in chat.
* \[x] Confirmed that with the MCP server stopped, the variable resolves to empty and the chat still works.

## Files changed

* `server/models/systemPromptVariables.js` — resolver, cache, timeout, extractor; new `mcp.` branch in `expandSystemPromptVariables`.
* `server/\_\_tests\_\_/models/systemPromptVariables.test.js` — 10 new tests, shared-mock refactor.
* `frontend/src/pages/WorkspaceSettings/ChatSettings/ChatPromptSettings/index.jsx` — hint line explaining the new syntax.
* `server/.gitignore` — ignore `\_\_tests\_\_/utils/agents/models/` cache files generated by `defaults.test.js`.

### DIFF
--- a/frontend/src/pages/WorkspaceSettings/ChatSettings/ChatPromptSettings/index.jsx
+++ b/frontend/src/pages/WorkspaceSettings/ChatSettings/ChatPromptSettings/index.jsx
@@ -132,7 +132,7 @@ export default function ChatPromptSettings({
           <p className="text-white text-opacity-60 text-xs font-medium py-1.5">
             {t("chat.prompt.description")}
           </p>
-          <p className="text-white text-opacity-60 text-xs font-medium mb-2">
+          <p className="text-white text-opacity-60 text-xs font-medium mb-1">
             You can insert{" "}
             <Link
               to={paths.settings.systemPromptVariables()}
@@ -157,6 +157,17 @@ export default function ChatPromptSettings({
                 +{availableVariables.length - 3} more...
               </Link>
             )}
+          </p>
+          <p className="text-white text-opacity-60 text-xs font-medium mb-2">
+            Call any running MCP server tool with{" "}
+            <span className="bg-theme-settings-input-bg px-1 py-0.5 rounded font-mono">
+              {"{mcp.<server>.<tool>}"}
+            </span>
+            {" "}— e.g.{" "}
+            <span className="bg-theme-settings-input-bg px-1 py-0.5 rounded font-mono">
+              {"{mcp.weather.get_forecast}"}
+            </span>
+            . Results are cached for 30 seconds.
           </p>
         </div>
 
--- a/frontend/src/pages/WorkspaceSettings/ChatSettings/ChatPromptSettings/index.jsx
+++ b/frontend/src/pages/WorkspaceSettings/ChatSettings/ChatPromptSettings/index.jsx
@@ -163,9 +163,13 @@ export default function ChatPromptSettings({
             <span className="bg-theme-settings-input-bg px-1 py-0.5 rounded font-mono">
               {"{mcp.<server>.<tool>}"}
             </span>
-            {" "}— e.g.{" "}
+            {" "}or pass arguments with{" "}
             <span className="bg-theme-settings-input-bg px-1 py-0.5 rounded font-mono">
-              {"{mcp.weather.get_forecast}"}
+              {"{mcp.<server>.<tool>:arg1,arg2}"}
+            </span>
+            {" "}— args can be literals or other variable keys like{" "}
+            <span className="bg-theme-settings-input-bg px-1 py-0.5 rounded font-mono">
+              {"workspace.name"}
             </span>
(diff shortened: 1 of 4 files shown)


timothycarambat:
This is an interesting approach and in fact, there was/is a `dynamic` systemVar type - that was the intended use case for this but it **was not** about MCPs, it was just for simple REST calls so you could fetch the data directly from the endpoint and inject it with a cache.

Since MCP *can* be just REST, would not moving this to just a REST call with caching be more simple/generic and not be coded specifically to MCP?

pageragatz:
That does seem to be the approach to simplify more workflows.

I'm thinking of extending the "System Prompt Variables" -> "Add New Variable" menu, building from the existing REST API agent flow feature.

I'm happy to close this and come back with that design, if that makes sense to you.

timothycarambat:
I think it does, might as well have it be more "generic" since the idea of system prompt variables is mostly fetching remote information and injecting it at prompt-runtime and caching it.

If that REST endpoint is an MCP, well then so be it! The complex part will be parsing the output from the JSON object since the user is certainly looking for a single value from the output - assuming it is even json!

PR #5668, opened 2026-05-21
docs: list Tuning Engines gateway

cerebrixos:
Adds Tuning Engines to the supported LLM provider list as an OpenAI-compatible gateway. Users can configure it through AnythingLLM's existing OpenAI Generic path.


timothycarambat:
That list is for explicitly supported providers, there are very many openAI compatible gateways out there - we cannot possibly list them all

cerebrixos:
Thanks for the earlier clarification. I opened an issue asking whether a native provider contribution would be preferable to a generic gateway-listing PR: https://github.com/Mintplex-Labs/anything-llm/issues/5724

PR #5677, opened 2026-05-21
fix: Various Agent Stream Reasoning Bugs & Presentation Refactor

### LINKED ISSUE #5676: [BUG]: Non-Ollama Agent Providers Do Not Parse and Present Reasoning Content

Docker (local)

### What happened?

Any agent provider besides Ollama does not parse and present a reasoning models visible CoT.

1. Setup a non Ollama provider such as LM Studio.
2. Use a reasoning model that outputs it's CoT (e.g `gemma4`). 
3. Prompt the model in agent mode or using `@agent` and notice no CoT is being outputted for the user to read.

Any Agent Provider that is not Ollama

_No response_

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5676
UPSTREAM PR: https://github.com/Mintplex-Labs/anything-llm/pull/5972

### TLDR

This PR fixes a few things:

- Surfaces thinking tokens in agent mode for both the `tooledStream` and `untooled` stream pathways (any provider emitting `reasoning_content`), instead of only Ollama.
- Unifies how reasoning is presented: it now streams live into the collapsible `StatusResponse` bubble (same place tool activity shows) and is ephemeral — a reloaded conversation shows only tool status + the final answer. This replaces the old `<think>` thought-pane that stacked a separate pane per step.
- Ollama stream refactor: Ollama does not use `tooled` or `untooled`stream pathways, instead its own bespoke implementation using `</think>` tags. It has been refactored to match the behavior in `tooled` and `untooled` 
- Fixes the pulsing head icon. It used to stop pulsing the moment a tool call or the answer rendered, and on untooled reasoning turns it could get stuck pulsing forever. It now pulses through the whole agent status stream (thinking + tool calls) and settles to static once the final response streams or the agent goes idle.
- Fixes untooled streams where the thoughts would vanish the instant the final response started streaming — they now persist in the collapsible status bubble, matching tooled.

### Description

In agent mode, only `ollama` surfaced reasoning — and it did so as a standalone `<think>` thought pane that was persisted into the saved message. Every other provider routing through the shared OpenAI-compatible helpers (`tooled.js` / `untooled.js`) dropped `reasoning_content` entirely, so users got a long pause then only the final answer.

This PR does two things:

- **Surfaces reasoning for all providers** routing through the shared helpers (any model emitting `reasoning_content`).
- **Unifies how reasoning is presented.** It now streams live into the collapsible `StatusResponse` bubble (the same place tool activity shows), framed with `"Thinking:"`, and is no longer persisted — a reloaded conversation shows only tool status + the final answer. This replaces the old `<think>` thought-pane approach, which stacked a separate pane per step in multi-step runs (per review feedback).

This is a behavior change for Ollama too: its agent reasoning used to render as a thought pane and survive reload, and is now ephemeral.

**Scope:**
- Covered: the streaming paths (`tooledStream`, `UnTooled.stream` / `streamingFunctionCall`) plus Ollama.
- Out of scope (native, non-OpenAI-shaped reasoning streams, follow-up?): `anthropic.js`, `openai.js`, `gemini.js`, `bedrock.js`. 
- Regular (non-agent) chat is unchanged — it still persists reasoning as `<think>` panes. The agent/regular asymmetry is noted for a separate discussion.

**Implementation notes:**
- Reasoning streams under its own event id (`${msgUUID}:reasoning`) so the answer's `textResponseChunk` doesn't merge into the reasoning bubble and render thoughts as visible text.
- Frontend: the TTS/copy action bar is now gated on a `visibleResponseContent()` helper so reasoning-only bubbles don't get a copy button that copies an empty string.

**Agent head animation (pulse) fix:**

The animated agent head is meant to pulse while the agent is actively thinking or invoking a tool, and go static once the final answer is streaming or the agent is idle. It was instead deriving "is the agent working" from the chat tail — pulse while a session is open and the last message is a status bubble. In an interactive `@agent` session the socket stays open between turns, so any turn that ended on a status bubble (e.g. an untooled reasoning turn that finished on a thought with no trailing answer) left the head pulsing forever.

It now drives the animation off the live stream events instead of the rendered tail:
- A `statusResponse` (thinking) or `toolCallInvocation` event starts the pulse.
- The final answer (`textResponseChunk` / `fullTextResponse`) or the end-of-turn markers (`usageMetrics` / `chatId`, which the agent backend emits at the end of every turn even when the answer is empty) settle it to static, as does a session start/end.

This is tracked with a small `useIsAgentThinking()` hook (mirroring `useIsAgentSessionActive`) fed by `AGENT_THINKING_START` / `AGENT_THINKING_STOP` events dispatched from the websocket handler.

**Untooled reasoning persistence fix:**

In tooled mode the reasoning bubble persists after the answer lands (collapsible), but in untooled mode the thoughts disappeared the moment the response finished. `streamingFunctionCall` was streaming both the reasoning and the tentative tool-call/answer preview under the same `msgUUID`, then calling `removeStatusResponse(msgUUID)` to clear the preview once the turn resolved — which wiped the thoughts along with it. Reasoning now streams under its own `${msgUUID}:reasoning` bubble so the cleanup only removes the preview, matching `tooled.js` and the existing second-stream path.

### Visuals (if applicable)

image

### Additional Information

- Tested across tooled and untooled paths with reasoning models (LM Studio, Koboldcpp, DeepSeek, Lemonade, Moonshot) in both agent and regular chat.

### Developer Validations

### DIFF
--- a/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
@@ -57,6 +57,14 @@ const HistoricalMessage = ({
   const isRefusalMessage =
     role === "assistant" && message === chatQueryRefusalResponse(workspace);
 
+  // Agent loop turns that reason and then call a tool emit a message whose
+  // entire content is a `<think>...</think>` block with no visible answer.
+  // Those bubbles have nothing to copy/speak/edit, so suppress the action bar
+  // (otherwise a hover copy button appears under each thought and copies an
+  // empty string once useCopyText strips the reasoning).
+  const hasActionableContent =
+    role !== "assistant" || visibleResponseContent(message).length > 0;
+
   if (completeDelete) return null;
 
   if (!!error) {
@@ -166,25 +174,27 @@ const HistoricalMessage = ({
             <HistoricalOutputs outputs={outputs} />
           </div>
         )}
-        <div className="flex items-start md:items-center gap-x-1">
-          <TTSMessage
-            slug={workspace?.slug}
-            chatId={chatId}
-            message={message}
-          />
-          <Actions
-            message={message}
-            feedbackScore={feedbackScore}
-            chatId={chatId}
-            slug={workspace?.slug}
-            isLastMessage={isLastMessage}
-            regenerateMessage={regenerateMessage}
-            isEditing={isEditing}
-            role={role}
-            forkThread={forkThread}
-            metrics={metrics}
-          />
-        </div>
+        {hasActionableContent && (
+          <div className="flex items-start md:items-center gap-x-1">
+            <TTSMessage
+              slug={workspace?.slug}
+              chatId={chatId}
+              message={message}
+            />
+            <Actions
+              message={message}
+              feedbackScore={feedbackScore}
+              chatId={chatId}
+              slug={workspace?.slug}
+              isLastMessage={isLastMessage}
+              regenerateMessage={regenerateMessage}
+              isEditing={isEditing}
+              role={role}
+              forkThread={forkThread}
+              metrics={metrics}
+            />
+          </div>
+        )}
         {role === "assistant" && <Citations sources={sources} />}
       </div>
     </div>
@@ -290,6 +300,29 @@ function TruncatableContent({ children }) {
       )}
     </>
   );
+}
+
+/**
+ * Returns the visible (non-reasoning) portion of an assistant message — i.e.
+ * what `RenderChatContent` would actually paint as the answer after thought
+ * blocks are collapsed into the ThoughtChainComponent. Used to decide whether
+ * a message is a pure reasoning bubble (no answer) and should skip the action bar.
+ * @param {string} message
+ * @returns {string} the message with `<think>...</think>` blocks removed, trimmed
+ */
+function visibleResponseContent(message = "") {
+  if (!message) return "";
+
+  // Open thought with no matching close = the whole message is reasoning
+  // (mid-stream or aborted), mirroring RenderChatContent's `msgToRender = ""`.
+  if (message.match(THOUGHT_REGEX_OPEN) && !message.match(THOUGHT_REGEX_CLOSE))
+    return "";
+
+  // Strip every complete thought block. The shared THOUGHT_REGEX_COMPLETE has
+  // no `g` flag (and is used with stateful `.match()` elsewhere), so build a
+  // fresh global copy here to catch the multiple-thought agent-loop case too.
+  const completeThoughts = new RegExp(THOUGHT_REGEX_COMPLETE.source, "g");
+  return message.replace(completeThoughts, "").trim();
 }
 
 const RenderChatContent = memo(
--- a/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
@@ -57,14 +57,6 @@ const HistoricalMessage = ({
   const isRefusalMessage =
     role === "assistant" && message === chatQueryRefusalResponse(workspace);
 
-  // Agent loop turns that reason and then call a tool emit a message whose
-  // entire content is a `<think>...</think>` block with no visible answer.
-  // Those bubbles have nothing to copy/speak/edit, so suppress the action bar
-  // (otherwise a hover copy button appears under each thought and copies an
-  // empty string once useCopyText strips the reasoning).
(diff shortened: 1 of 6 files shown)


shatfield4 (changes requested):
Seems to work just fine for me on the backend using tooled and untooled providers with reasoning enabled. Tested both Deepseek and Lemonade providers with thinking models in both regular chat and agent modes. Some small UI issues here with the copy button appearing under the thoughts UI (when pressing copy nothing is copied so we need to make sure we hide this in the UI or make it functional).


Image
Image

angelplusultra:
> Seems to work just fine for me on the backend using tooled and untooled providers with reasoning enabled. Tested both Deepseek and Lemonade providers with thinking models in both regular chat and agent modes. Some small UI issues here with the copy button appearing under the thoughts UI (when pressing copy nothing is copied so we need to make sure we hide this in the UI or make it functional).
> Image Image


@shatfield4 I was able to fix this in https://github.com/Mintplex-Labs/anything-llm/pull/5677/commits/bcfad04272b176e796fef8e79bfcf2f8c0e368e8

image

shatfield4 (approved):
Tested again and UI LGTM now.

timothycarambat:
@angelplusultra this seems to work well - however my only feedback is about the UI - I can understand wanting to break out thoughts but the very nature of thoughts is more like a debug log - they are ephemeral to understanding reasoning but we should collapse these into the `StatusResponses` we get from agent tools.

Otherwise, worse case on multi-step is we get a list of thoughts/status alternating until finally the user gets an answer - which is a lot of waste screenspace when all the user wants is the real output!

shatfield4 (changes requested):
Is fully functional and now stacks all thoughts/agent calls into the single thoughts/agent call container.

I am noticing we are not seeing the animation being played though when thinking is in progress. Can you trigger the animation so it plays the animation while thinking is in progress?

shatfield4 (approved):
LGTM

frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
```diff
@@ -57,6 +57,14 @@ const HistoricalMessage = ({
...
     role === "assistant" && message === chatQueryRefusalResponse(workspace);
 
+  // Agent loop turns that reason and then call a tool emit a message whose
+  // entire content is a `<think>...</think>` block with no visible answer.
+  // Those bubbles have nothing to copy/speak/edit, so suppress the action bar
+  // (otherwise a hover copy button appears under each thought and copies an
+  // empty string once useCopyText strips the reasoning).
+  const hasActionableContent =
+    role !== "assistant" || visibleResponseContent(message).length > 0;
+
```
timothycarambat: This should be handled by not adding this content to the chat history in  frontend/src/utils/chat/agent.js

With this chance, we are pruning the UI chat history in two different places. What makes it into `chatHistory` should be managed from agents.js

frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
```diff
@@ -166,25 +174,27 @@ const HistoricalMessage = ({
...
-            slug={workspace?.slug}
-            isLastMessage={isLastMessage}
-            regenerateMessage={regenerateMessage}
-            isEditing={isEditing}
-            role={role}
-            forkThread={forkThread}
-            metrics={metrics}
-          />
-        </div>
+        {hasActionableContent && (
```
timothycarambat: The main reason about handling data in agent is so we dont have to wrap this.

frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
```diff
@@ -292,6 +302,29 @@ function TruncatableContent({ children }) {
...
 
+/**
+ * Returns the visible (non-reasoning) portion of an assistant message — i.e.
+ * what `RenderChatContent` would actually paint as the answer after thought
+ * blocks are collapsed into the ThoughtChainComponent. Used to decide whether
+ * a message is a pure reasoning bubble (no answer) and should skip the action bar.
+ * @param {string} message
+ * @returns {string} the message with `<think>...</think>` blocks removed, trimmed
+ */
+function visibleResponseContent(message = "") {
```
timothycarambat: can be moved into agent.js if we can move this rendering logic to agent handler

timothycarambat (changes requested):
Let's look at the frontend changes here - I think we can simplify the diff here a lot. Feels like we are doing a lot of change for something that should, ideally, be smaller footprint

PR #5680, opened 2026-05-22
router-models: accurate token/message counting for model router decisions

timothycarambat:
### Description

Previously the router only counted tokens from the current user message, ignoring system prompt, chat history, pinned docs, and parsed files.

Introduce ModelRouterService.gatherRoutingContext() to compute the true conversation token count and reuse the fetched context downstream to avoid redundant DB/filesystem lookups.

Also adds model router support to the OpenAI-compatible and API chat handlers, propagates usageMetrics through ephemeral agents, and fixes the router input to use the attachment-injected message.

### DIFF
--- a/frontend/src/pages/GeneralSettings/LLMPreference/index.jsx
+++ b/frontend/src/pages/GeneralSettings/LLMPreference/index.jsx
@@ -434,6 +434,7 @@ export const AVAILABLE_LLM_PROVIDERS = [
       "GenericOpenAiTokenLimit",
       "GenericOpenAiKey",
     ],
+    connectionConfig: ["GenericOpenAiBasePath"],
   },
 ];
 
--- a/frontend/src/pages/GeneralSettings/ModelRouters/LLMProviderModelPicker/index.jsx
+++ b/frontend/src/pages/GeneralSettings/ModelRouters/LLMProviderModelPicker/index.jsx
@@ -50,8 +50,9 @@ export default function LLMProviderModelPicker({
   function isConfigured(providerValue) {
     if (!settings) return true;
     const llm = availableProviders.find((l) => l.value === providerValue);
-    if (!llm?.requiredConfig?.length) return true;
-    return llm.requiredConfig.every((key) => !!settings[key]);
+    const keys = llm?.connectionConfig || llm?.requiredConfig;
+    if (!keys?.length) return true;
+    return keys.every((key) => !!settings[key]);
   }
 
   useEffect(() => {
--- a/frontend/src/pages/GeneralSettings/ModelRouters/NewRouterModal/index.jsx
+++ b/frontend/src/pages/GeneralSettings/ModelRouters/NewRouterModal/index.jsx
@@ -168,11 +168,7 @@ export default function NewRouterModal({
               {loading ? (
                 <>
                   <CircleNotch className="h-4 w-4 animate-spin" />
-                  {t(
-                    isEdit
-                      ? "model-router.edit-router.saving"
-                      : "model-router.new-router.creating"
-                  )}
+                  {t("common.saving")}
                 </>
               ) : isEdit ? (
                 t("model-router.edit-router.save")
--- a/server/__tests__/utils/chats/openaiCompatible.test.js
+++ b/server/__tests__/utils/chats/openaiCompatible.test.js
@@ -1,7 +1,7 @@
 /* eslint-env jest, node */
 const { OpenAICompatibleChat } = require('../../../utils/chats/openaiCompatible');
 const { WorkspaceChats } = require('../../../models/workspaceChats');
-const { getVectorDbClass, getLLMProvider } = require('../../../utils/helpers');
+const { getVectorDbClass, getLLMProvider, resolveProviderConnector } = require('../../../utils/helpers');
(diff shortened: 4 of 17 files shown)


timothycarambat:
AI generated summary of changes because PR is so in depth...

### 1. Accurate token & message counting for router decisions (core fix)

Previously, `ModelRouterService` only counted tokens from the current user message when deciding which model to route to. This ignored the system prompt, chat history, pinned documents, and parsed files — meaning routing rules based on token thresholds were firing against a fraction of the real context size.

**New:** `ModelRouterService.gatherRoutingContext()` assembles the full conversation context (system prompt + history + pinned docs + parsed files + current message), counts tokens across all of it, and queries the true DB message count (+1 for the in-flight message). It returns all fetched data so downstream code can reuse it without redundant DB/filesystem lookups.

### 2. Centralized provider resolution via `resolveProviderConnector()`

Every chat handler had its own copy of the "is this a router provider? if so, instantiate router, count tokens, resolve" logic. This is now consolidated into a single `resolveProviderConnector()` in `server/utils/helpers/index.js`.

It handles:
- Non-router providers (passthrough to `getLLMProvider`)
- Router providers (instantiates the router, calls `gatherRoutingContext`, resolves the route)
- Returns `{ connector, routingMetadata, prefetchedContext }` so callers get everything they need

A deprecation notice was added to `getLLMProvider` directing callers to use `resolveProviderConnector` instead.

### 3. Router support added to all chat surfaces

Every entry point that previously called `getLLMProvider` directly now goes through `resolveProviderConnector`:

| Surface | File |
|---|---|
| **Main workspace chat** | `stream.js` — `resolveLLMConnector` now delegates to `resolveProviderConnector`; reuses prefetched context for history, pinned docs, parsed files, and system prompt |
| **API chat handler** | `apiChatHandler.js` — both `chatSync` and `streamChat` |
| **OpenAI-compatible** | `openaiCompatible.js` — both sync and stream, with `messageCountOverride` set to `history.length` (no +1 since OAI history already ends with the user message) |
| **Embed widget** | `embed.js` — `resolveLLMConnectorForEmbed` refactored to use `resolveProviderConnector`, passing embed-specific chat history and message count |
| **Telegram bot** | `telegramBot/chat/stream.js` |
| **API similarity search** | `endpoints/api/workspace/index.js` — the `/search` endpoint |
| **Agent memory plugin** | `agents/aibitat/plugins/memory.js` — vector search inside agent tool calls |

### 4. Pre-fetched context reuse to avoid redundant lookups

When the router is active, `gatherRoutingContext` already fetches chat history, pinned docs, parsed files, and builds the system prompt. This prefetched data is now threaded through to the chat handlers so they skip re-fetching:

- `stream.js` — reuses `rawHistory`, `chatHistory`, `pinnedDocs`, `parsedFiles`, and `systemPrompt` from the prefetch
- `embed.js` — reuses `rawHistory`, `chatHistory`, and `pinnedDocs`

For non-router paths, `prefetchedContext` is `null` and the existing fetch logic runs as before.

### 5. Router input now uses the attachment-injected message

In `stream.js`, the router was being called with the raw `message` before attachment content was injected. It now receives `updatedMessage` (the post-attachment-injection text), so token counts reflect what actually gets sent to the LLM.

### 6. `usageMetrics` propagation through ephemeral agents

`EphemeralEventListener` now understands `reportStreamEvent` messages with types `textResponseChunk`, `fullTextResponse`, and `usageMetrics`. These are:

- Forwarded as proper `writeResponseChunk` calls during streaming (so clients get incremental text and usage data)
- Collected in `packMessages()` which now returns `{ thoughts, textResponse, outputs, metrics }` instead of just the first three

Both `apiChatHandler.js` sync and stream paths capture the new `metrics` field and persist it alongside chat records.

### 7. Correct message counting for the `/chat` API

The current in-flight message is now included in the conversation message count (`+1`) so that a routing rule like ">=3 messages" fires on the user's 3rd message, not the 4th.

### 8. Smaller fixes

- **OpenAI `/v1/models` endpoint** — no longer instantiates an LLM provider per workspace just to build the `owned_by` field; uses the provider string directly
- **`BaseLLMProvider` JSDoc** — added `className`, `model`, and `defaultTemp` properties; removed "from OpenAI" wording from method descriptions
- **Frontend `NewRouterModal`** — minor UI adjustments (6 lines changed)
- **Tests** — `openaiCompatible.test.js` updated to mock `resolveProviderConnector` alongside the existing `getLLMProvider` mock

Test suite (manually confirmed bc I am insane)
```
Router Fix

- [ ] With Router
    - [x] Test Conversation Count
        - [x] Agent
        - [x] Chat
        - [x] Query
        - [x] API
            - [x] Stream
            - [x] Chat
            - [x] OpenAI (chat/stream)
        - [x] Embed
            - [x] Resetting embed will then reset message count
    - [ ] Test Conversation Token Count
        - [x] Agent
        - [x] Chat
        - [x] Query
        - [x] API
            - [x] Stream
            - [x] Chat
            - [x] OpenAI
        - [x] Telegram
        - [x] Embed
- [ ] Without Router
        - [x] Agent
        - [x] Chat
        - [x] Query
        - [x] API
            - [x] Stream
            - [x] Chat
            - [x] OpenAI (chat/stream)
        - [x] Telegram
        - [x] Embed
            - [x] Resetting embed will then reset message count

```

PR #5703, opened 2026-05-25
docs: clarify text chunking strategy settings

narutamaaurum:
## Summary
- explain that the chunking preferences page currently uses LangChain's `RecursiveCharacterTextSplitter`
- link the settings copy directly to the upstream strategy docs
- keep the change narrowly scoped to the existing admin settings page

## Validation
- `git diff --check`
- attempted `node frontend/src/locales/verifyTranslations.mjs`, but this runner only has Node v12.22.9 and the repo now expects a newer Node runtime (`Intl.DisplayNames` is unavailable here)

Closes Mintplex-Labs/anythingllm-docs#257

### DIFF
--- a/frontend/src/locales/en/common.js
+++ b/frontend/src/locales/en/common.js
@@ -1130,7 +1130,7 @@ const TRANSLATIONS = {
     "desc-start":
       "Sometimes, you may want to change the default way that new documents are split and chunked before being inserted into your vector database.",
     "desc-end":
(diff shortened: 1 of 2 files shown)


timothycarambat:
As mentioned in the issue [(now transferred)](https://github.com/Mintplex-Labs/anythingllm-docs/issues/257) - this is better suited for the actual docs repo

PR #5705, opened 2026-05-26
fix: respect IME composition in prompt input

RitwijParmar:
## Summary
- Guard prompt keyboard shortcuts while an IME composition is active, including Enter/keyCode 229 events from Android/Japanese keyboards.
- Track composition start/end on the chat textarea and sync the final composed value/height after conversion completes.
- Avoid adding undo snapshots while composition is in progress so the prompt does not interfere with conversion state.

## Why
Issue #5696 reports Japanese IME input as unusable on Android. The prompt input currently handles Enter and tool-menu shortcuts before checking whether the key event belongs to an active IME composition, which can turn conversion/confirmation keystrokes into submit/menu behavior.

This PR addresses the Japanese IME input portion of #5696. It does not claim to fix the separate Android MCP tools issue from the same report.

## Testing
- `yarn eslint src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx`
- `yarn build`

### DIFF
--- a/frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
@@ -54,6 +54,7 @@ export default function PromptInput({
   const formRef = useRef(null);
   const textareaRef = useRef(null);
   const [_, setFocused] = useState(false);
+  const isComposingRef = useRef(false);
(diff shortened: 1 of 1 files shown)


timothycarambat:
This is not the repo for the android app! The android app is a totally separate repo which is going to be OSS soon!

PR #5719, opened 2026-05-29
Improve agent summarizer tool

### LINKED ISSUE #5718: [FEAT]: Rework summarize tool

The summarize tool tends to lock up on small local models. We should remove langchain from this tool and do it in a simpler way where we chunk it and prompt the user after 3 chunks if they would like to continue. This will greatly improve the UX on smaller context window models.

When entering the summary loop:
- Get model context window (if needed again)
- Chunk the content to be 45% of context window of the model
- Send these chunks one at a time asking ONLY for the key points of this section
- Include the previous points in the next loop iteration, but previous points should not make up more than 5% of the window. If it exceeds that truncate the content from the start so that the details of the last chunk summary are appended to the next chunk in case we split in the middle of a piece of content.
- After 3 chunk loops, before doing chunk 4 if it exists - ask the user if it is OK to continue or if it should skip
- We can count details/concepts from non-empty newlines?

User Experience:
- I ask model to do some deep research in an on-demand chat (regular chat)
- Model finds some VERY verbose website or content
- Event sent to UI that this content is very long and will need to be summarized
- chunk logic applied to content -> start processing chunk by chunk
- if chunks <4 model finishes and returns its bullet list from the tool
- if more:
    - Model returns to user a `requestToolApproval` event asking if it is okay to continue
    - Inform the user via this action "There are x more pieces of content I can summarize - Continue?"
    - If they say yes -> continue again without asking through the whole document
    - If they deny -> return the content as it is in its current state.

This should prevent this tool from needlessly locking up the whole UI because it comes across some large content.

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5718 

### Description

- Greatly improves UX of summarizing with small context window models
- Removes LangChain deps from the summarize agent tool
- Rewrites summarization to walk content one chunk at a time instead of the old map_reduce chain that never converged on small windows
- Chunks content to 45% of the model's context window using the model tokenizer _(see note in additional information section about other percentages tested)_
- Asks the model for only the key points of each section and accumulates them
- Carries prior key points into the next chunk for continuity, capped at 5% of the window (truncated from the start so the latest details are kept)
- Asks the user to continue after the first 3 sections so a long document never silently locks up the chat
- Approving continues to the end without asking again, denying returns the key points gathered so far
- Reports progress to the UI as each section is summarized

### Additional Information

- I also tested `CHUNK_CONTEXT_RATIO` @ `0.5`, `0.55`, and `0.6` with no issues at all and would allow local models to be more performant (since less overall chunks) but we do run the risk of getting close to the actual context window limit the higher we make this percentage
- Side note: tested a local version of https://github.com/Mintplex-Labs/anything-llm/pull/5548 merged with this PR (#5719) and it works much better than the current HTML scraper version we have
- If combined with the markdown scraper PR, it removes all the css and js artifacts and the text is much cleaner to parse resulting is much less overall chunks (and more accurate links)
- Chunks are much smaller and using a local model is still slow but much more bearable than the current implementation we have (no more infinite looping issue due to small context windows)

### Developer Validations

Tested on:
- https://en.wikipedia.org/wiki/Artificial_intelligence
- https://anythingllm.com/
- https://github.com/Mintplex-Labs/anything-llm


timothycarambat:
@shatfield4 can you list some websites or things you tested this on for long-form summaries?

server/utils/agents/aibitat/plugins/summarize.js
```diff
@@ -170,6 +170,15 @@ const docSummarizer = {
...
                 content: document.content,
+                introspect: (message) =>
+                  this.super.introspect(`${this.caller}: ${message}`),
+                requestApproval: this.super.requestToolApproval
+                  ? (description) =>
+                      this.super.requestToolApproval({
+                        skillName: this.name,
+                        description,
+                      })
+                  : null,
```
timothycarambat: Why is this on the plugin?

You can reach approval like this:
```js
 if (this.super.requestToolApproval) {
                const approval = await this.super.requestToolApproval({
                  skillName: this.name,
                  payload: { filename },
                  description: `Create PDF document "${filename}"`,
                });
                if (!approval.approved) {
                  this.super.introspect(
                    `${this.caller}: User rejected the ${this.name} request.`
                  );
                  return approval.message;
                }
              }
```

and for introspection:
```js
this.super.introspect(
                `${this.caller}: Creating PDF document "${filename}"`
              );
```

Same Q for server/utils/agents/aibitat/plugins/web-scraping.js - why is this change required?

server/utils/agentFlows/executors/web-scraping.js
```diff
@@ -67,6 +67,14 @@ async function executeWebScraping(config, context) {
...
     model: aibitat.defaultProvider.model,
     content,
+    introspect,
+    requestApproval: aibitat.requestToolApproval
+      ? (description) =>
+          aibitat.requestToolApproval({
+            skillName: "web-scraping",
+            description,
+          })
+      : null,
```
timothycarambat: AgentFlow needs a change for this too?

server/utils/agents/aibitat/plugins/summarize.js
shatfield4: Refactored to pass aibitat object instead to make the implementation simpler. We do need to pass the aibitat object to the `summarizeContent` util so that we can do the approval inside of this util.

server/utils/agentFlows/executors/web-scraping.js
shatfield4: Same as the regular tool and yes, it is required in the agent flow executor as well so that if the context overflows when scraping a site in an agent flow, it will still prompt the user after 3 summary chunks if they would like to continue with the summarization. I did simplify this to just pass the aibitat object now though to avoid having to pass `introspect` and `requestToolApproval` individually.

shatfield4:
> @shatfield4 can you list some websites or things you tested this on for long-form summaries?

Updated description with links tested on and a side note about how this works with the markdown scraper PR.

PR #5721, opened 2026-05-29
apply universal sdk timeout

### LINKED ISSUE #5603: [BUG]:  LMStudio Agent loop socket timeout

### LINKED ISSUE #5631: [FEAT]: Custom Timeout Values for Socket Connections

### LINKED ISSUE #5681: OpenAI and Anthropic provider calls should use explicit timeouts and retry/backoff

timothycarambat:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5631
resolves #5603
resolves #5681
closes #5637

### Description

- Adds monkeypatch to `openai` and `anthropic` sdk layers to use a system level fetch `undici` Agent so the user can set a custom timeout.
- Base timeout bumped to 10min
- Removes `maxRetries` and sets them globally to 0 to get socket timeouts back sooner.
- Adds these ENVs to `updateENV` for persistence.

### Developer Validations


osabdo77-max:
Hi Timothy,

First of all, thank you for the work on this issue and for the improvements introduced in version 1.14.0.

After testing the new release, I can confirm that there is a noticeable improvement compared to previous versions.

One of the biggest improvements I noticed is with Agent Skills. In older versions, socket timeout issues could occur very early, sometimes even on the first response or during small and medium-sized tasks. In version 1.14.0, these cases have improved significantly, and the agent is much more stable during normal workloads.

However, I have noticed that the issue can still occur in some specific scenarios. The remaining cases seem to be mostly related to heavy workloads, large local models, long RAG workflows, or the very first Agent Skills task immediately after loading a model. Since the model is still initializing and loading resources at that stage, the first request can take significantly longer and may still trigger a timeout.

As a possible future enhancement, would it make sense to allow users to configure a custom timeout value? I believe this could be especially helpful for users with mid-range hardware, CPU-heavy inference, model offloading, or large local AI workloads where processing times can vary greatly.

Overall, version 1.14.0 is a noticeable improvement, and I appreciate the work that has gone into addressing this issue.

Thank you again for your work and for continuing to improve AnythingLLM.

PR #5723, opened 2026-05-30
fix: avoid double-encoding SearXNG search queries

### LINKED ISSUE #5722: SearXNG search query is double encoded
SearXNG searches seem to be double encoding the query.

Example query:

```text
Colossus and the Crab ending D.F. Jones book 3 plot summary
```

gets sent as:

```text
/search?q=Colossus%2520and%2520the%2520Crab%2520ending%2520D.F.%2520Jones%2520book%25203%2520plot%2520summary&format=json
```

So SearXNG receives `Colossus%20and%20...` as the query instead of normal text.

Looks like this line is the cause:

```js
searchURL.searchParams.append("q", encodeURIComponent(query));
```

`URLSearchParams` already handles encoding, so passing `query` directly fixes it:

```js
searchURL.searchParams.append("q", query);
```

I monkeypatched this locally and it resolves the issue, but I’d rather not carry that patch in my deployment.

trick77:
### Pull Request Type

- fix (Bug fix)

### Relevant Issues

resolves #5722

### Description

SearXNG search queries were being passed through `encodeURIComponent` before being appended to `URLSearchParams`. Since `URLSearchParams` encodes values already, this produced double-encoded query strings like `%2520`.

This passes the raw query to `URLSearchParams` and adds a regression test for the SearXNG URL.

### Visuals (if applicable)

N/A

### Developer Validations

Focused validation run:

```sh
npx --yes jest server/__tests__/utils/agents/aibitat/plugins/web-browsing.test.js --runInBand
```


timothycarambat:
Removed test since it was basically only testing the functionality of a native NodeJS util.

PR #5747, opened 2026-06-03
feat: make document sync stale-after interval configurable

### LINKED ISSUE #3988: [FEAT]: Make defaultStaleAfter configurable for document sync frequency
Currently, the defaultStaleAfter value for auto document sync is hardcoded to 604800000 (7 days) in [documentSyncQueue.js](https://github.com/Mintplex-Labs/anything-llm/blob/master/server/models/documentSyncQueue.js#L21). While this works well for most cases, it would be extremely helpful if this value could be configured through an environment variable or workspace setting.

In some use cases (like syncing frequently changing operational documents), a daily sync (1 day) is much more suitable than 7 days. Having this as a configurable value would:
	•	Provide more flexibility for users with different document update frequencies
	•	Avoid modifying core source code for self-hosted deployments
	•	Support more real-time knowledge base scenarios

Would you consider making defaultStaleAfter configurable via .env, a global setting, or workspace-level override?

Thanks for the great work on this project.

sanidhyasin:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #3988

### Description

The document sync/watch background worker hardcoded its stale-after interval to 7 days (`604800000ms`) in `server/models/documentSyncQueue.js`. As noted in #3988, this is too infrequent for operational documents that change often, and self-hosters had no way to tune it short of patching the source.

This PR makes the interval configurable via a new optional `DOCUMENT_SYNC_STALE_AFTER_MS` environment variable, keeping the existing 7-day behavior as the default.

**Changes**

- `defaultStaleAfter` is now a getter that reads `DOCUMENT_SYNC_STALE_AFTER_MS`:
  - falls back to the 7-day default when the value is unset, non-numeric, or non-positive;
  - enforces a **1-hour minimum** to guard against overloading embedders by re-syncing too frequently (the concern raised by @timothycarambat in the issue thread).
- `watch()` now persists `staleAfterMs` on the queue record at creation time, so the configured interval governs **recurring** syncs (via `calcNextSync`), not just the first one. Previously recurring syncs always fell back to the Prisma column default of 7 days.
- Documented the variable in `server/.env.example` and `docker/.env.example`.

### Additional Information

This follows the existing env-var pattern used for `COMETAPI_LLM_TIMEOUT_MS` (parse → validate → clamp → default).

### Developer Validations

> Testing notes: Added `server/__tests__/models/documentSyncQueue.test.js` covering the default, valid override, sub-minimum clamping, and invalid/non-positive fallbacks. The clamping/fallback logic was verified locally; a full monorepo install for the complete jest/lint toolchain was skipped as heavy.


timothycarambat:
Great PR 👍

sanidhyasin:
Thanks @timothycarambat! Glad it fit, appreciate the quick review. 🙏

PR #5751, opened 2026-06-04
feat: native tool calling on by default, add PROVIDER_DISABLE_NATIVE_TOOL_CALLING

### LINKED ISSUE #5640: [FEAT]: detect native tool support for Generic OpenAI-compatible endpoints before falling back to UnTooled
## Summary

Generic OpenAI currently appears to require manual opt-in via:

```env
PROVIDER_SUPPORTS_NATIVE_TOOL_CALLING="generic-openai"
```

before AnythingLLM uses native tool calling. If the flag is absent, Generic OpenAI falls back to `UnTooled`.

It would be useful for AnythingLLM to detect or probe native tool support for OpenAI-compatible endpoints before falling back.

## Use Case

Many OpenAI-compatible routers/proxies expose model capability metadata or support standard `tools` parameters even though they are configured through AnythingLLM as `generic-openai`.

Example endpoint behavior:

```json
{
  "id": "cx/gpt-5.4-mini",
  "capabilities": {
    "tool_calling": true,
    "reasoning": true
  }
}
```

In this setup, the endpoint and selected model support tools, but AnythingLLM still uses `UnTooled` unless the operator manually sets the env flag.

## Observed Behavior

With AnythingLLM `1.12.1`:

```env
LLM_PROVIDER='generic-openai'
GENERIC_OPEN_AI_MODEL_PREF='cx/gpt-5.4-mini'
PROVIDER_SUPPORTS_NATIVE_TOOL_CALLING=<unset>
```

logs show:

```text
Generic OpenAI supports native tool calling is DISABLED. Will use UnTooled instead.
```

The endpoint's `/v1/models` response advertises `capabilities.tool_calling: true`, but AnythingLLM does not appear to use this metadata for Generic OpenAI capability detection.

After setting:

```env
PROVIDER_SUPPORTS_NATIVE_TOOL_CALLING="generic-openai"
```

logs show:

```text
Generic OpenAI supports native tool calling is ENABLED.
Provider.stream (tooled) - will process this chat completion.
```

and native tool calls work.

## Proposed Behavior

Before falling back to `UnTooled`, Generic OpenAI could attempt one or more of:

- Read `/v1/models` for the selected model and check common capability fields:
  - `capabilities.tool_calling`
  - `capabilities.tools`
  - `tool_calling`
  - `supports_tools`
- Optionally perform a small non-destructive tool-call probe against the configured model.
- Cache the detected capability per provider/base URL/model.
- Keep the env var as an explicit override:
  - env true: force native tool calling
  - env false/absent: auto-detect if possible
  - detected unsupported: fall back to `UnTooled`

## Safety / Compatibility

The current conservative behavior is understandable because not every OpenAI-compatible endpoint supports tools. The enhancement could preserve that safety by:

- Keeping `UnTooled` fallback when detection fails.
- Logging the detection result clearly.
- Allowing operators to force-enable or force-disable via env.
- Avoiding probes unless explicitly enabled, if runtime probes are considered too invasive.

## Why This Matters

Without auto-detection, users of OpenAI-compatible routers can end up with degraded agent behavior even when their endpoint and model support native tools.

In one real case, this caused Gmail agent behavior where:

- `gmail-search` was invoked repeatedly.
- `gmail-read-thread` was never invoked.
- The agent claimed it could not access/read Gmail message bodies.
- Setting `PROVIDER_SUPPORTS_NATIVE_TOOL_CALLING="generic-openai"` fixed the issue.

This makes the current fallback difficult to diagnose and can look like a connector, network, or model refusal issue rather than a provider capability configuration issue.

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5640
docs: https://github.com/Mintplex-Labs/anythingllm-docs/pull/261

### Description

Native tool calling used to be opt-in per provider via `PROVIDER_SUPPORTS_NATIVE_TOOL_CALLING`. This flips that: native tool calling is now always enabled by default, and a new `PROVIDER_DISABLE_NATIVE_TOOL_CALLING` flag lets you force a provider back to UnTooled (prompt-based) tool calling.

Affected providers (the ones that leaned on the old flag): bedrock, groq, lemonade, litellm, localai, openrouter, and generic-openai.

Changes:

- Renamed the base helper `Provider.supportsNativeToolCallingViaEnv()` to `optsOutOfNativeToolCallingViaEnv()` in `ai-provider.js`, now keyed off `PROVIDER_DISABLE_NATIVE_TOOL_CALLING`.
- Each provider's `supportsNativeToolCalling()` now returns `true` unless the provider tag is listed in the disable flag. This keeps it as the single source of truth, so the `stream`/`complete` paths and all external consumers (`Workspace.supportsNativeToolCalling`, `isAgentCommandAvailable`, agent handlers, telegram bot) keep working.
- Lemonade: dropped its label/capability fallback for enabling tools, since native is now on by default unless explicitly disabled.
- generic-openai: its `tools` capability in `getModelCapabilities()` now defaults to enabled unless disabled via the new flag. The other capabilities (reasoning, vision, image generation) keep their existing `PROVIDER_SUPPORTS_*` opt-in behavior.
- Updated the env allow-list in `updateENV.js` and both `.env.example` files to the new flag + semantics.

### Behavior change to be aware of

Because native tool calling is now on by default, workspaces in **automatic** chat mode will auto-run the agent for these providers without needing `@agent`. Setting `PROVIDER_DISABLE_NATIVE_TOOL_CALLING` for a provider reverts it to the previous default, where the agent runs only when `@agent` is explicitly invoked.

### Additional Information

No frontend or DB changes. The `PROVIDER_DISABLE_NATIVE_TOOL_CALLING` value is a comma-separated list of provider tags, e.g. `PROVIDER_DISABLE_NATIVE_TOOL_CALLING="openrouter,groq"`.

### Developer Validations


shatfield4 (changes requested):
Small nit:
- In `server/utils/agents/aibitat/providers/genericOpenAi.js` lines 54-55 has a stale comment about the env and how default is false for supporting native tool calling
- Update the comment to reflect the new behavior

Other than that, LGTM. I was able to test it groq, lemonade, and openrouter. I also wrote a simple test script to call `optsOutOfNativeToolCallingViaEnv()` and `supportsNativeToolCalling()` to ensure we get expected results out of each provider.

timothycarambat:
@angelplusultra This also needs an associated documentation PR - there is a whole section about this behavior and now that it wont be an issue it should really be a notice about how if you are using a non-tool-calling enabled model you may need to set this flag.

Clearly, the likelyhood of someone using a non-tooled model is lower since that is basically the standard now.

server/utils/AiProviders/genericOpenAi/index.js
```diff
@@ -442,14 +441,30 @@ class GenericOpenAiLLM {
...
+   * @returns {boolean}
+   */
+  #optsOutOfNativeToolCallingViaEnv() {
+    if (!("PROVIDER_DISABLE_NATIVE_TOOL_CALLING" in process.env)) return false;
+    return (
+      process.env.PROVIDER_DISABLE_NATIVE_TOOL_CALLING?.includes(
+        "generic-openai"
+      ) || false
+    );
+  }
```
timothycarambat, line 458: This function can be rolled up to a local code path inside of  `supportsCapabilityFromENV` to just check for the lack of the ENV - this does not need its own method since all the other ENVs share the same method.

timothycarambat (changes requested):
This needs to be reworked as it missed the key point of the PR/issue. It does function in the capacity of having env blocked providers now being unblocked, but it misses all the other providers we did not extend this flag to.

Example: Cohere is still disabled with this change: https://github.com/Mintplex-Labs/anything-llm/blob/feat-disable-native-tool-calling-env-var/server/utils/agents/aibitat/providers/cohere.js#L40

AI provider sets the class method to always return false. It then must be overriden by the implementer to enable it or add additional logic:
https://github.com/Mintplex-Labs/anything-llm/blob/feat-disable-native-tool-calling-env-var/server/utils/agents/aibitat/providers/ai-provider.js#L145-L151
^IMO this method is the one that should be overridden to use the ENV and the provider tag if we can get it from this method.

This should now be true and then through test if we need to override a specific class because we cannot support tools that should be false with a comment as to why - the only acceptable reason being the API does not support tool calls.

PR #5762, opened 2026-06-06
Minify Generic OpenAI request bodies for better server compatibility

### LINKED ISSUE #5752: [BUG]: Prompts to local Jan endpoint unresponsive

AnythingLLM desktop app

### What happened?

Prompting local OpenAI-compatible Jan endpoint no longer works in **AnythingLLM Desktop v1.13.0**.

I can confirm that my local endpoint works fine, I run [aider v0.86.2](https://aider.chat/) using the same endpoint and there are no problems. This is hosted via [Jan v0.8.2](https://www.jan.ai/) using the llama.cpp backend. I can also confirm that the local endpoint received the request from AnythingLLM via the logs, same with aider. It seems like AnythingLLM does not respond or do anything after sending the request.

Image

**Edit:** Another possible explanation is that this bug is caused by Jan. Since I also updated Jan to the latest version, maybe the way it exposes its API endpoint in the new version caused AnythingLLM to break. But that's strange considering aider currently still works. I tried doing a fresh install of AnythingLLM v1.13.0, v1.12.1, v1.12.0, deleting all user data, and the problem still persists, which is very weird. So it seems like it's not a regression bug. Maybe AnythingLLM expects a strict response that Jan no longer follows? Maybe my OS has some problems? I have no idea.

> [!NOTE]
> The fact that aider works but AnythingLLM doesn't, qualifies as a bug worth looking into in my opinion.

1. Start Jan server.
2. Configure "Generic OpenAI" provider in AnythingLLM to point to Jan local endpoint (e.g. `http://127.0.0.1:1337/v1`)

Image

3. Enter a prompt in the chat.

Image

Dots appear with no response forever.

AnythingLLM Version: v1.13.0
Model: Qwen3_6-35B-A3B-UD-IQ4_XS
Backend: llama.cpp
API Server: Jan v0.8.2
OS: Windows 11

_No response_

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5752 

### Description

- Chats and agents sent to a Generic OpenAI endpoint would hang forever with no response, no error, and no timeout — even though the endpoint was healthy and `curl`/other clients got normal responses from the same URL.
- Root cause: the `openai` Node SDK serializes every request body as pretty-printed JSON (`JSON.stringify(body, null, 2)`), so the body contains literal newlines, and some OpenAI-compatible servers mishandle that whitespace and never respond.
- Confirmed by replaying the exact request with `curl`: a compact single-line body streamed instantly, while the same body pretty-printed with newlines hung indefinitely.
- This is a server-side regression, not ours — the SDK has always pretty-printed bodies (verified `4.38.5` and `4.95.1` behave identically). It surfaced on Jan after the unified llama.cpp router process introduced in v0.8.0, which the reporter's v0.8.2 inherits.
- Fix: a small `fetch` wrapper on the Generic OpenAI clients (chat + agent paths) re-serializes string JSON request bodies compactly before sending. Minified JSON is a lossless round-trip and the more universally compatible form on the wire, so it works across every backend the provider targets.
- Safe for other providers: the transform only removes formatting whitespace — values, string contents (including `\n`/`\t` inside messages), escapes, and numbers stay byte-identical to what the SDK already produced. Only string JSON bodies are touched; multipart/binary and unparseable bodies pass through untouched.
- Scoped to the Generic OpenAI provider's clients only — OpenAI, Anthropic, Groq, Together, etc. build their own clients that are not modified and keep sending requests exactly as before.
- Jan regression reference: [v0.8.0 changelog](https://www.jan.ai/changelog/2026-05-22-jan-v0.8.0) (router process introduced), [v0.8.0 release](https://github.com/janhq/jan/releases/tag/v0.8.0), [v0.8.2 release](https://github.com/janhq/jan/releases/tag/v0.8.2) (version reported in #5752).

### Developer Validations

### DIFF
--- a/server/utils/AiProviders/genericOpenAi/index.js
+++ b/server/utils/AiProviders/genericOpenAi/index.js
@@ -24,6 +24,7 @@ class GenericOpenAiLLM {
     this.openai = new OpenAIApi({
       baseURL: this.basePath,
       apiKey: process.env.GENERIC_OPEN_AI_API_KEY ?? null,
+      fetch: GenericOpenAiLLM.compactJsonFetch,
       defaultHeaders: {
         "User-Agent": getAnythingLLMUserAgent(),
         ...GenericOpenAiLLM.parseCustomHeaders(),
@@ -74,6 +75,25 @@ class GenericOpenAiLLM {
     }
 
     return headers;
+  }
+
+  /**
+   * A `fetch` that minifies JSON request bodies before sending. The OpenAI SDK
+   * pretty-prints bodies (`JSON.stringify(body, null, 2)`), but compact JSON is a
+   * lossless round-trip and the more universally compatible form across backends.
+   * Only string JSON bodies are touched; multipart/binary and unparseable bodies pass through as-is.
+   * @param {string|URL} url - the request url
+   * @param {Object} [init] - the fetch init options
+   * @returns {Promise<Response>}
+   */
+  static compactJsonFetch(url, init = {}) {
+    const { fetch: openAiFetch } = require("openai/_shims/index.js");
+    if (init && typeof init.body === "string") {
+      try {
+        init.body = JSON.stringify(JSON.parse(init.body));
+      } catch {}
+    }
+    return openAiFetch(url, init);
   }
 
   #appendContext(contextTexts = []) {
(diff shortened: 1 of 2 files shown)


angelplusultra (approved):
LGTM.

I would argue this isn't our problem to deal with and adds complexity to the codebase to appease an extremely small subset of users. Which even if JanAI fixes this problem themselves how would even know to revert this back to normal? We would have to keep an eye out on their release notes? Doesn't seem worth it to me.

shatfield4:
@timothycarambat @angelplusultra and I discussed this and we'll be closing this PR/issue. This is more of a JanAI problem not being fully OpenAI compatible.

We did test this on multiple providers and it did not surface any bugs but the side effects it may cause is not worth fixing because it may break support for providers other users may be using.

PR #5795, opened 2026-06-10
feat: expose requestToolApproval to custom agent skills

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #

### Description

Built-in agent tools (gmail send/reply, filesystem writes, calendar edits, etc.) can pause the agent and show an Approve/Reject card before performing a destructive action. They do this by calling `requestToolApproval`, a method that lives on the live aibitat instance (attached by the websocket plugin).

Custom skills (user-authored imported plugins — a `plugin.json` + `handler.js`, usually installed from the Community Hub) had no sanctioned way to do the same. They already get first-class helpers like `this.introspect` and `this.logger` injected into their handler context, but not an approval helper.

This PR exposes `requestToolApproval` as a first-class helper on the custom-skill handler context, so a skill's `handler.js` can ask the user to confirm its own potentially destructive actions:

```js
const approval = await this.requestToolApproval({
  payload: { recordId },
  description: `Permanently delete record ${recordId}? This cannot be undone.`,
});
if (!approval.approved) return approval.message;
```

It returns the same `{ approved, message }` contract the built-in tools use and triggers the same Approve/Reject card. The frontend card, the "Always allow" whitelist, the `AGENT_AUTO_APPROVED_SKILLS` auto-approve, and the 120s timeout are all reused unchanged — there are no frontend or websocket changes.

Key behaviors:

- `skillName` is injected automatically as the skill's own hubId. The author only passes `payload` and `description`, so a skill cannot spoof another tool's name to ride that tool's whitelist and bypass its own approval.
- The helper is defined after the skill's own handler exports are spread in, so a skill's `handler.js` cannot override it.
- If no approval channel exists (e.g. a scheduled job, where there is no human to approve), the helper resolves approved so the skill still runs — matching how the built-in tools fall through their approval guard when the method is absent.

### Additional Information

The capability was technically reachable before this PR via `this.super.requestToolApproval`, but that was undocumented internal plumbing and unsafe to hand to skill authors directly (no skillName injection, overridable by the skill). This PR makes it a sanctioned, documented, and safe helper that sits alongside the existing `introspect`/`logger` helpers. The new code is roughly 15 lines in `server/utils/agents/imported.js`, plus a new unit test file.

How to test:

1. Create a custom skill under `server/storage/plugins/agent-skills/<hubId>/` whose `handler.js` calls `this.requestToolApproval({ description: "Confirm test action?" })` and returns the result.
2. Invoke it from an agent session in a workspace.
3. Confirm the Approve/Reject card appears, that Reject returns the rejection message to the chat, and that checking "Always allow" then re-invoking skips the card (the whitelist is keyed on the skill's hubId).

Follow-up (out of scope here): the external Community Hub skill-authoring docs should document `this.requestToolApproval(...)`.

### Developer Validations

### DIFF
--- a/server/utils/agents/imported.js
+++ b/server/utils/agents/imported.js
@@ -178,6 +178,7 @@ class ImportedPlugin {
       name: this.name,
       config: this.config,
       setup(aibitat) {
+        const skillName = this.name; // hubId the skill is registered + whitelisted under.
         aibitat.function({
           super: aibitat,
           name: this.name,
@@ -196,6 +197,39 @@ class ImportedPlugin {
             additionalProperties: false,
           },
           ...customFunctions,
+          /**
+           * Pause the agent and ask the user to approve a potentially
+           * destructive action before the skill performs it. Shows the same
+           * Approve/Reject card the built-in tools use (e.g. gmail send/reply).
+           *
+           * `skillName` is supplied automatically (the skill's hubId) so a skill
+           * cannot spoof another tool's name to bypass its own approval - the
+           * author only provides the payload and description.
+           *
+           * Defined after `...customFunctions` so a skill's handler.js cannot
+           * override it. When no approval channel exists (e.g. scheduled jobs)
+           * it resolves approved so the skill still runs, matching how built-in
+           * tools fall through their approval guard.
+           *
+           * @param {{payload?: object, description?: string}} options
+           * @returns {Promise<{approved: boolean, message: string}>}
+           */
+          requestToolApproval: async ({
+            payload = {},
+            description = null,
+          } = {}) => {
+            if (typeof aibitat.requestToolApproval !== "function") {
+              return {
+                approved: true,
+                message: "Approval not required in this context.",
+              };
+            }
+            return aibitat.requestToolApproval({
+              skillName,
+              payload,
+              description,
+            });
+          },
         });
       },
     };
--- a/server/utils/agents/imported.js
+++ b/server/utils/agents/imported.js
@@ -154,6 +154,43 @@ class ImportedPlugin {
     return this.isValidLocation(handlerLocation);
   }
 
+  /**
+   * Creates a requestToolApproval function bound to a specific aibitat
+   * instance and skill name via closure — no `this` binding required.
+   *
+   * The returned function pauses the agent and asks the user to approve a
+   * potentially destructive action before the skill performs it.
+   *
+   * `skillName` is captured from the skill's hubId so a skill cannot spoof
+   * another tool's name to bypass its own approval.
+   *
+   * When no approval channel exists (e.g. scheduled jobs) it resolves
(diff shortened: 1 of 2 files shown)


shatfield4 (approved):
LGTM, tested by making a test custom skill that calls `this.requestToolApproval()` and works as expected when approving and denying a tool call.

server/utils/agents/imported.js
```diff
@@ -178,6 +178,7 @@ class ImportedPlugin {
       name: this.name,
       config: this.config,
       setup(aibitat) {
+        const skillName = this.name; // hubId the skill is registered + whitelisted under.
```
timothycarambat: This is not needed if you instead just attach the function to aibitat like we do for logger and introspect

server/utils/agents/imported.js
```diff
@@ -196,6 +197,39 @@ class ImportedPlugin {
...
+                approved: true,
+                message: "Approval not required in this context.",
+              };
+            }
+            return aibitat.requestToolApproval({
+              skillName,
+              payload,
+              description,
+            });
+          },
```
timothycarambat: This is very messy for the skill setup. Instead we can make a factor creation method on this file and then attach it to the plugin.

```js
/**
   * Creates a requestToolApproval function bound to a specific aibitat
   * instance and skill name via closure — no `this` binding required.
   *
   * The returned function pauses the agent and asks the user to approve a
   * potentially destructive action before the skill performs it.
   *
   * `skillName` is captured from the skill's hubId so a skill cannot spoof
   * another tool's name to bypass its own approval.
   *
   * When no approval channel exists (e.g. scheduled jobs) it resolves
   * approved so the skill still runs, matching how built-in tools fall
   * through their approval guard.
   *
   * @param {object} aibitat - The aibitat instance.
   * @param {string} skillName - The skill's hubId.
   * @returns {function({payload?: object, description?: string}): Promise<{approved: boolean, message: string}>}
   */
  static createToolApprovalFn(aibitat, skillName) {
    return async function requestToolApproval({
      payload = {},
      description = null,
    } = {}) {
      if (typeof aibitat?.requestToolApproval !== "function") {
        return {
          approved: true,
          message: "Approval not required in this context.",
        };
      }
      return aibitat.requestToolApproval({
        skillName,
        payload,
        description,
      });
    };
  }
```

Which then lets this block become
```js
 requestToolApproval: ImportedPlugin.createToolApprovalFn(
            aibitat,
            this.config.name
          ),
```

timothycarambat (changes requested):
This can be cleaned up, which it does work.

One thing i notice in testing this actual change was that it reports the `hubID` in the UI using `this.name` in the plugin.

This should be the `this.config.name` because a UUID for a calling skill to the UI is not informative at all to the user who needs to determine if that is indeed the tool they wished to call.

PR #5799, opened 2026-06-10
feat: highlight variable references in agent flow blocks

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)
- [x] 💄 style (UI style changes)

### Relevant Issues

N/A (internal ticket)

### Description

It was not obvious how to reference flow variables inside agent flow blocks. The `${variableName}` syntax was only explained in the official docs, never surfaced in the UI, and there was no visual indication of which text in a field would be swapped for a variable at runtime.

This PR makes variables discoverable in the agent flow builder:

- **Live highlighting.** Variable references (`${variableName}`) are now highlighted inline as you type, the same way system prompt variables are highlighted on the Default System Prompt page. This is done with a new shared `VariableInput` component that layers a highlight backdrop behind a transparent-text field, so the field stays fully editable and the existing "insert variable" button keeps working.
- **A "Variables" hint.** The Flow Variables block now explains the `${variableName}` syntax and shows example chips for the variables you have defined, so you no longer have to read the docs to know how to reference them.

Highlighting and the hint cover the fields where variables are actually used: API Call (URL, request body, header values, form-data values), LLM Instruction (instruction), and Web Scraping (URL, query selector). Header and form-data names are left as plain inputs since variables there are rare.

### Visuals (if applicable)

image

image

image

image

### Additional Information

- The highlight uses `bg-sky-300/30` (a 30% tint). It is intentionally a standard Tailwind palette color rather than the theme's `cta-button`: the var-based theme colors do not support Tailwind opacity modifiers, and a translucent tint keeps the text readable against both the light and dark themes (a solid fill washed the text out in dark mode). `sky-300` matches the cta blue.

### Developer Validations

### DIFF
--- a/frontend/src/pages/Admin/AgentBuilder/nodes/StartNode/index.jsx
+++ b/frontend/src/pages/Admin/AgentBuilder/nodes/StartNode/index.jsx
@@ -1,5 +1,6 @@
-import React from "react";
+import React, { Fragment } from "react";
 import { Plus, X } from "@phosphor-icons/react";
+import { VARIABLE_HIGHLIGHT_CLASS } from "../../VariableInput";
 
 export default function StartNode({
   config,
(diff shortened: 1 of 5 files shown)


frontend/src/pages/Admin/AgentBuilder/nodes/StartNode/index.jsx
```diff
@@ -13,9 +14,40 @@ export default function StartNode({
...
+            {definedVariables.slice(0, 3).map((variable, index) => (
+              <Fragment key={variable.name}>
+                <span
+                  className={`${VARIABLE_HIGHLIGHT_CLASS} px-1 py-0.5 text-theme-text-primary`}
+                >
+                  {`\${${variable.name}}`}
+                </span>
+                {index < Math.min(definedVariables.length, 3) - 1 && ", "}
+              </Fragment>
+            ))}
```
shatfield4: Seems a bit hard to follow here. Instead of doing the `.slice()` on definedVariables then also recomputing the `index < Math.min(definedVariables.length, 3)` can we just do the slice outside of the jsx?

Something like this

```
  const definedVariables = config.variables.filter((v) => v.name);
  const exampleVariables = definedVariables.slice(0, 3);

  {exampleVariables.length > 0 && (
    <p className="text-xs text-theme-text-secondary">
      For example:{" "}
      {exampleVariables.map((variable, index) => (
        <Fragment key={variable.name}>
          <span
            className={`${VARIABLE_HIGHLIGHT_CLASS} px-1 py-0.5 text-theme-text-primary`}
          >
            {`\${${variable.name}}`}
          </span>
          {index < exampleVariables.length - 1 && ", "}
        </Fragment>
      ))}
    </p>
  )}
```

shatfield4 (changes requested):
Small nit to clean up rendered jsx.

frontend/src/pages/Admin/AgentBuilder/nodes/StartNode/index.jsx
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5799/commits/a2d8f4efdc8f91579adca9219a060e4b8c1c8b75

shatfield4 (approved):
LGTM

timothycarambat:
Coming back to issues with this PR:
- `mono` prop breaks the highlight and instead shows the boxes really far off of inputs
- This highlight anything could even potentially be a variable - even if it does not exist in the variable block. That is even more confusing UX wise - it should only be highlighting valid vars defined at top of scope.

PR #5800, opened 2026-06-11
feat: export workspace chat/chat thread to pdf

### LINKED ISSUE #1981: [FEAT]: Ability to export all or part of a specific thread content as PDF

It would be nice to export all or part of the conversation in a specific thread as a PDF document (or of other document format), which makes it easier to discuss the conversation with other people online. Currently the margin and line gaps are huge in the chat window, it's difficult to take a screenshot of even one full answer from the AI, not to mention the whole conversation.

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #1981
translations: https://github.com/Mintplex-Labs/anything-llm/pull/5845

### Description

- Adds an "Export chat" option to the chat settings menu (the sliders icon at the top of any workspace or thread chat)
- Selecting it generates a PDF of the conversation on the server and saves it through the standard file save dialog, so it works the same in the browser and the desktop app
- Reuses the same markdown-to-PDF pipeline as the existing create-pdf-file agent skill, including the AnythingLLM logo branding on each page
- Builds the PDF from the thread's stored chats, with a header showing the workspace, thread, and export date
- Renders the conversation as markdown including formatting, links, tables, and code blocks
- Includes images the user uploaded in the conversation
- Leaves out the assistant's thinking steps and agent-only items (tool approvals, clarification requests, status updates, charts/file cards) so only the actual chat is exported
- Saves the file with a clean name like "AnythingLLM Export - 6-11-2026 3-26-45 PM.pdf"
- Adds two backend endpoints (default workspace chat and thread) and a small frontend model method; no new dependencies

Also supports markdown, plaintext, JSON, and HTML format exports.

### Visuals (if applicable)

Sample export:
[AnythingLLM Export - 6-11-2026 3-41-35 PM.pdf](https://github.com/user-attachments/files/28857941/AnythingLLM.Export.-.6-11-2026.3-41-35.PM.pdf)

### Additional Information

- Tested in both thread and workspace chats
- Had to use a developer api key to create the default thread and send some chats since we now hide this in the UI unless default thread already existed
- Works as expected in both thread and workspace chats

### Developer Validations


timothycarambat:
@shatfield4 Do not do this using the browser printer because that will not work in Desktop since they do not function the same.

Since all chats should be markdown (or close to it) we can use the same exact stuff we do for PDF generation as we do for `create-pdf-file` using `marked` and then saving that as a blob we can return to frontend which will then be open the standard file picker save location selection and will work cross platform for browser/native.

If we want to capture _everything_ we could even send the frontend as the payload and parse it that way but TBH this should work like LM Studio thread exports since that is what people expect - which isnt very fancy.

frontend/src/components/WorkspaceChat/ChatContainer/ChatSettingsMenu/Export/index.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+  async function handleClick() {
+    if (exporting || !workspace?.slug) return;
+    setExporting(true);
+    const blob = await Workspace.exportChatsToPDF(workspace.slug, threadSlug);
+    if (blob) {
+      const stamp = new Date()
+        .toLocaleString()
+        .replace(", ", " ")
+        .replace(/[/:]/g, "-");
```
timothycarambat: Can probably just use moment here? 
`moment().format('MM-DD-YYYY HH-mm-ss');`

server/endpoints/workspaces.js
```diff
@@ -423,6 +424,27 @@ function workspaceEndpoints(app) {
...
+  app.get(
+    "/workspace/:slug/export/pdf",
+    [validatedRequest, flexUserRoleValid([ROLES.all]), validWorkspaceSlug],
+    async (request, response) => {
+      try {
+        const user = await userFromSession(request, response);
+        const workspace = response.locals.workspace;
+        const chats = multiUserMode(response)
+          ? await WorkspaceChats.forWorkspaceByUser(workspace.id, user.id)
+          : await WorkspaceChats.forWorkspace(workspace.id);
```
timothycarambat: Feel like we should do something here for no chats - is this just an empty PDF? Since we block this in the UI lets just throw so the catch gets it since its an invalid UI op anyway?

timothycarambat (changes requested):
very small nits/questions.

Other things:
- What do `outputs` like when it generates a PDF look like? or like a powerpoint/text file etc.
- Any idea what surveys look like? Should be okay - just asking.
...that should be all the weird stuff since routing wont show in this.

i saw the image stuff - looks excellent. Also did a bunch of other markdown elements and they all render great - stripping thinking is smart IMO but i fully expect someone ask to print it 😓  - lets wait until some asks.

PR #5813, opened 2026-06-12
Migrate Cohere to OpenAI sdk

shatfield4:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #

### Description

- Migrate the Cohere provider to the OpenAI SDK via Cohere's OpenAI-compatible API across the whole app (chat, agents, and embeddings)
- Enable native tool calling for Cohere agent models, with automatic fallback to UnTooled for models that don't support tools
- Move Cohere embeddings onto the OpenAI SDK as well
- Keep Cohere model listing (with chat/embed filtering) working without the Cohere SDK
- Remove the Cohere-specific dependencies (`cohere-ai` and `@langchain/cohere`)
- Preserve accurate token/usage metric reporting for both streaming and non-streaming chat

### Additional Information

- Tested `command-a-plus-05-2026` as the workspace LLM agent tool calls worked perfectly using native tool calling
- Confirmed models without tool support (`c4ai-aya` models) automatically fall back to the UnTooled approach with no errors
- Verified Cohere embeddings work and citations appear correctly after embedding

### Developer Validations


timothycarambat (changes requested):
You can detect if a model supports tool user via the main REST API:
```bash
curl --location 'https://api.cohere.com/v1/models/command-a-03-2025' \
--header 'Authorization: Bearer xxxxxxx'
```

Note: this also is reported for all models via GET `https://api.cohere.com/v1/models`

```json
{
    "name": "command-a-03-2025",
    "endpoints": [
        "chat"
    ],
    "finetuned": false,
    "context_length": 288000,
    "tokenizer_url": "https://storage.googleapis.com/cohere-public/tokenizers/command-a-03-2025.json",
    "features": [
        "json_mode",
        "json_schema",
        "strict_tools",
        "safety_modes",
        "tools",
        "tool_choice"
    ],
    "default_endpoints": [
        "chat",
        "generate"
    ]
}
```

By checking for `tools` or `tool_choice` in `features` we can make this detection smart instead of doing a failover double-send per request for every agent chat - which is a waste of tokens and HTTP overhead for non-tc models.

Reimplement this to be similar to the other OpenAI-like clients like LMStudio and add the supported AIProvider class method `getModelCapabilities` and then in the agent provider override the ai-provider default `supportsNativeToolCalling`

```js
/**
   * Whether the loaded model supports native OpenAI-compatible tool calling.
   * Checks the LMStudio /api/v1/models endpoint for the model's capabilities.
   * @returns {Promise<boolean>}
   */
  async supportsNativeToolCalling() {
    if (this.optsOutOfNativeToolCallingViaEnv(this.providerTag)) return false;
    if (this._supportsToolCalling !== null) return this._supportsToolCalling;
     // code for cohere provider to detect tool-calling for model.
    this._supportsToolCalling = capabilities.tools === true;
    return this._supportsToolCalling;
  }
```

This way we are actually checking model capability and not just letting it failover via OpenAI Errors.

PR #5815, opened 2026-06-12
Document TrustedRouter OpenAI-compatible setup

jperla:
This pull request adds a short README note for using TrustedRouter through AnythingLLM’s existing OpenAI-compatible provider configuration. TrustedRouter is an open-source, verifiable attested router for privacy-sensitive workflows with zero prompt and output logging by default.

Users can connect with the base URL `https://api.trustedrouter.com/v1` and a TrustedRouter API key.

Verification: `git diff --check`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,7 @@ AnythingLLM supports multiple users as well where you can control the access and
 - [Fireworks AI (chat models)](https://fireworks.ai/)
 - [Perplexity (chat models)](https://www.perplexity.ai/)
 - [OpenRouter (chat models)](https://openrouter.ai/)
+- [TrustedRouter](https://trustedrouter.com) through the OpenAI-compatible base URL `https://api.trustedrouter.com/v1`
 - [DeepSeek (chat models)](https://deepseek.com/)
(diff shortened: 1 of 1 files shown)


timothycarambat:
That list is only for explicitly supported providers, if we include any service under generic we have to include every one of them to be fair. Otherwise, it looks like free advertising to people who are not in the list.

jperla:
Thanks, that makes sense. I do not want this to look like free advertising in a generic list. Is there any format you would accept, for example a real explicit provider integration, a neutral OpenAI-compatible router example, or a docs issue before a PR? TrustedRouter is built as an end-to-end encrypted, open-source router, so I am happy to adapt this to whichever format is useful to AnythingLLM users.

jperla:
One clarification on why I am asking about an acceptable format: TrustedRouter is built to be the #1 end-to-end encrypted open-source router, not just another provider listing. The value for users is a verifiable private OpenAI-compatible route for sensitive prompts. If listing it generically feels unfair, would a neutral custom OpenAI-compatible router example or a real explicit integration be acceptable?

PR #5856, opened 2026-06-17
Toggle agent tools during an active agent session

### LINKED ISSUE #5855: [FEAT]: Ability to toggle agent skills in live agent session

While an agent session is active we should be able to manage the tools available to the agent.
This should work by modifying the plugins available to the aibitat agent. There should be a remove function exactly for this.

- This should be handled via a special WSS message or something that we can send over the WS that the aibitat can handle specifically while connection is open
- This should be done with the least amount of code change possible on the server/aibitat side

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5855

### Description

- Adds the ability to enable or disable an agent's tools while an agent session is already active
- Toggles are sent over the existing agent websocket and applied live, taking effect on the agent's next turn with no session restart
- The agent rebuilds the tools it can call from a map on every tool-call turn, so toggles apply on the next turn
- Disabling moves a tool out of the active map into a removed map so it is skipped that turn; enabling moves it back, preserving the definition so re-enabling never reloads the plugin
- Works across every tool type: default skills, configurable skills, app integrations, custom skills, agent flows, MCP tools, and sub-skills
- Multi-tool skills are resolved server-side so the whole skill toggles correctly (SQL tool)
- Removes the old behavior that disabled the tools menu during an agent session, and the unused tooltip
- Tool toggles continue to persist to the global agent skill settings

### Additional Information

- [ ] Prune translations for the `agent-skill-disabled-tooltip` translation key
- [ ] Possibly add a needs setup hint to certain skills that are not setup yet but enabled in the menu? (Currently fails gracefully and skips the failed tool call if not setup in agent skills settings page)

### Developer Validations


timothycarambat:
@shatfield4 Quick feedback just from a glance - the relayToSocket interception should be handled likely within the websocket tool or something provided it can access aibitat.

Second, you can inspect `agentHandler.aibitat.functions` for the list of active plugins on an instance. THen from here it is a matter of finding and popping off the right function from the Map.

Then vice versa for adding plugins back - which must be done via the `aibitat.use(plugin)` caller. We would need another helper for `remove(skillname)` which could do the deletion off of `aibitat.functions` and the skill would not longer be available.

The frontend<>backend skill name resolutions + all the tricks will be annoying. Things like `sql-agent` are not granular so we may need some special cases for that - or maybe even a way to determine if an enabled/disabled skill is a parent or child function and turn on/off accordingly. 

I believe you can get the unresolved function names from `agentHandler.aibitat.agents.get("@agent")` and it should list them IIRC

server/utils/agents/aibitat/plugins/websocket.js
```diff
@@ -87,6 +87,20 @@ const websocket = {
           },
         };
 
+        // Toggle a tool/skill on or off for the running agent mid-session. The
+        // change applies on the agent's next turn. Returns true once handled so
+        // the socket message router stops further dispatch.
+        socket.handleToolToggle = (message) => {
+          const data = safeJsonParse(message, {});
+          if (data?.type !== "agentToolToggle") return false;
```
angelplusultra, line 116: is there any authorization here for this? Only admins can see the Agent Skills tab in the prompt bar so it'd make sense to gate it on the backend like we do with the regular endpoint for toggling skills.

angelplusultra (changes requested):
Nice job.

I left one comment about auth check for toggling tools via the socket since that's an admin action.

server/utils/agents/aibitat/plugins/websocket.js
shatfield4: 44e9de0a3

angelplusultra (approved):
LGTM

The-SMG:
- name: Delete PR tag from GHCR
        env:
          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: |
          # 1. Define variables (adjust package name to match yours)
          PACKAGE_NAME="your-image-name"
          TAG_NAME="pr-${{ github.event.pull_request.number }}"
          OWNER="${{ github.repository_owner }}"

          echo "Searching for version ID for tag: $TAG_NAME..."

          # 2. Fetch the Version ID using GitHub CLI
          VERSION_ID=$(gh api orgs/$OWNER/packages/container/$PACKAGE_NAME/versions \
            --paginate \
            --jq ".[] | select(.metadata.container.tags[] | contains(\"$TAG_NAME\")) | .id" | head -n 1)

          if [ -z "$VERSION_ID" ]; then
            echo "No package version found with tag $TAG_NAME. Nothing to delete."
          else
            echo "Found Version ID: $VERSION_ID. Proceeding with deletion..."
            # 3. Execute the delete operation
            gh api --method DELETE orgs/$OWNER/packages/container/$PACKAGE_NAME/versions/$VERSION_ID
            echo "Successfully deleted $TAG_NAME from GHCR!"
          fi

PR #5863, opened 2026-06-17
add missing headers to ollama embeder engine isAlive method

### LINKED ISSUE #5862: [BUG]: Ollama embedding engine #isAlive() fails for auth-protected instances

Docker (local)

### What happened?
 
The `#isAlive()` health-check method in the Ollama embedding engine does not include authentication headers when making its request. This causes the method to return `false` for any Ollama instance that requires a `Bearer` token, even though the instance is running and healthy.
 
The rest of the embedding engine correctly uses `this.authToken` when making inference requests, but the liveness check was missed.
 
## Steps to reproduce
 
1. Run an Ollama instance behind an auth proxy (or with any setup that requires a `Bearer` token).
2. Configure AnythingLLM to use that instance as the embedding engine, providing the auth token.
3. Attempt to embed a document.
**Expected:** The `#isAlive()` check passes and embedding proceeds normally.
**Actual:** `#isAlive()` returns `false` because the unauthenticated request gets a `401`/`403`, blocking embedding operations.
 
## Proposed fix
 
Pass the existing `this.authToken` as a `Bearer` header in the fetch call:
 
```javascript
// Before
async #isAlive() {
  return await fetch(this.basePath)
    .then((res) => res.ok)
    .catch((e) => {
      this.log(e.message);
      return false;
    });
}
 
// After
async #isAlive() {
  const headers = this.authToken
    ? { Authorization: `Bearer ${this.authToken}` }
    : {};
 
  return await fetch(this.basePath, { headers })
    .then((res) => res.ok)
    .catch((e) => {
      this.log(e.message);
      return false;
    });
}
```
 
This is a minimal, backwards-compatible change — instances without auth are unaffected since the headers object will be empty.
 
I'm happy to open a PR for this.

_No response_

Ollama

Ollama

TomaszZakrzewski:
### Pull Request Type
- [ ] ✨ feat (New feature)
- [X] 🐛 fix (Bug fix)
### Relevant Issues
resolves #5862
 
### Description
The `#isAlive()` method in the Ollama embedding engine makes an unauthenticated fetch to `this.basePath`. For Ollama instances that require a Bearer token, this returns a 401/403, causing the health check to fail even though the instance is running.
 
This PR adds the existing `this.authToken` as a Bearer header to the liveness check, matching how the rest of the engine already handles auth. Instances without auth are unaffected since the headers object defaults to empty.
 
### Visuals (if applicable)
N/A — no UI changes.
 
### Additional Information
Minimal, single-line change. No new dependencies or config required.
 
### Developer Validations

### DIFF
--- a/server/utils/EmbeddingEngines/ollama/index.js
+++ b/server/utils/EmbeddingEngines/ollama/index.js
@@ -43,7 +43,11 @@ class OllamaEmbedder {
    * @returns {Promise<boolean>} - A promise that resolves to true if the service is alive, false otherwise.
    */
   async #isAlive() {
-    return await fetch(this.basePath)
+    const headers = this.authToken
+      ? { Authorization: `Bearer ${this.authToken}` }
+      : {};
+
+    return await fetch(this.basePath, { headers })
       .then((res) => res.ok)
       .catch((e) => {
         this.log(e.message);


timothycarambat:
Great catch!

PR #5867, opened 2026-06-17
Report attached & pinned documents as citations in agent chat

### LINKED ISSUE #5840: [BUG]: No citations when dnd file into agent chat

Docker (local)

### What happened?

When I drag and drop a document into the chat window in an agentic chat, I get document based answers, but I do not get reported citations.

_No response_

_No response_

_No response_

sanidhyasin:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5840

### Description

When a document is drag-and-dropped into an agent chat (a "parsed" file) or a document is pinned, its content is injected into the user message via `fetchParsedFileContext()` so the agent can answer from it. However, those documents were never registered as citations, so agent replies returned document-based answers with **no citations** — unlike non-agent chat (`server/utils/chats/stream.js`), which pushes both parsed files and pinned docs into `sources`.

This change makes the agent flow mirror non-agent chat:

- `fetchParsedFileContext()` (in both `AgentHandler` and `EphemeralAgentHandler`) now returns `{ context, sources }`, where `sources` are citation objects built for each injected parsed/pinned document.
- `AIbitat.reply()` injects the context as before and registers the document sources via the existing `addCitation()` mechanism, so they flush to the UI and persist with the chat exactly like other agent citations (web browsing, etc.).
- Sources are deduped against already-pending citations so a multi-turn reply (e.g. tool calls cause `reply()` to run more than once within a single response) does not cite the same document repeatedly.

Citation objects use the same shape the frontend already expects (`{ id, title, text, chunkSource, score }`). For uploaded files `chunkSource` is empty, which the frontend resolves to the standard file icon.

A small pure helper `documentToCitationSource()` was extracted (`server/utils/agents/aibitat/utils/parsedFileSources.js`) and unit tested.

### Visuals (if applicable)

N/A — citations now appear under agent replies that use attached/pinned documents.

### Additional Information

New unit tests in `server/__tests__/utils/agents/aibitat/utils/parsedFileSources.test.js` cover the citation shape, title/id fallbacks, text truncation, and `chunkSource` passthrough.

### Developer Validations

> Note: code was hand-formatted to match the repo's Prettier config; the new helper is covered by passing unit tests (`parsedFileSources.test.js`).

### DIFF
--- a/server/__tests__/utils/agents/aibitat/utils/parsedFileSources.test.js
+++ b/server/__tests__/utils/agents/aibitat/utils/parsedFileSources.test.js
@@ -1,0 +1,79 @@
+/* eslint-env jest */
+const {
+  documentToCitationSource,
+} = require("../../../../../utils/agents/aibitat/utils/parsedFileSources");
+
+describe("documentToCitationSource", () => {
+  test("builds a citation source for a parsed/uploaded document", () => {
+    const doc = {
+      id: "doc-123",
+      title: "report.pdf",
+      pageContent: "The quarterly revenue grew by 12%.",
+      chunkSource: "",
+    };
+
+    expect(documentToCitationSource(doc, "Uploaded Document")).toEqual({
+      id: "doc-123",
+      title: "report.pdf",
+      text: "The quarterly revenue grew by 12%....continued on in source document...",
+      chunkSource: "",
+      score: null,
+    });
+  });
+
+  test("falls back to the provided title when the document has none", () => {
+    const source = documentToCitationSource(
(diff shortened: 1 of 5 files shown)


timothycarambat:
When you submit this PR that issue already had a PR out #5849 - this PR also does not effect the frontend "Ghost document" file that is part of the scope of this PR!

I appreciate you wanting to make a contribution but dont do accidental double work! That issue on creation was assigned so it meant someone was working it

PR #5886, opened 2026-06-19
feat: add a per-message "Thinking" toggle to suppress reasoning

dennisliu1112:
### Pull Request Type

- [x] ✨ feat (New feature)

### Relevant Issues

N/A

### Description

Reasoning-capable models (Qwen3, DeepSeek-R1, QwQ, etc.) emit a `<think>…</think>` block before their answer. AnythingLLM already **displays** this nicely (collapsible reasoning), but there is currently no way to **skip** it. For quick or generative tasks (short writing, formatting, casual chat) the model can spend most — or all — of its token budget reasoning and never reach the answer.

This PR adds a per-message **"Thinking" toggle** to the chat input bar. When turned **off**, the server prefills an empty assistant `<think>\n\n</think>` block so the model treats reasoning as already complete and answers directly. Default is **on** (current behavior — zero change for existing users).

It complements the existing reasoning *display* (`stripThinkingFromText`) by adding reasoning *suppression*.

**Design — intentionally minimal & decoupled:**
- Core change is a single injection in `server/utils/chats/stream.js` (right after the messages array is assembled). No LLM provider classes are modified.
- Everything else just threads an optional `suppressThinking` flag (forward-compatible additions): chat endpoints → `streamChatWithWorkspace` → frontend request body / models.
- Toggle state is persisted per-workspace in `localStorage` and read at send-time, so there is **no schema/DB change**.

### Visuals (if applicable)

A 🧠 **Think** toggle appears in the input bar next to **Tools**. Highlighted = reasoning on (default); click to dim = answer directly.

Thinking toggle in the chat input bar

### Additional Information

- The empty-think prefill `"<think>\n\n</think>\n\n"` is the de-facto token shared by Qwen3 / DeepSeek-R1 / QwQ. It is extracted to a named constant (`REASONING_SUPPRESSION_PREFILL`) — happy to promote it to a configurable per-workspace setting if maintainers prefer.
- Applies to **chat / query** mode. In **agent** mode the flow exits early (native tool calling), so the toggle has no effect there — by design.
- Harmless for non-reasoning models (an empty assistant turn is a no-op).

**Testing:** verified end-to-end against a local LM Studio serving a Qwen3.5-35B-A3B reasoning model:
- Toggle **off** → response contains no `<think>`, answers directly (~3.5s).
- Toggle **on** → reasons first (`<think>` present, ~37s).

### Developer Validations

- [x] `eslint` passes on the changed files and `vite build` succeeds (lint clean)

### DIFF
--- a/frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/index.jsx
@@ -1,6 +1,6 @@
 import { useState, useRef, useEffect } from "react";
 import debounce from "lodash.debounce";
-import { ArrowUp, At } from "@phosphor-icons/react";
+import { ArrowUp, At, Brain } from "@phosphor-icons/react";
 import StopGenerationButton from "./StopGenerationButton";
 import SpeechToText from "./SpeechToText";
 import { Tooltip } from "react-tooltip";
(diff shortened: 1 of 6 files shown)


dennisliu1112:
anythingllm_think_full
anythingllm_think_button

PR #5904, opened 2026-06-24
feat: add image generation via /img command

### LINKED ISSUE #943: [FEAT]: Image generation

In addition to text generation, is it possible to allow anythingLLM to generate images?
Via the image generation API of OpenAI, or Gemini, etc.? 
PS: i mean text to image and not image to text

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #943

### Description

Adds first-class image generation to AnythingLLM. Users can generate an image directly in any workspace or thread chat by typing the `/img <prompt>` slash command.

- New Image Generation subsystem with day-one support for OpenAI, Ollama (macOS), Lemonade, and OpenRouter
- Configured from a new **AI Providers › Image Generation** settings page that works the same way as the Embedder and Vector Database pages
- Reuses each provider's existing API keys/credentials — only the provider selection and model are image-specific, so there are no new keys to manage
- Model selection is a dropdown populated from each provider's available models, with automatic fallback to manual model-name entry when a provider can't list them
- Generated images are stored on disk, rendered inline as a compact card with hover download/copy actions, and open full-size in the existing image lightbox
- Dedicated pending state while an image is generating, since generation can take a while (no misleading time estimate)
- Image responses are persisted to chat history and survive reload (history/RAG aware)
- Adds an OpenAI-compatible image generation endpoint to the developer API
- Background job cleans up generated images that are no longer referenced by any active chat to keep storage in check
- Per-model output sizes are handled automatically — targets 512x512, picks the best supported size up front for known models, and retries on the model's supported sizes when needed
- Added image editing in UI and API via image reference attachments or pure prompt edits
- Abort signal handling where available (Ollama + cloud only - lemonade does not support abort)

### Visuals (if applicable)

generating

generated-image

### Additional Information

- Translations: https://github.com/Mintplex-Labs/anything-llm/pull/6093

- [x] Simplify background job that cleans up files and test it fully
- [x] Add support for images as context when generating an image and image is attached to chat
- [x] Add docs PR - https://github.com/Mintplex-Labs/anythingllm-docs/pull/270

### Tested Providers

- [x] OpenAI
- [x] Ollama - Tested with [flux2-klein](https://ollama.com/x/flux2-klein)
- [x] OpenRouter
- [x] Lemonade

### Developer Validations


timothycarambat (changes requested):
Looks awesome from a first pass from what I can see. Very well structured and looks very plug and play.

only thing that I think needs a second pass is the ENVs for the providers. We made this mistake a long long time ago with Azure and OpenAI sharing an ENV for model preference and it annoyed people who wanted to use different endpoints or keys for LLM vs Embedding.

So while tedious, this should be shaped to where the image generation provider keys and endpoint **should not** be overlapping with the LLM. I want us to avoid that collision especially with image generation because this is the exact use case I have heard before:

- I have Lemonade on my actual PC
- I have lemonade on my AMD Ryzen Halo/DGX Spark *just for image gen* because these models are so demanding.

Both local, two different endpoints.

---

So that is the only suggested change I have, lets not re-use keys here so at least they are all meaningfully separated becauses thats going to be the first issue we get post-merge. 

We already need to do a big re-arch for tracking "known connections" across the system  and I have a plan for that. The less we duplicate an ENV use for a provider the better.

server/jobs/cleanup-generated-images.js
```diff
@@ -0,0 +1,92 @@
...
+
+    const referenced = await referencedImageFilenames();
+    const orphans = candidates.filter((name) => !referenced.has(name));
+    if (orphans.length === 0) return;
+
+    log(`Found ${orphans.length} orphaned images to delete.`);
+    let deleted = 0;
+    for (const name of orphans) {
+      try {
+        fs.unlinkSync(path.join(generatedImagesPath, name));
```
angelplusultra, line 39: might wanna make this an async file system operation.

server/jobs/cleanup-generated-images.js
```diff
@@ -0,0 +1,92 @@
...
+// Generated images are always stored as `img-<uuid>.png`.
+const IMAGE_FILENAME_PATTERN = /^img-[a-f0-9-]{36}\.png$/i;
+
+// Ignore images younger than this so a freshly generated file isn't deleted in
+// the window between writing it to disk and persisting its chat reference.
+const MIN_AGE_MS = 60 * 60 * 1000;
+
+(async () => {
+  try {
+    if (!fs.existsSync(generatedImagesPath)) return;
```
angelplusultra, line 18: async here if possible

server/jobs/cleanup-generated-images.js
```diff
@@ -0,0 +1,92 @@
...
+
+    const now = Date.now();
+    const candidates = fs
+      .readdirSync(generatedImagesPath)
```
angelplusultra, line 22: async here if possible

server/endpoints/agentFileServer.js
```diff
@@ -86,6 +86,51 @@ function agentFileServerEndpoints(app) {
...
+          user,
+          isMultiUser: multiUserMode(response),
+        });
+        if (!fileSource)
+          return response
+            .status(404)
+            .json({ error: "Image not found or access denied" });
+
+        const imagePath = path.resolve(generatedImagesPath, filename);
+        if (!fs.existsSync(imagePath))
```
angelplusultra: async if possible

server/utils/chats/commands/img.js
```diff
@@ -0,0 +1,104 @@
...
+    const canConfigure = !user || user.role === "admin";
+    return {
+      uuid: msgUUID,
+      type: "textResponse",
+      textResponse: canConfigure
+        ? "Image generation isn't set up yet. Choose a provider in Settings → Image Generation."
+        : "Image generation isn't set up yet. Contact an admin to configure it.",
+      sources: [],
+      close: true,
+      error: false,
```
angelplusultra: This error state comes back as regular text in the UI and doesn't really look like a system level error. Consider putting the text in the `error` field that way an actual chat error renders instead.

server/utils/chats/commands/img.js
```diff
@@ -0,0 +1,104 @@
...
+      outputs,
+      chatId: chat?.id || null,
+      close: true,
+      error: false,
+    };
+  } catch (e) {
+    return {
+      uuid: msgUUID,
+      type: "textResponse",
+      textResponse: `Image generation failed: ${e.message}`,
```
angelplusultra: Same here, feel like this should be an actual chat error not regular text

server/utils/helpers/customModels.js
```diff
@@ -1332,6 +1339,59 @@ async function getBedrockModels(_apiKey = null, options = {}) {
...
+ */
+async function getOpenAiImageModels(apiKey = null) {
+  const { OpenAI: OpenAIApi } = require("openai");
+  const openai = new OpenAIApi({ apiKey: apiKey || process.env.OPEN_AI_KEY });
+  const models = await openai.models
+    .list()
+    .then((results) => results.data)
+    .then((all) =>
+      all
+        .filter((model) => /dall-e|gpt-image/i.test(model.id))
```
angelplusultra: Does this need to be constantly maintained? Is there away to just filter on capability instead of model name? If they have no support for this, you could use models.dev

heres a basic example:

```ts
const FALLBACK_OPENAI_IMAGE_MODELS = [
  "chatgpt-image-latest",
  "gpt-image-1.5",
  "gpt-image-1",
  "gpt-image-1-mini",
  "gpt-image-2",
];

async function getOpenAIImageModels() {
  try {
    const res = await fetch("https://models.dev/api.json");

    if (!res.ok) {
      throw new Error(`Failed to fetch models.dev data: ${res.status}`);
    }

    const data = await res.json();
    const openAiModels = data?.openai?.models ?? {};

    return Object.entries(openAiModels)
      .filter(([, model]) => model?.modalities?.output?.includes("image"))
      .map(([modelName]) => modelName);
  } catch {
    return FALLBACK_OPENAI_IMAGE_MODELS;
  }
}
```

server/utils/ImageGenerators/base.js
```diff
@@ -0,0 +1,134 @@
...
+// Not every model supports the 512x512 target. For known model families we pick
+// the best size up front; otherwise we send the target and, if the API rejects
+// it, parse the supported sizes from the error and retry once. Keys are matched
+// as substrings of the model id, so "gpt-image" covers gpt-image-1, -mini, -2.
+const TARGET_SIZE = 512;
+const MODEL_SUPPORTED_SIZES = {
+  "dall-e-2": ["256x256", "512x512", "1024x1024"],
+  "dall-e-3": ["1024x1024", "1024x1792", "1792x1024"],
+  "gpt-image": ["1024x1024", "1024x1536", "1536x1024"],
+};
```
angelplusultra: This whole process of finding a supported resolution feels kinda fragile to me, and it's something we'll have to keep maintaining. We've got a hardcoded list of model sizes that'll need updating over time, plus we're parsing sizes out of provider error messages, which feels like it'll break the second a provider changes their error text.

Couldn't we just let the user set the resolution in the provider config (e.g. IMAGE_GENERATION_SIZE_PREF)? Seems like it'd avoid the hardcoded-data trap, and it fits with how we already config the provider and model there.

Also noticed the developer API already lets you pass an arbitrary size, but /img is locked to 512×512, so that's kinda inconsistent between the two as-is.

We might also receive complaints about it as is where users cant specify the desired size via chat, this would mitigate that.

Maybe we could keep this as a fallback if a size isn't configured? Just spitballin ideas.

angelplusultra (changes requested):
Nice job!

I don’t have many concerns with the implementation. The only small thing I noticed is that some of the file system operations are synchronous, and it might be worth converting those to async. Not a huge issue, but worth pointing out.

Theres also a bit about image size being customizeable via the developer API but static via chat. Worth opening a discussion about.


### Questions

1. Is LM Studio not compatible to be an image provider? I see Ollama and Lemonade were added just wondering why LM Studio was left out.


### LLM Vison Requirement for Follow Ups
One thing that may be worth thinking through is follow-up chats after image generation. Right now, those follow-ups seem to rely on the workspace model being vision-capable, but I don’t think the implementation gates the inclusion of generated images based on whether the selected model supports vision.

I get that checking this capability could range from easy to basically impossible depending on the provider (though [models.dev](https://models.dev) is a great resource for this) so I understand why adding that kind of gate could be painful. But it may still be something to be aware of. For example, if my workspace model is non-vision and I ask what it sees after an image is generated, I get a chat error.

If the expectation is that users are responsible for configuring a vision-capable model, that’s fine. But it might be worth considering some kind of warning that follow-up prompts involving generated images require a vision model. I noticed drag-and-drop image uploads also don’t seem to gate based on model capability, so maybe this is already an accepted limitation. Just wanted to flag it.

Image

server/jobs/cleanup-generated-images.js
shatfield4: Keeping synchronous since this job is spawned in the bree worker and will not block the event loop. Doing this to keep it consistent with all other background jobs since they all use synchronous operations.

server/jobs/cleanup-generated-images.js
shatfield4: Keeping synchronous since this job is spawned in the bree worker and will not block the event loop. Doing this to keep it consistent with all other background jobs since they all use synchronous operations.

server/endpoints/agentFileServer.js
shatfield4: 75c919b62

server/jobs/cleanup-generated-images.js
shatfield4: Keeping synchronous since this job is spawned in the bree worker and will not block the event loop. Doing this to keep it consistent with all other background jobs since they all use synchronous operations.

server/utils/chats/commands/img.js
shatfield4: 0a16bfdec

server/utils/chats/commands/img.js
shatfield4: 0a16bfdec

server/utils/helpers/customModels.js
shatfield4: I looked into using models.dev for this but it seems overkill for what we are trying to do here. We will only have to maintain this by adding to the new `OPENAI_MODEL_FAMILIES` const if new image models with different names drop here. I also tested models.dev against Openrouter but it already allows us to filter by models that support image generation. The only provider that would benefit from models.dev is OpenAI and filtering by the model family name is a much simpler solution.

Ollama and Lemonade may benefit from this but from my testing, local models typically have very different names so filtering from models.dev may actually cause create more problems by accidentally filtering out models that models.dev does not have. I decided we should leave it up to the user to choose which models support image generation and give them full control over this.

Note for the future: models.dev is very powerful and we can replace the existing cached models that we pull from LiteLLM's repo in another PR.

server/utils/ImageGenerators/base.js
shatfield4: I removed this fallback logic now to help keep it much cleaner. We now default to 1024x1024 for all image models since this causes the least amount of problems with sizing. There are some models that do not support 512x512 which was the initial spec provided for this feature. I added an `IMAGE_GEN_SIZE_PREF` env variable which allows the user to configure this. This is just phase 1 of image generation so we may come back later to add options for image generation sizing within the frontend. For now this is more than viable for users to be able to begin generating images.

shatfield4:
> Nice job!
> 
> I don’t have many concerns with the implementation. The only small thing I noticed is that some of the file system operations are synchronous, and it might be worth converting those to async. Not a huge issue, but worth pointing out.
> 
> Theres also a bit about image size being customizeable via the developer API but static via chat. Worth opening a discussion about.
> 
> ### Questions
> 1. Is LM Studio not compatible to be an image provider? I see Ollama and Lemonade were added just wondering why LM Studio was left out.
> 
> ### LLM Vison Requirement for Follow Ups
> One thing that may be worth thinking through is follow-up chats after image generation. Right now, those follow-ups seem to rely on the workspace model being vision-capable, but I don’t think the implementation gates the inclusion of generated images based on whether the selected model supports vision.
> 
> I get that checking this capability could range from easy to basically impossible depending on the provider (though [models.dev](https://models.dev) is a great resource for this) so I understand why adding that kind of gate could be painful. But it may still be something to be aware of. For example, if my workspace model is non-vision and I ask what it sees after an image is generated, I get a chat error.
> 
> If the expectation is that users are responsible for configuring a vision-capable model, that’s fine. But it might be worth considering some kind of warning that follow-up prompts involving generated images require a vision model. I noticed drag-and-drop image uploads also don’t seem to gate based on model capability, so maybe this is already an accepted limitation. Just wanted to flag it.
> 
> Image

LMStudio does not support image generation models as of writing this so that is why it was left out. The warning for a vision based model is out of scope for this PR since we already have decided for other places of the frontend that the user is responsible for setting up a model that supports images. The current error message provided to the user is more than enough for them to know that they need to setup another model that supports images if they run into this.

server/.env.example
```diff
@@ -254,11 +254,19 @@ SIG_SALT='salt' # Please generate random string at least 32 chars long.
 ######## Image Generation Selection #######
 ###########################################
 # Used by the /img chat command. Images are generated at 512x512.
```
angelplusultra: Stale comment, images are now defaulted to 1024x1024

frontend/src/locales/en/common.js
```diff
@@ -1119,6 +1120,12 @@ const TRANSLATIONS = {
       title: "Embedding Provider",
     },
   },
+  imageGeneration: {
+    title: "Image Generation Preference",
+    description:
+      "Configure the provider used to generate images from the /img chat command. Images are generated at 512x512.",
```
angelplusultra: Stale text. Image default size is 1024x1024

angelplusultra (changes requested):
Some stale comments and UI text, after those get fixed LGTM

server/.env.example
shatfield4: 92b6987d2

frontend/src/locales/en/common.js
shatfield4: 92b6987d2

angelplusultra (approved):
LGTM. I also verified Lemonade works as well. I was able to run StableDiffusion.cpp on my m5 macbook pro and test it using `SD-Turbo`.

Robinnnnn:
Nice feature @shatfield4! If you're routing through OpenRouter, consider using the dedicated `POST /api/v1/images` endpoint instead of chat completions — cleaner response shape (b64_json + media_type), correct image billing, and support for image-specific params. Docs: https://openrouter.ai/docs/api-reference/images. Glad to help if useful.

timothycarambat:
@shatfield4 - the default command is missing in the slash command input
Screenshot 2026-08-04 at 2 39 19 PM

shatfield4:
> @shatfield4 - the default command is missing in the slash command input Screenshot 2026-08-04 at 2 39 19 PM

An image generation provider needs to be set in order for it to show up. I just pulled in and tested this and works as expected.

PR #5906, opened 2026-06-24
Unlock Chat History Cap

### LINKED ISSUE #5899: [FEAT]: Increase chat history limit beyond 45 messages for large-context models (with dynamic auto-sizing option)
## What would you like to see?

The chat history slider in workspace settings is hardcoded to a maximum of 45 messages (`max={45}` in `ChatHistorySettings/index.jsx`). This was a reasonable default when AnythingLLM was primarily used with small-context models (8K–32K tokens), but it has become a significant functional limitation as 1M-token context window models have become mainstream.

## The problem in concrete terms

At 45 messages and ~1,000 tokens per message, you're sending roughly 45,000 tokens of history. Against a 1M token context window, that's **less than 5% utilization**. The remaining 95% of the model's capacity sits completely idle every single request.

## Models where this is actively limiting today

| Model | Context Window | Max useful history messages |
|---|---|---|
| Claude Opus 4.8 | 1,000,000 tokens | ~500–1,000 |
| Claude Sonnet 4.6 | 1,000,000 tokens | ~500–1,000 |
| GPT-5.5 / latest OpenAI models | 1,000,000 tokens | ~500–1,000 |
| Minimax M3 | 1,000,000 tokens | ~500–1,000 |
| GLM-5.2 (744B MoE) | 1,000,000 tokens | ~500–1,000 |
| Nemotron-3-Super | 1,000,000 tokens | ~500–1,000 |

This isn't a niche concern — 1M token context is now the standard for frontier models both via API and for local inference on capable hardware.

## Why this is easy to fix

The backend enforces **no cap whatsoever**. Every server-side handler simply does:

```js
const messageLimit = workspace?.openAiHistory || 20;
```

...and passes the value directly to `recentChatHistory()`. The `compressMessages()` function already handles overflow gracefully by trimming to fit the model's actual context window. The 45-message ceiling exists **only** in the frontend slider. The database field accepts any integer.

## Proposed solution

This request has two parts — a quick fix and a more powerful enhancement.

### Part 1 — Replace the slider with a synced slider + numeric text input combo

The slider provides intuitive drag behavior while a free-type number field beside it allows precise entry. Both controls stay in sync: dragging the slider updates the number, typing a number repositions the slider. The ceiling should be raised to at least 1,000. No arbitrary cap, maximum flexibility. This is a standard UI pattern and an immediately shippable improvement requiring only frontend changes.

The one-line minimum fix in `frontend/src/pages/WorkspaceSettings/ChatSettings/ChatHistorySettings/index.jsx`:

```jsx
// Before
max={45}

// After
max={1000}
```

### Part 2 — Add an "Auto-size to model context window" checkbox

In addition to the manual control, add an optional checkbox: **"Automatically size history to model context window."**

**Behavior:**
- **Unchecked (default):** Manual slider + text input. Existing behavior, just with the raised ceiling. No change for users who want explicit control.
- **Checked:** AnythingLLM reads the workspace's configured model context window (which is already queried elsewhere in the codebase), subtracts a calculated headroom buffer, and derives the optimal history message count automatically. The slider becomes read-only, displaying the computed value so the user understands what's happening.

The headroom calculation would follow this logic:

```
available_history_tokens = model_context_window
                         - system_prompt_tokens
                         - (max_context_snippets × avg_chunk_size)
                         - output_token_headroom

suggested_history_messages = floor(available_history_tokens / avg_message_tokens)
```

Where `avg_message_tokens` could initially use a conservative default (e.g. 1,000 tokens/message) and optionally be refined over time using the workspace's actual chat history — making the estimate progressively more accurate the longer a workspace is used.

This is particularly powerful because AnythingLLM **already has the model context window available** for other purposes. This feature is connecting existing plumbing to the history setting, not building new infrastructure.

## Why both parts together matter

Part 1 unblocks power users immediately with a trivial change. Part 2 makes the right behavior the easy behavior — users who don't want to think about token math just check a box and AnythingLLM does the right thing automatically. Together they bring the history setting in line with the reality of modern large-context models.

## Additional context

Users running large-context models via API (Claude, OpenAI) or locally (via LM Studio, Ollama) on capable hardware are increasingly common. AnythingLLM's memory and personalization features are excellent, but they complement — not replace — raw conversation history for deep, ongoing work sessions. The 45-message cap is the last vestige of an era when 8K context was a ceiling, not a floor.

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5899 

### Description

Removes the `max={45}` cap on the chat history limit input field for workspace settings.

App still defaults a new workspace to a chat history limit of 20.

Verified ceiling isn't enforced anywhere else in the app.

### Developer Validations

### DIFF
--- a/frontend/src/pages/WorkspaceSettings/ChatSettings/ChatHistorySettings/index.jsx
+++ b/frontend/src/pages/WorkspaceSettings/ChatSettings/ChatHistorySettings/index.jsx
@@ -16,7 +16,6 @@ export default function ChatHistorySettings({ workspace, setHasChanges }) {
         name="openAiHistory"
         type="number"
         min={1}
-        max={45}
         step={1}
         onWheel={(e) => e.target.blur()}
         defaultValue={workspace?.openAiHistory ?? 20}


shatfield4 (approved):
LGTM

PR #5916, opened 2026-06-25
feat: create-scheduled-job agent tool

### LINKED ISSUE #5903: [FEAT]: Ability to schedule jobs via chat - like you would with an agent or open claw etc
Ability to schedule jobs via chat - like you would with an agent or open claw etc
 
e.g. if you asked question like 

“Can you do research for me? e.g. each week look up and find new tv series or movies for UK streaming services like Amazon, Apple TV, Disney +, NetFlix and Sky?”

Then it would automatically create a scheduled job to run it each week for you?

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)

### Relevant Issues

resolves #5903

Docs: https://github.com/Mintplex-Labs/anythingllm-docs/pull/272
Translations: #5932 

### Description

Adds a `create-scheduled-job` agent tool so users can create Scheduled Jobs conversationally in agent chat (e.g. *"every weekday at 9am summarize my inbox and email me"*). Single-user-mode only.

- **New agent tool** (`create-scheduled-job`): takes a structured schedule (frequency + local time) and the tools the job may use. The server converts local→UTC deterministically (`cronFromLocal.js`, via `Intl` — no extra deps), so the LLM never does timezone math or writes cron. `listTools` mode + validation ensures the agent only picks configured, ready-to-use tools (mirrors the manual UI). No tools selected = no tools, same as the manual form.
- **Timezone transport**: the browser's IANA timezone rides the existing HTTP→websocket invocation cache (generalized from attachments to a context object) so the tool knows the user's local time.
- **Enable/disable**: a toggle in the Agent Skill Settings modal (`agent_create_scheduled_job_enabled`), hidden in multi-user mode.
- **Chat card**: a clickable "job created" card appears in chat (like file-download cards) that deep-links to the job's edit modal via `?action=edit-job&jobId=`.

### Visuals (if applicable)
image
image

### Additional Information

Scheduled Jobs is single-user-mode only; the tool is gated out of multi-user mode at the tool-loading chokepoint plus defense-in-depth on the job-runner path.

### Developer Validations

## TODO
- [x] Translations PR
- [x] Docs


server/utils/scheduling/cronFromLocal.js
```diff
@@ -0,0 +1,122 @@
...
+    case "week": {
+      const utc = localToUtcHM(s.hour, minute, s.timezone);
+      // NOTE: weekday list is intentionally NOT shifted when the hour crosses
+      // midnight in UTC - this matches the frontend builder behavior exactly.
+      const days = [...new Set(s.weekdays?.length ? s.weekdays : [1])]
+        .map((d) => ((d % 7) + 7) % 7)
+        .sort((a, b) => a - b)
+        .join(",");
+      return `${utc.minute} ${utc.hour} * * ${days}`;
+    }
```
shatfield4: This same normalize + sort is duplicated in `create-scheduled-job.js`. Lets pull it into a `normalizeWeekdays(weekdays)` util here in `cronFromLocal.js`, export it, and reuse it in both spots. We will probably reuse this logic as we add more features to scheduled jobs.

server/utils/agents/aibitat/plugins/create-scheduled-job.js
```diff
@@ -0,0 +1,385 @@
...
+    case "hour":
+      return `every hour at :${String(minute).padStart(2, "0")}`;
+    case "day":
+      return `daily at ${hhmm(args.hour, minute)} ${timezone}`;
+    case "week": {
+      const days = [...new Set(args.weekdays?.length ? args.weekdays : [1])]
+        .map((d) => ((d % 7) + 7) % 7)
+        .sort((a, b) => a - b)
+        .map((d) => WEEKDAY_NAMES[d])
+        .join(", ");
```
shatfield4: Duplicated logic. Check `cronFromLocal.js` review comment.

server/utils/scheduling/cronFromLocal.js
```diff
@@ -0,0 +1,122 @@
...
+    case "month": {
+      const utc = localToUtcHM(s.hour, minute, s.timezone);
+      return `${utc.minute} ${utc.hour} ${s.dayOfMonth || 1} * *`;
+    }
+    default:
+      throw new Error(`Unsupported schedule frequency: ${s.frequency}`);
+  }
+}
+
+module.exports = { tzOffsetMinutes, localToUtcHM, buildUtcCronFromSchedule };
```
shatfield4: `tzOffsetMinutes` and `localToUtcHM` seem to only be used in this file. Let's remove it from exports.

shatfield4 (changes requested):
Great job! This functions and feels really good to me. Just a few small nits/areas we can clean up.

One other thing is that I know you made this an option to configure inside the Agent Skill Settings modal. Should this just be a default skill so that users can enable/disable it from the tools popup on the chat page?

server/utils/agents/aibitat/plugins/create-scheduled-job.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5916/commits/c8f7557c08ca40b024ae90eb3646bacbbdef878d

server/utils/scheduling/cronFromLocal.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5916/commits/c8f7557c08ca40b024ae90eb3646bacbbdef878d

server/utils/scheduling/cronFromLocal.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/5916/commits/c8f7557c08ca40b024ae90eb3646bacbbdef878d

timothycarambat:
@angelplusultra Can we move this to the real skills menu and just hide if in single user mode? We do this with the app integrations section already.

Screenshot 2026-06-26 at 10 42 01 AM

Lets tuck it under SQL connections.

shatfield4 (approved):
Nice, LGTM.

timothycarambat:
Changes I made/had to make:

The original plugin took structured inputs (frequency, hour, minute, etc.) and did server-side timezone conversion via a dedicated `cronFromLocal.js` utility, threading `timezone` through chat endpoints, stream handlers, agent handlers, and frontend models - a lot of plumbing for something the model can just write itself.

- Rewrote the plugin to accept a single 5-field cron string from the model in the user's **local time**. The server auto-converts it to UTC.
- Deleted `server/utils/scheduling/cronFromLocal.js` and its associated `timezone` threading through `chat.js`, `stream.js`, `agents.js`, `index.js`, and the frontend models (`workspace.js`, `workspaceThread.js`).

Too much prop passing for something we can hold in a state manager since its basically like a few bytes.


After simplifying, the server still needs to know the user's timezone to convert the local-time cron to UTC without threading it through every request body.

- Added `X-Timezone` and `X-Language` headers to `frontend/src/utils/request.js` `baseHeaders()` so every authenticated request carries the browser's locale implicitly.
- Created `server/utils/userLocale.js` as a singleton `UserMetaCache` class with `get(userId)`, `set(userId, data)`, and `setFromRequest(userId, request)` methods, storing `{ timezone, language }` per user (key `"primary"` in single-user mode).
- Updated `server/utils/middleware/validatedRequest.js` to call `UserMetaCache.setFromRequest()` in all three authentication paths (dev, single-user, multi-user), so the cache is always populated before any handler runs.

Now we have a global cache for users we can lookup to get timezone or language pref.

The plugin was a single flat file with helper functions mixed in, and the cron utilities had no clear home. Just broke those up into utils, no big deal


After creating a job, the UI navigated to the edit modal instead of the runs list. Would make more sense to go to runs imo - just a nit.

The `create-scheduled-job` skill is single-user only, there was some code guards here but it was a lot of code it seemed like to blocking the UI for this and I just simplified it to look like other modules.

 If the model hallucinates or misinterprets the user's request, it could create background scheduled jobs without the user's explicit confirmation. Just added a `requestToolApproval` call before creation.

The agent was repeatedly calling the plugin with guessed tool IDs (e.g. `"functions.web-scraping"`) and then hallucinating a failure message instead of retrying with `listTools: true`. Now its a two step tool so that this doesnt happen and is not optional. It would refuse randomly for some SLMs until i traced that.

Issue #5926, opened 2026-06-29
[BUG]: AnythingLLM UI does not show any Kokoro Voice Models


F1zzyD:
### How are you running AnythingLLM?

AnythingLLM desktop app

### What happened?

When attempting to connect to my Kokoro instance running as a docker container on my host machine, ALLM refuses to show any voice models in the UI. Kokoro sees that ALLM is requesting data, however it refuses to correctly display the voice models. This makes it impossible to save any settings and utilize Kokoro.

Shashank200345:
interesting would like to work on this .

timothycarambat:
> When attempting to connect to my Kokoro instance running as a docker container on my host machine

Are you able to hit the models endpoint in your browser? If so what does that endpoint look like and how are you running the container? We tested it against the kokoro TTS FastAPI service - but your container might be something different and have an different API endpoint

sanidhyasin:
One thing that could explain this (alongside the container question above): `kokoro-fastapi` changed the shape of `GET /v1/audio/voices` in `0.3.x`.

- **< 0.3.x:** `{ "voices": ["af_bella", "af_sarah", ...] }` — plain id strings
- **>= 0.3.x:** `{ "voices": [{ "id": "af_bella", "name": "af_bella" }, ...] }` — objects

AnythingLLM currently reads `voice.id`/`voice.name` only, so against the older string shape every entry becomes `{ id: undefined, name: undefined }` — the dropdown ends up non-empty but blank/unselectable, which lines up with "Kokoro sees the request but the voices won't display."

@F1zzyD to confirm which case you're in: if you open `http://<your-kokoro-host>:8880/v1/audio/voices` in a browser, what does it return? That'll show whether it's the legacy string shape or a different container/endpoint as @timothycarambat mentioned. I opened #5930 to normalize both shapes, which would cover the former.

timothycarambat:
resolved by https://github.com/Mintplex-Labs/anything-llm/pull/5930

PR #5679, opened 2026-05-22
Kokoro TTS provider

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)
- [ ] ♻️ refactor (Code refactoring without changing behavior)
- [ ] 💄 style (UI style changes)
- [ ] 🔨 chore (Build, CI, maintenance)
- [ ] 📝 docs (Documentation updates)

### Relevant Issues

resolves #5678

### Description

- Implements Kokoro as a TTS provider
- Works with [Kokoro-FastAPI](https://github.com/remsky/Kokoro-FastAPI) since this is the easier way for users to get it up and running quickly with the Kokoro model in an OpenAI compatible API endpoint

### Developer Validations

- [x] I ran `yarn lint` from the root of the repo & committed changes
- [x] Relevant documentation has been updated (if applicable)
- [x] I have tested my code functionality
- [x] Docker build succeeds locally

PR #5930, opened 2026-06-29
fix: render kokoro-tts voice list for pre-0.3.x kokoro-fastapi servers

sanidhyasin:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

connect #5926

### Description

A backwards-compatibility fix for the Kokoro TTS voice list. `kokoro-fastapi` changed the response shape of its `GET /v1/audio/voices` endpoint in `0.3.x`:

- **< 0.3.x** returns voices as plain id strings: `{ "voices": ["af_bella", "af_sarah", ...] }`
- **>= 0.3.x** returns objects: `{ "voices": [{ "id": "af_bella", "name": "af_bella" }, ...] }` (this is what the integration in #5679 was written and tested against)

`kokoroTtsVoices()` in `server/utils/helpers/customModels.js` only reads `voice.id` and `voice.name`. Against the older string shape, every entry becomes `{ id: undefined, name: undefined }`. Because the resulting array is non-empty, the frontend doesn't fall back to its "enter a voice id manually" input — it renders a `<select>` full of blank, unselectable options, so no voice can be chosen or saved.

This PR normalizes both response shapes to `{ id, name }` so the voice list loads regardless of the `kokoro-fastapi` version, falling back to the voice id when an object omits its `name`, and dropping any malformed entries. The object path that currently works is unchanged.

### Visuals (if applicable)

N/A — restores population of the existing "Voice Model" dropdown.

### Additional Information

- **Scope / relation to #5926:** This addresses the documented pre-`0.3.x` string-response shape, which is one concrete way the voice list can come back blank. Whether it is the cause of #5926 specifically is still being triaged — that reporter may be running a non-standard container with a different endpoint, in which case this would not resolve their report but would still benefit users on older `kokoro-fastapi` builds. Marked `connect` rather than `resolves` for that reason.
- Root cause for the string-shape case is confirmed against `kokoro-fastapi`'s own endpoint, whose docstring documents the `0.3.x` change and exposes `?legacy=true` for the old plain-string shape.
- The TTS playback path (`server/utils/TextToSpeech/kokoro/index.js`) is unchanged: the stored `TTSKokoroVoiceModel` is always a plain voice id string, which both shapes resolve to.
- Note on validations: I did not run a full `yarn install` locally (heavy dependency tree), so `yarn lint` / `yarn test` were not run on my machine. The change follows the file's existing prettier formatting, and the added Jest tests live under `server/__tests__/` so they run via the existing backend test workflow on this PR.

### Developer Validations


timothycarambat:
@sanidhyasin TY 🚀 Will review tomorrow

PR #5382, opened 2026-04-07
Fix streaming issue for LLM instruction blocks

timothycarambat:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)
- [ ] ♻️ refactor (Code refactoring without changing behavior)
- [ ] 💄 style (UI style changes)
- [ ] 🔨 chore (Build, CI, maintenance)
- [ ] 📝 docs (Documentation updates)

### Relevant Issues

resolves #5374

### Description
- Fix issue with Anthropic SDK not liking completions for LLM Instructions causing a flow to fail

### Developer Validations

- [ ] I ran `yarn lint` from the root of the repo & committed changes
- [ ] Relevant documentation has been updated (if applicable)
- [ ] I have tested my code functionality
- [ ] Docker build succeeds locally

PR #5941, opened 2026-06-30
fix: stream Anthropic non-streaming chat completions for REST API workspace chat

### LINKED ISSUE #5924: [BUG]: Meeting summary fails with "s.includes is not a function" when default LLM is Anthropic (Claude)

AnythingLLM desktop app

### What happened?

**AnythingLLM Desktop v1.15.0 (macOS, Apple Silicon)**

The Meeting Assistant **transcribes audio successfully**, but the **summary generation step always fails** when the system **Default LLM** (Settings → AI Providers → LLM Provider) is set to an **Anthropic Claude** model.

The user-facing push notification is:

> **Failed to summarize meeting** — "We failed to summarize your meeting. Error: `s.includes is not a function`"

#### Key evidence from `backend-*.log`

```
[summary Recording:564ca121-...] [Job] Starting job (summary) for UUID: 564ca121-...
   Will use LLM provider: AnthropicLLM::claude-opus-4-8
[summary Recording:564ca121-...] [Job] Generating summary for UUID: 564ca121-...
[summary Recording:564ca121-...] [Job] Context window: 1000000 tokens. System tokens: 78 tokens. Available tokens: 699922 tokens. Transcript tokens: 17238 tokens.
[summary Recording:564ca121-...] [Job] Error generating summary for UUID: 564ca121-...
[summary Recording:564ca121-...] {"type":"send-push-notification","payload":{"title":"Failed to summarize meeting","body":"We failed to summarize your meeting. Error: s.includes is not a function"}}
```

#### This is NOT a context-window problem

The maintainer's prior diagnosis on the similar (now-closed) issue #5452 was that "the context window is too small for the meeting." The log above **disproves that for this case**: the job reports a **1,000,000-token** context window with only **17,238 transcript tokens** and **699,922 tokens available**. There is plenty of headroom. The failure is a **JavaScript `TypeError`**, not a token-limit rejection.

#### This is model-independent within Anthropic

I reproduced the identical error on **two different Anthropic models**:
- `claude-sonnet-4-6` (a.k.a. `claude-sonnet-4-5`)
- `claude-opus-4-8`

Both fail with the exact same `s.includes is not a function`. Switching models within the Anthropic provider does **not** help, which rules out the `noTemperatureModels` / temperature-parameter path as the cause.

#### Likely location

The error string `s.includes is not a function` indicates that a value expected to be a **string** is actually a non-string (object/array/undefined) when `.includes()` is called on it, somewhere in the **`AnthropicLLM`** meeting-summary code path (`server/utils/AiProviders/anthropic/index.js` and/or the summary `Job`). The regular **chat** path (which passes plain strings) works fine with the same Anthropic model — only the **Meeting Assistant summary Job** triggers the crash, suggesting the summary job passes the transcript/content in a shape the Anthropic provider's string handling does not expect.

Notably, per issue #5452's logs, the summary job **succeeds** when the provider is `GenericOpenAiLLM` (it reaches "Summary generated"), so the defect appears **specific to the Anthropic provider path**.

1. Set Settings → AI Providers → **LLM Provider** to **Anthropic** with a Claude model (reproduced on `claude-sonnet-4-6` and `claude-opus-4-8`).
2. Open **Meeting Assistant**, record/transcribe a short meeting (transcription succeeds — ~17k tokens in this case).
3. Let the summary job run (or click **Regenerate summary**).
4. Observe the push notification: **"We failed to summarize your meeting. Error: `s.includes is not a function`"**.
5. Check `~/Library/Application Support/anythingllm-desktop/storage/logs/backend-*.log` for the `[Job] Error generating summary` lines shown above.

**Workaround:** Setting the default LLM to a non-Anthropic provider (e.g., a local OpenAI-compatible / `GenericOpenAiLLM` endpoint) allows the summary to generate successfully.

Anthropic — reproduced on `claude-sonnet-4-6` and `claude-opus-4-8`

_No response_

### LINKED ISSUE #5925: [BUG]: REST API workspace chat returns empty textResponse for Anthropic-backed workspaces (non-streaming SDK throw at apiChatHandler.chatSync)

Docker (local)

### What happened?

REST API workspace chat (`POST /api/v1/workspace/{slug}/chat` with `"mode": "chat"`) silently returns an empty response when the workspace is configured with an Anthropic provider override (Claude Sonnet 4.6 in my case).

The HTTP response is structurally well-formed but content-empty:

```json
{
  "id": "e56bfac4-21fa-4c16-833c-23107d5daff6",
  "type": "textResponse",
  "close": true,
  "error": null,
  "chatId": 39,
  "textResponse": {},
  "sources": [],
  "metrics": {}
}
```

`error: null` and `textResponse: {}` (object, not string) make this look like a successful empty response on the wire. The real failure is server-side and visible only in container logs:
[backend] info: [AnthropicLLM] Initialized with claude-sonnet-4-6. Cache enabled (5m)
[backend] info: [TokenManager] Returning existing instance for model: claude-sonnet-4-6
[backend] info: [AnthropicLLM] Initialized with claude-sonnet-4-6. Cache enabled (5m)
[backend] info: [TokenManager] Returning existing instance for model: claude-sonnet-4-6
[backend] info: [AnthropicLLM] Model claude-sonnet-4-6 max tokens: 128000
[backend] info: Error: Streaming is strongly recommended for operations that may take longer than 10 minutes. See https://github.com/anthropics/anthropic-sdk-python#streaming-responses for more details

at Anthropic._calculateNonstreamingTimeout (/app/server/node_modules/@anthropic-ai/sdk/core.js:284:19)
at Messages.create (/app/server/node_modules/@anthropic-ai/sdk/resources/messages/messages.js:46:54)
at AnthropicLLM.getChatCompletion (/app/server/utils/AiProviders/anthropic/index.js:217:33)
at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
at async Object.chatSync (/app/server/utils/chats/apiChatHandler.js:421:5)
at async /app/server/endpoints/api/workspace/index.js:696:24
[backend] info: [Event Logged] - api_sent_chat

**Root cause:** the Anthropic SDK refuses non-streaming requests whose `max_tokens` is high enough that the worst-case latency could exceed 10 minutes. The REST API workspace chat path (`chatSync` → `AnthropicLLM.getChatCompletion`) calls `Messages.create()` non-streaming with `max_tokens` set from the model's max output (128000 for Claude Sonnet 4.6), which trips the SDK's guard. The thrown error is caught somewhere upstream and silently flattened into an empty `textResponse`, so callers see "success with empty content" rather than a meaningful error.

**Relationship to existing work:** this is the same SDK throw as #5374, which was fixed in #5382 by switching the Agent Flow LLM Instruction path to streaming. That fix patched `AnthropicProvider.complete` in `server/utils/agents/aibitat/providers/anthropic.js`. The workspace-chat REST API path lives in `server/utils/AiProviders/anthropic/index.js` and was not covered by #5382 — it still calls the SDK non-streaming and still hits the guard for any recent Claude model whose max output tokens are large.

**Impact:** any user invoking the REST workspace-chat endpoint against an Anthropic-backed workspace receives `textResponse: {}` instead of model output. The UI path is unaffected because it streams. The bug is invisible to UI testing — the workspace appears to work in chat mode in the UI, then silently fails when the same workspace is accessed via the API.

1. Create a workspace and override its LLM provider to Anthropic, with Claude Sonnet 4.6 (or any recent Claude model with large max output tokens) selected.
2. Confirm the workspace works in the UI in chat mode — returns real responses.
3. Issue an API call:
curl -X POST http://<host>:3001/api/v1/workspace/<slug>/chat 
-H "Authorization: Bearer <api-key>" 
-H "Content-Type: application/json" 
-d '{"message":"hello","mode":"chat"}'

4. Observe response: `{"textResponse": {}, "metrics": {}, "error": null, ...}` — empty content, no error indication on the wire.
5. Inspect server logs: `_calculateNonstreamingTimeout` throw visible there.

**Control:** the same API call against a workspace using a non-Anthropic provider (Generic OpenAI / Ollama / OpenAI) returns content correctly. Confirms the bug is Anthropic-provider-specific and not a general API or auth issue.

**Expected behavior**

API workspace chat against an Anthropic-backed workspace returns the model's response, the same as the UI path does.

**Suggested direction (not prescriptive)**

The fix pattern from #5382 — switch the call site to the SDK's streaming method and assemble the result — would apply directly. Alternatively, capping `max_tokens` below the SDK's threshold before the non-streaming call would also resolve the immediate throw, though streaming is the more durable answer for parity with the UI behaviour.

A secondary issue worth noting independent of the primary fix: the empty-response-on-error masking in `apiChatHandler.chatSync` makes this bug invisible without container logs. Surfacing the error in the API response (`error: "..."` rather than `error: null`) would shorten future diagnosis significantly.

Anthropic / Claude Sonnet 4.6 (reproduces on any recent Claude model — the trigger is large `max_tokens`, not the specific model)

Not relevant to this bug — bug occurs before any embedding/RAG involvement, in the LLM call itself.

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5925
resolves #5924 

### Description

The REST API workspace chat (`POST /api/v1/workspace/{slug}/chat` with `"mode": "chat"`) returned an empty `textResponse: {}` with `error: null` for Anthropic-backed workspaces, while the UI worked fine.

Two causes, both in `server/utils/AiProviders/anthropic/index.js`:

1. **Non-streaming SDK guard.** `getChatCompletion` called `messages.create` non-streaming with `max_tokens` set to the model's full output ceiling (128k for Claude Sonnet 4.6). The Anthropic SDK refuses non-streaming requests whose worst-case latency could exceed 10 minutes (`_calculateNonstreamingTimeout`), so the call threw. The UI was unaffected because it streams. Fixed by assembling the response from the streaming endpoint (`messages.stream(...).finalMessage()`), which bypasses the guard, same approach as #5382.

2. **Error masking.** The `catch` returned the `Error` object as `textResponse`, which serialized to `{}` and left `error: null`, hiding the real failure. It now throws (matching the other providers, e.g. Gemini), so the endpoint's existing handler surfaces a real error message.

### Additional Information

`max_tokens` is left as-is — it is a required field for the Anthropic Messages API and acts only as an upper bound, so it is harmless on the streaming path.

### Developer Validations


shatfield4 (approved):
LGTM

PR #5949, opened 2026-07-01
Update default fallback models for OpenAI/Anthropic

### LINKED ISSUE #5948: Update model defaults for OpenAI/Anthropic

Some provider defaults still fall back to `gpt-3.5-turbo` and other older/pricier models when nothing is configured. We should modernize these:

- OpenAI defaults → `gpt-4.1-nano`
- Anthropic defaults → `claude-haiku-4-5` (cheapest current Claude per token)

These only apply as a last resort when no model preference or env var is set, so configured setups are unaffected.

shatfield4:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)
- [x] 🔨 chore (Build, CI, maintenance)

### Relevant Issues

resolves #5948

### Description

- Change OpenAI default fallback models to `gpt-4.1-nano` (previously `gpt-3.5-turbo` / `gpt-4o`)
- Change Anthropic default fallback models to `claude-sonnet-4-6`, the cheapest current Claude model per token (previously older Claude 3 defaults)

### Developer Validations

### DIFF
--- a/server/models/systemSettings.js
+++ b/server/models/systemSettings.js
@@ -854,7 +854,7 @@ const SystemSettings = {
     return {
       // OpenAI Keys
       OpenAiKey: !!process.env.OPEN_AI_KEY,
-      OpenAiModelPref: process.env.OPEN_MODEL_PREF || "gpt-4o",
+      OpenAiModelPref: process.env.OPEN_MODEL_PREF || "gpt-4.1-nano",
 
       // Azure + OpenAI Keys
       AzureOpenAiEndpoint: process.env.AZURE_OPENAI_ENDPOINT,
@@ -868,7 +868,7 @@ const SystemSettings = {
       // Anthropic Keys
       AnthropicApiKey: !!process.env.ANTHROPIC_API_KEY,
       AnthropicModelPref:
-        process.env.ANTHROPIC_MODEL_PREF || "claude-sonnet-4-6",
+        process.env.ANTHROPIC_MODEL_PREF || "claude-haiku-4-5",
       AnthropicCacheControl: process.env.ANTHROPIC_CACHE_CONTROL || "none",
 
       // Gemini Keys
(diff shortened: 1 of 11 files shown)


server/models/systemSettings.js
```diff
@@ -868,7 +868,7 @@ const SystemSettings = {
       // Anthropic Keys
       AnthropicApiKey: !!process.env.ANTHROPIC_API_KEY,
       AnthropicModelPref:
-        process.env.ANTHROPIC_MODEL_PREF || "claude-sonnet-4-6",
+        process.env.ANTHROPIC_MODEL_PREF || "claude-haiku-4-5",
```
angelplusultra: For some reason claude-haiku-4-5 isnt an actual valid model id for Anthropic. You need to use a specific snapshot version such as `claude-haiku-4-5-20251001`.

Anywhere that fallsback to haiku should change to this value.

angelplusultra (changes requested):
All the OpenAI stuff looks good, but the new Anthropic default has a settings-UI problem: `claude-haiku-4-5` is an alias,  it works for chat requests, but Anthropic's `/v1/models` endpoint only lists the dated ID `claude-haiku-4-5-20251001`, and the settings dropdown pre-selects by strict equality against listed IDs.

Repro:
 1. Set up Anthropic, then comment out `ANTHROPIC_MODEL_PREF` so the system falls back to `claude-haiku-4-5`
 2. Go to Settings → LLM Preference and notice the selected model shows sonnet (the first model in the list), not haiku
 3. Saving the form from this state persists the sonnet model, silently switching the user off the intended default

 The old default `claude-sonnet-4-6` didn't hit this because current-gen models have no dated variant, the alias is the listed ID. Haiku 4.5 is the one current model still on the dated-snapshot scheme.

 Fix: use `claude-haiku-4-5-20251001` as the default here and in `AiProviders/anthropic/index.js:41` and `aibitat/providers/anthropic.js:25`.

server/models/systemSettings.js
shatfield4: Decided to change all Anthropic fallbacks to use `claude-sonnet-4-6` for simplicity.

angelplusultra (approved):
LGTM

PR #5967, opened 2026-07-06
Give Downloaded Models its Own `<optgroup>` in Workspace Model Selector

### LINKED ISSUE #5960: [FEAT]: Lemonade LLM provider - only show downloaded models, not all discovered models

When using Lemonade as LLM provide it is listing ALL discovered models, even if they haven't actually been downloaded and installed in Lemonade.

Either show discovered and downloaded separately, or have option in setup to only should models which have actually been downloaded and are currently available via Lemonade API.

This will be a lot easier to use, as don't need to then scroll through a great long list simply to find the 1 or 2 models I have downloaded in Lemonade to use.

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5960

### Description

Adds a **Downloaded Models** option group to the top of the workspace model selection dropdown. This group only appears when downloaded models are detected.

### What Constitutes a “Downloaded Model”

Local AI providers such as Ollama and LM Studio only return models that are already downloaded locally, so those models will not appear in this option group.

However, providers such as Lemonade and Docker Model Runner return **all available models from the provider**, along with a `downloaded` boolean. These are the providers impacted by this feature and will have the "Downloaded models" `<optgroup>`

### Developer Validations

### DIFF
--- a/frontend/src/components/EmbeddingSelection/LemonadeOptions/index.jsx
+++ b/frontend/src/components/EmbeddingSelection/LemonadeOptions/index.jsx
@@ -196,6 +196,8 @@ function LemonadeModelSelection({ settings, basePath = null }) {
     findCustomModels();
   }, [basePath]);
 
+  const downloadedModels = customModels.filter((model) => model?.downloaded);
+
   if (loading || customModels.length == 0) {
     return (
       <div className="flex flex-col w-60">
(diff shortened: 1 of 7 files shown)


shatfield4 (approved):
LGTM, shows downloaded models at the top for both Lemonade and Docker Model Runner providers.

timothycarambat (changes requested):
This works but does not resolve the same UX pattern in several other key areas of the app. If we are going to fix it in one place we should go and fix it in others as well.

Workspace Chat Model and Workspace Agent Selection dropdowns are two areas that have this issue. I would imagine that the Model Router does as well since it uses dropdowns.

This can be abstracted out to a util or something if we use it in that many places though.

PR #5970, opened 2026-07-07
Fix: Restore Stop Generation Button During Agent Session Execution Loop

### LINKED ISSUE #5969: [BUG]: `<StopGenerationButton />` Disappears on The First Prompt That Initiates an Agent Session

Docker (local)

### What happened?

When initiating a new agent session, such as `@agent tell me a joke`, or simply `tell me a joke` when the workspace is already set to Agent mode, the `<StopGenerationButton />` does not appear for the entire duration of the initial execution loop.

If you send a subsequent prompt while the agent session is still active, the button appears as expected. However, if you end the session with `/exit` and then send a new prompt to initialize a fresh agent session, the issue repeats.

### Steps to Reproduce

1. Ensure the workspace is in Agent mode, or prefix the prompt with `@agent`.
2. Go to the empty chat view for the workspace:
   `http://localhost:3000/workspace/:workspace-slug`
3. Send the prompt `tell me a joke`, or `@agent tell me a joke` if the workspace is not already in Agent mode.
4. Observe that where the abort button (`<StopGenerationButton />`) should appear, the send prompt button remains visible, greyed out, and disabled.
5. Type into the prompt box. Notice that the send button becomes enabled, allowing another prompt to be sent while the agent is still streaming.
6. Once the initial agent loop finishes, send another prompt to trigger a new loop.
7. Notice that the abort stream button appears correctly on subsequent loops.

This issue appears to occur only during the first execution loop of a newly initialized agent session from the empty chat view.

_No response_

_No response_

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5969

### Description

`<StopGenerationButton />` never appeared during the first execution loop of a new agent session. Two causes:

1. The HTTP handoff that starts an agent session ends with a `statusResponse` ("Swapping over to agent chat"), and `handleChat` treats every `statusResponse` as end-of-exchange, clearing `loadingResponse` right as the agent starts working.
2. The websocket message listener called `setLoadingResponse(true)` then `setLoadingResponse(false)` in the same synchronous handler, which React 18 batches into a net `false` on every incoming agent event — so the loading state could never recover.

Fix (frontend only):

- `loadingResponse` is set on when the agent websocket is established, and the closing handoff `statusResponse` is no longer allowed to clear it once the session is active (guards against either ordering of the handoff chunks).
- Each socket event now maps to a loading state via `agentEventLoadingState()` in `utils/chat/agent.js`: work events (status updates, text chunks, tool calls) keep it on; awaiting-user events (`WAITING_ON_INPUT` — previously ignored by the frontend entirely — plus `toolApprovalRequest`/`clarificationRequest`) turn it off; passive bookkeeping events (`rename_thread`, `chatId`, `usageMetrics`, `citations`, etc.) leave it untouched. The passive bucket matters because `rename_thread` arrives after the async chat save + thread auto-rename complete, i.e. after `WAITING_ON_INPUT`, and would otherwise re-show the stop button indefinitely.

The stop button now stays visible for the entire execution loop and hands back to the send button when the agent is waiting on the user.

### Developer Validations


frontend/src/utils/chat/agent.js
```diff
@@ -8,6 +8,49 @@ import { THREAD_RENAME_EVENT } from "@/components/Sidebar/ActiveWorkspaces/Threa
...
+// of these can arrive after the execution loop already finished (e.g.
+// rename_thread fires once the async chat save + thread auto-rename complete),
+// so they must not re-show the stop generation button.
+const AGENT_PASSIVE_EVENTS = ["rename_thread"];
+const AGENT_PASSIVE_STREAM_EVENTS = [
+  "chatId",
+  "usageMetrics",
+  "citations",
+  "removeStatusResponse",
+  "modelRouteNotification",
```
shatfield4: Nit: this only fires at the start of the loop when loading is already true so it is not really doing anything. Not a huge problem but let's remove it just to simplify code a tiny bit more.

shatfield4 (changes requested):
Just 1 small nit here, other than that LGTM. It works and solves the bug in a simple way.

timothycarambat:
@angelplusultra can we get this moving?

timothycarambat (changes requested):
Noting this because i was testing this prior to review:

With LMStudio, if I kill the response when generating a response - the UI properly shows "Agent session ended" and the button returns to a sendable state. However on the backend the message continues.

The intention for this PR is JUST to fix the button state, yeah? Because on reload of the page the completed message is present.

If this PR is JUST for button state, we need to stack PRs as merging just this to mater results in still a misleading state and we should merge this fix + actual stopping generation in a single commit.

angelplusultra:
> Noting this because i was testing this prior to review:
> 
> With LMStudio, if I kill the response when generating a response - the UI properly shows "Agent session ended" and the button returns to a sendable state. However on the backend the message continues.
> 
> The intention for this PR is JUST to fix the button state, yeah? Because on reload of the page the completed message is present.
> 
> If this PR is JUST for button state, we need to stack PRs as merging just this to mater results in still a misleading state and we should merge this fix + actual stopping generation in a single commit.

Yes, the intention of this PR is just to restore the functionality of the abort feature as is. The state of the button was lost somehow in a v2 migration. We are splitting the true abort feature into 2 steps. First fix it as is. Then refactor to be a full end-to-end abort.

shatfield4 (approved):
LGTM

PR #5978, opened 2026-07-08
(OpenComputer) Update model version in use case demos

FI-Mihej:
Updated the model version in use case demos for consistency.

;)

I'm actually watching your video right now, by the way. Nice! :)

By the way, my plan is to build a virtualized environment for running LLM-generated code - something like llm-sandbox - once I release a couple of other tools (should be in the next few days). The idea is to use lightweight virtual machines rather than QEMU. Although it looks like on macOS I'll probably have to borrow some of your experience with QEMU after all. If that's the case, I'll definitely reach out if I find or fix any bugs.

Also, feel free to message me if you ever need to test something. I have a Windows on ARM laptop, so I can run things whenever I have a few spare minutes.

Cheers!

P.S. No need to worry about merging this PR if it's not worth the hassle. My main goal was just to let you know about the issue that immediately stood out to me. You can fix it however you prefer, even with a direct commit. :)

### Pull Request Type

- [ ] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #

### Developer Validations

### DIFF
--- a/open-computer/README.md
+++ b/open-computer/README.md
@@ -29,13 +29,13 @@ AI agents are most capable in a Linux shell, but most people truthfully dont fee
 ## Use Case Demos
 
 <video src="https://github.com/user-attachments/assets/cd5363d9-1daf-4867-8d97-d7fd903ad1b5"></video>
-<p align="center"><em>Open Computer building a data visualization from an uploaded CSV file | LM Studio Gemma 4 13B QAT on MacBook Pro M4 Pro for inference</em></p>
+<p align="center"><em>Open Computer building a data visualization from an uploaded CSV file | LM Studio Gemma 4 12B QAT on MacBook Pro M4 Pro for inference</em></p>
 
 <video src="https://github.com/user-attachments/assets/cf8444d8-15f9-489d-8800-fb25a1769804"></video>
 <p align="center"><em>Open Computer building a data visualization from web research | OpenRouter Qwen 3.6 35B A3B</em></p>
 
 <video src="https://github.com/user-attachments/assets/79334c87-c5ae-4c2c-8384-d7ef922e4184"></video>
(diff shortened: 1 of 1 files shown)


timothycarambat:
Man i made that typo everywhere!! Great catch, was going to make some more demos anyway to show the "Human in Loop" use case. 

For sure if you have any questions about the VM stuff let me know. What were you going to use for VMs? Firecracker?

FI-Mihej:
😊👍

Yes - on Linux, definitely Firecracker. AWS has clearly demonstrated how powerful it is. On Windows, I'm looking at the Host Compute System (a system behind both WSL2 and Windows Sandbox (WSB)). There are some licensing considerations (it requires Win Pro, so on other editions we'd need to fall back to Docker), as well as some details around configuring networking restrictions and similar settings. I'll dig into those in depth later.

By the way, for the Win, Windows Sandbox (WSB) might potentially be a good fit for you. It would eliminate the need to ship a QEMU image with your application, since it already provides a ready-to-use Windows image.

But those are just preliminary thoughts for now. At the moment, I'm focused on getting the release out. :)

timothycarambat:
> Windows Sandbox

I looked into this, from what I can find there was two issues (at least from my limited research):
- Sandbox is not built for persistence, so starting/stopping was a requirement for OpenComputer so we can "sleep" them when not in use without blowing away all the work, memories, and apps

- Sandbox does not have an API (?) - so getting a bunch of stuff crammed into the sandbox would have to be done via powershell script that downloads/installed **every** boot - which would take a long time for the instance to go from "available" to "useful".

FI-Mihej:
Ah! Yes - it's definitely a blocker. In this situation, a universal solution using QEMU indeed looks much more attractive than a native solution. I was simply looking at it from my own perspective: the feature of self-cleaning lightweight VMs is particularly useful to me, and in my case, carrying around an OS image is overkill. However, for a class of applications like AnythingLLM, it is perfectly acceptable.

PR #6010, opened 2026-07-17
refactor: remove workspace profile picture feature

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Description

Removes the workspace profile picture feature entirely. The server endpoints (`/workspace/:slug/pfp`, `/workspace/:slug/upload-pfp`, `/workspace/:slug/remove-pfp`) and their frontend client methods existed, but no UI component ever called them — the feature was dead code with no user-facing impact.

Changes:
- Drop the three workspace pfp endpoints from `server/endpoints/workspaces.js`
- Remove `uploadPfp`/`fetchPfp`/`removePfp` from `frontend/src/models/workspace.js`
- Remove `determineWorkspacePfpFilepath` helper from `server/utils/files/pfp.js`
- Drop the `pfpFilename` column from the `workspaces` table (schema + migration)

User profile pictures (`/system/*pfp` endpoints, `users.pfpFilename`) are a separate feature and are untouched.

### Visuals (if applicable)

N/A — no UI changes.

### Additional Information

Includes a Prisma migration that drops `workspaces.pfpFilename`; any previously uploaded workspace pfp filenames are discarded on migrate.

### Developer Validations


server/prisma/migrations/20260717185508_init/migration.sql
timothycarambat, line 1: Do not do this, if the column cannot be dropped without having to drop and recreate the `workspaces` table then it will just remain an unused column and keep it null which is cheap.

Migrations of full drop/creates on SQLite should be avoided at *all costs* because if the migration does fail for any reason the entire application will break and this is not a change worth taking that risk on. Considering someone could have 5K workspaces and then then downstream associated records to that - we cannot do this op.

server/prisma/schema.prisma
```diff
@@ -134,7 +134,6 @@ model workspaces {
   chatModel                    String?
   topN                         Int?                           @default(4)
   chatMode                     String?                        @default("chat")
-  pfpFilename                  String?
```
timothycarambat, line 137: Referring to above - if we cannot just drop the col (likely cannot) then keep this record so generation schema does not change but then add a comment about this field noting it is deprecated and no longer used.

timothycarambat (changes requested):
See prisma comments - we may not be able to do a schema change here.

server/prisma/migrations/20260717185508_init/migration.sql
angelplusultra: sounds good, ill remove the schema change + migration.

server/prisma/schema.prisma
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6010/commits/43f4daf4171be92ead59d7ab2a4b5bf36bbbbed1

server/prisma/migrations/20260717185508_init/migration.sql
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6010/commits/43f4daf4171be92ead59d7ab2a4b5bf36bbbbed1

shatfield4 (approved):
LGTM

timothycarambat (changes requested):
https://github.com/Mintplex-Labs/anything-llm/blob/refactor-remove-workspace-pfp/frontend/src/components/UserIcon/index.jsx#L4-L21

I dont think we need this fallback/SVG anymore?

angelplusultra:
> https://github.com/Mintplex-Labs/anything-llm/blob/refactor-remove-workspace-pfp/frontend/src/components/UserIcon/index.jsx#L4-L21
> 
> I dont think we need this fallback/SVG anymore?

Good catch, we don't need those SVGs or the `<UserIcon />` and `<ChatBubble />` components anymore. The SVGs were only used by `<UserIcon />` and `<UserIcon />` only used by `<ChatBubble />`, which is not referenced anywhere in the app. It's safe to delete all of it.

https://github.com/Mintplex-Labs/anything-llm/pull/6010/commits/d9a26fda1d33854c9ac674822c0467e527ad7aff

PR #6015, opened 2026-07-18
fix: allow users to exit auto-scroll while a response is streaming

### LINKED ISSUE #5846: [BUG]: when scrolling up, scrolling jumps

Docker (local)

### What happened?

when scrolling through past messages, the scroll progress isnt linear but jumps. 

i imagine "its not a bug, but a feature" - the feature it tries to solve is more conveniently solved by horizontal bars on the  window side next to the scrollbar, which take the user back to the last message start. (ie one bar per message, colorcoded by prompt and response)

linearity of scrolling is important especially in long prompts/responses.

have several very long prompts / responses and try to scroll to the top of the last response - bam: back at the top of the last prompt.

_No response_

_No response_

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5846

### Description

While a response is streaming, users could not scroll up to read earlier messages, the chat history kept snapping back to the bottom on every token, forcing them to "fight" the viewport.

**Root cause:** the auto-follow effect in `ChatHistory` ran `scrollToBottom` on every history change whenever a stream was active (`isAtBottom || isStreaming`). The only opt-out was `isUserScrolling`, which is set by the `onScroll` handler — but that handler is debounced 50ms and skips events flagged as programmatic. During streaming, the per-token programmatic scrolls consistently won that race: by the time the debounced handler ran, the viewport had already been yanked back to the bottom, so the user's upward scroll was never registered.

**Fix:**
- Auto-follow now only engages when the viewport is actually at the bottom (`!isUserScrolling && isAtBottom`) — streaming alone no longer forces it.
- Added `onWheel` / `onTouchStart` + `onTouchMove` handlers that set `isUserScrolling` immediately on upward movement. Unlike `scroll` events, wheel/touch events are never fired by programmatic `scrollTo()`, so this opt-out cannot be swallowed by the streaming race.
- Sending a new prompt still snaps to the bottom and re-enables auto-follow (detected via the appended `pending` assistant reply), preserving the previous on-send behavior that the removed `isStreaming` clause provided.

Resulting behavior:
1. At the bottom, the view follows the tail of the stream as before.
2. Scrolling up (wheel, touch, or scrollbar drag) immediately stops auto-follow; the user can read freely mid-stream.
3. Auto-follow resumes when the user scrolls back to the bottom, clicks the scroll-to-bottom arrow, or sends a new prompt.

### Developer Validations


shatfield4 (changes requested):
After testing this out, the behavior does not feel natural. When sending another chat after a historical chat, the view is not scrolling to the bottom of the screen but seems to choose a certain point somewhere in the middle of the conversation.

- What should happen here, is on message sent by user -> scroll to bottom and lock into auto scroll unless the user scrolls out of it.

Also something else I noticed is that the scroll to bottom button no longer works for snapping to the autoscroll.

- What should happen is that, if the user clicks the scroll to bottom button as the chats is streaming -> it should snap to the bottom and stay locked into autoscroll.

angelplusultra:
@shatfield4 I pushed a fix to address your behavior concerns https://github.com/Mintplex-Labs/anything-llm/pull/6015/commits/acaf320157938c1f21ab56b0ab6ce025150e4459. Let me know if its working for you. Its pretty ugly though. It makes me wonder if this auto scroll feature is really worth it, do people really want this? Seems like people just complain about it and want autoscroll off and would prefer to just read at their own pace.

shatfield4 (changes requested):
So this functions just as I was imagining but the code is very hard to follow.

I think we should remove the autoscroll since this is a big complaint we have with many users.

What we can do is just scroll into view a new message being sent but get rid of autoscroll completely. This will allow users to keep reading the message as it streams from top to bottom since autoscroll makes it almost impossible to read as streaming is happening.

Removing the autoscroll should prevent this hard to read code with all these refs and conditional scrolling logic.

(I have tried many times to get this exact feature to work but could never keep the code clean)

@timothycarambat thoughts on this?

angelplusultra:
> So this functions just as I was imagining but the code is very hard to follow.
> 
> I think we should remove the autoscroll since this is a big complaint we have with many users.
> 
> What we can do is just scroll into view a new message being sent but get rid of autoscroll completely. This will allow users to keep reading the message as it streams from top to bottom since autoscroll makes it almost impossible to read as streaming is happening.
> 
> Removing the autoscroll should prevent this hard to read code with all these refs and conditional scrolling logic.
> 
> (I have tried many times to get this exact feature to work but could never keep the code clean)
> 
> @timothycarambat thoughts on this?

Definitely agree. I think this feature could be removed entirely, with the chat only scrolling to the bottom in two cases:

- When the user sends a message
- When the user clicks the down-arrow button (or via keybinding CMD + ArrowDown)

During an active stream, users should remain in control of scrolling so they can read at their own pace.

timothycarambat:
@angelplusultra @shatfield4 I disagree, we should meet the same UX as other providers. This isnt an impossible problem it just might be a small refactor and change we need to make to have this work. If we do abandon this UX totally where the user is in control of scroll we will be back where we started before this change of following ever came in and get issues about having the scroll work the way people want it to.

This is not a hard problem, just an annoying one - this change seems to work: https://github.com/Mintplex-Labs/anything-llm/pull/6046/changes

Its a bit involved, but that was mostly because of historical code.

PR #6020, opened 2026-07-18
ci: publish source archive checksums

### LINKED ISSUE #4743: [DOCS]: Checksums on Release
### Description

Ref: https://github.com/Mintplex-Labs/anything-llm/issues/4734

> Additionally, the GitHub release zips/tars do not include checksums or signatures. This makes it difficult for users to verify authenticity.

Publish a table of checksums for sources on release

zcxGGmu:
## Summary

- publish SHA-256 values for GitHub-generated zip and tar.gz source archives when a release is published
- stream archive hashing so large repositories do not require buffering an entire archive in memory
- preserve existing release notes and idempotently replace only the marked checksum section
- run focused Node tests on pull requests before enabling the release update job

Fixes #4743

## Testing

- `node --test .github/scripts/release-source-checksums.test.mjs`
- `node --check .github/scripts/release-source-checksums.mjs`
- `node --check .github/scripts/release-source-checksums.test.mjs`
- `npx --yes prettier@3.0.3 --check .github/scripts/release-source-checksums.mjs .github/scripts/release-source-checksums.test.mjs .github/workflows/publish-source-checksums.yaml`
- `actionlint .github/workflows/publish-source-checksums.yaml` (v1.7.12)
- streamed and hashed both real `v1.15.0` source archives: zip (`9,866,074` bytes, `6722e1f0...0ca10`) and tar.gz (`8,839,053` bytes, `519fc8ad...57f24`)
- repeated the complete zip download and confirmed identical size and SHA-256

## Caveats

- The live release PATCH was not exercised locally because that would modify an existing upstream release. The complete download/read/update request sequence is covered with a fetch mock.

### DIFF
--- a/.github/scripts/release-source-checksums.mjs
+++ b/.github/scripts/release-source-checksums.mjs
@@ -1,0 +1,138 @@
+import { createHash } from "node:crypto";
+import { pathToFileURL } from "node:url";
+
+export const CHECKSUMS_START = "<!-- source-checksums:start -->";
+export const CHECKSUMS_END = "<!-- source-checksums:end -->";
+
+export function sha256(content) {
+  return createHash("sha256").update(content).digest("hex");
+}
+
+export async function sha256Stream(stream) {
+  const hash = createHash("sha256");
+  for await (const chunk of stream) hash.update(chunk);
+  return hash.digest("hex");
+}
+
+export function sourceArchiveUrls(repository, tag) {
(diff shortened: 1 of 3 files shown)


timothycarambat:
This does not so what the core ask is - which is checksums for the desktop binaries, which we do not auto-attach on releases. Then the workflows run on every PR and not just on release.

Issue #5887, opened 2026-06-21
[FEAT]: Change WorkSpaces to be like a file manager


EDavidzar:
### What would you like to see?

Modify the workspaces so that they are accessible like a file manager, allowing users to sort files by name,date, size, file type/extension, delete them, search documents, search for duplicates, and also receive a warning when duplicates are added to the workspace.

Issue #5000, opened 2026-02-13
[FEAT]: Support GPT4All-style LocalDocs RAG with one-click indexing of multiple folders and all nested directories.


TheAlex25:
## Feature Request: LocalDocs-Style RAG with One-Click Multi-Folder Indexing in AnythingLLM

**Title:**
Support GPT4All-style LocalDocs RAG with one-click indexing of multiple folders and all nested directories.

**Summary:**
Add a LocalDocs-equivalent retrieval-augmented generation (RAG) feature to AnythingLLM that allows users to select one or more folders — including nested subfolders — from the operating system file explorer (Windows Explorer, macOS Finder, Ubuntu Files) and index them for semantic search and contextual retrieval in a single action. This would replace the current manual and clunky ingestion process.

**Current Limitation:**
AnythingLLM’s existing document ingestion for RAG requires manual uploads and repeated steps to add large collections or complex directory structures. There is no easy way to select entire folder trees at once for indexing.

**Proposed Behavior:**

1. **One-Click Folder Selection:**
   Allow users to right-click one or multiple folders in the OS file explorer (Windows, macOS, Linux) and choose “Index with AnythingLLM LocalDocs” (or similar). This should recursively include all nested files and subfolders.

2. **Automatic Embedding and Vector Storage:**
   Automatically process supported file formats (PDF, DOCX, TXT, MD, etc.), chunk content, generate embeddings, and store them in a local vector database without extra manual steps.

3. **Integrated Per-Query Retrieval:**
   At runtime, automatically retrieve the most relevant document snippets from the indexed collection and include them in the prompt context.

4. **Source Transparency:**
   Show clear source citations (file path and excerpt) for information used in responses.

**Benefits:**

* Significantly simplifies ingestion of large document collections and complex directory structures.
* Reduces friction for users with existing folder hierarchies.
* Improves the RAG experience with transparent source attribution.

**Success Criteria:**

* Users can select entire folder trees with one action in file explorer.
* Indexing runs without repeated manual uploads.
* Per-query retrieval is automatic and contextually relevant.
* Responses include clear source references.

**GPT4All's RAG Interface:**

<img width="1281" height="1456" alt="Image" src="https://github.com/user-attachments/assets/5d4e4b46-36a3-44f9-a0d7-5aa29d1b7c14" />

TheAlex25:
[...]

Issue #4257, opened 2025-08-07
[FEAT]: Improvements on File Handling and deduplication


fscheps:
### What would you like to see?

Hi Team, I am new to all this. Yet, I thought it would be a good idea to share this with the project. I hope its correct to share it here and that I am saying something that makes sense. 

While I found incredible the work that has been put into Anything LLM to make it as simple as possible, maybe there is space for improvement on the way that files are fed to different workspaces. 

I somehow don't get very well the way that copies of the files that I want to "feed" a workspace are kept as a reference. This way I would know which files not to feed again for example. Ideally I think it would be cool if we can have Anything LLM monitor a folder structure where we can place items and an engine would look at all the files. If files are removed, then this information is removed from the vector DB. If files are added then these are auto added. This could be scheduled at a certain time or triggered manually by a user. Ideally we should be able to keep our Workspaces folder structures of files in a way that we can keep some sort of organization. Also the possibility for the engine to detect exact duplicates even if the file has a different name might optimize things quite a bit. 

I hope what I am sharing makes sense and helps improve the project. 

Thank you for everything you do!

Fernando

timothycarambat:
Does this not work for that use case? https://docs.anythingllm.com/beta-preview/active-features/live-document-sync

For desktop specifically the automatic detection/embed of files is something we want to do and is more inline with the request, but there is no issue for it currently so this can suffice.

fscheps:
Hi there, thank you for your feedback, this is the key caveheat "_You currently cannot watch an entire directory._" :/ which kind of deceives the purpose to use it as a solution to monitor several files. Hope this could be enhanced in the neat future. Thanks a lot!!!

kaidoloor:
[...]

Issue #3561, opened 2025-03-28
[FEAT]: Auto-Add ALL Docs in a Given Folder


Tylast:
### What would you like to see?

I would like to have an option that allows AnythingLLM to auto-add all documents from a specified folder and continuously watch the folder for new documents to be added automatically. This would work very similarly to "Automatic document sync" but for the whole designated folder. BTW, gpt4all already has this & it works great!

The reason I'm asking for this is that I'm creating new work almost daily & I want my LLM to be up-to-date & responsive to my current corpus of work.

I see that #1873 was closed 9 months ago, with others asking for the exact same thing. PLEASE reconsider. :|

MrMarans:
If you are using the docker version, then you can use this: https://github.com/MrMarans/AnythingLLM_File-Manager
Thats what i coded myself. Its tested for windows file paths. It uploads the files to anythingLLM by its own, detects changes, creates/deletes workspaces based on folders and such. See the readme.
In a few small updates im going to ask for a mention in anythingLLMs "External Apps & Integrations" part of the readme.

owlowluio:
I may have found a solution to your problem. You can look here [click](https://javierlopez.eu/captha/) Hope this helped

winsontsg:
I am currently using the Docker version of AnythingLLM and attempting to integrate it with AnythingLLM_File-Manager, but I've encountered the issue below

<img width="556" alt="Image" src="https://github.com/user-attachments/assets/d21dce1f-0618-452d-b4ac-050bc2a4549e" />


> If you are using the docker version, then you can use this: https://github.com/MrMarans/AnythingLLM_File-Manager Thats what i coded myself. Its tested for windows file paths. It uploads the files to anythingLLM by its own, detects changes, creates/deletes workspaces based on folders and such. See the readme. In a few small updates im going to ask for a mention in anythingLLMs "External Apps & Integrations" part of the readme.

MrMarans:
First off, thanks for using that. I would love to help you, but
1. I need a bit more context. It seems you wanted to use the REST API to request an update? I would need to have the full address you wanted to call (you can hide the IP/Port if you want). Please be a bit more verbose for me to investigate your issue. Usually its the /update_files/update/ endpoint you need to call. 
2. Please create a Issue in my Repository. This thread is an Issue for anythingLLM not having such a feature by its own. Your problem is an issue for my app. I think the people of mintplex would be happy to have their issue threads be more AnythingLLM itself and not external Apps.

[...]

PR #6027, opened 2026-07-20
feat: folder upload with relative paths preserved in document manager

### LINKED ISSUE #2235: [BUG]: How to upload a folder with subfolders with files to AnythingLLM?

AnythingLLM desktop app

### What happened?

Hi, I want to prepare all the folders with its files for another person on his AnythingLLM. I'm trying to find a folder where to upload this on my computer as it seems the AnythingLLM just doesn't allow to upload folders just file by file what is tedious and not very useful if trying to use a RAG in its real power. So I would appreciate if somebody knows where (PATH) to go to upload a folder with all the files for the workspace. And also not clear what type of files it accepts, the info modal where to updates it says "and more!", well what does this even mean? I try to upload a small .rtf file and it holds forever without success. 

BTS, docs doesn't explain much.

Appreciate your time.

_No response_

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #2235

### Description

Adds folder uploading to the workspace document manager. Users can now drag and drop a folder (with nested subfolders) or use the new "or select a folder" picker, and the folder shows up in the document manager as its own folder with each file's relative path preserved.

How it works:

- A custom dropzone file aggregator (`frontend/src/utils/folderUpload.js`) walks dropped directories with `webkitGetAsEntry()` and records each file's path relative to the drop. It intentionally never reads `file.path`, since Electron fills that with the file's absolute OS path,  this keeps the feature working in the desktop app.
- Folder uploads send a `folderName` and metadata (title/docSource set to the file's relative path) with each file. Upload names are flattened (`sub/file.md` → `sub_file.md`) so same-named files from different subfolders can't clobber each other in the hotdir.
- On the server, the folder-move logic that already existed in the `/v1/document/upload/:folderName` API endpoint was extracted into a shared `moveProcessedDocsToFolder()` helper, now used by both that endpoint and the workspace upload endpoint.
- Hidden files and folders (`.DS_Store`, `.git`, etc.) are skipped during folder uploads.

Includes tests for `moveProcessedDocsToFolder`, covering folder creation, merging into existing folders, and path traversal attempts.

### Developer Validations


frontend/src/utils/folderUpload.js
```diff
@@ -0,0 +1,188 @@
...
+ *   FileSystemEntry tree and stamps each file's path relative to the drop
+ *   (eg: "/rootFolder/sub/file.md") onto RELATIVE_PATH_KEY.
+ * - Folder picker: webkitdirectory inputs populate file.webkitRelativePath
+ *   (eg: "rootFolder/sub/file.md").
+ *
+ * We deliberately never read file.path. react-dropzone's default aggregator
+ * carries the drop-relative path there, but Electron (<32) pre-fills that
+ * same property with the file's absolute OS path (eg: "/Users/me/notes/a.md"),
+ * which cannot be grouped into folders.
+ */
```
shatfield4, line 16: This makes sense here, I took a stab at getting this to work in desktop and it seems like this workaround is the only way to get it working. Looks good 👍

frontend/src/utils/folderUpload.js
```diff
@@ -0,0 +1,188 @@
...
+ * @param {File} file
+ * @param {string} fullPath
+ * @returns {File}
+ */
+function stampRelativePath(file, fullPath) {
+  Object.defineProperty(file, RELATIVE_PATH_KEY, {
+    value: fullPath,
+    writable: false,
+  });
+  return file;
```
shatfield4: Could we simplify this by doing this instead?

```
return file [RELATIVE_PATH] = fullPath;
```

It took me a second of rereading this a few times to understand what was going on and I think this is clearer.

frontend/src/utils/folderUpload.js
```diff
@@ -0,0 +1,188 @@
...
+  const fileItems = items.filter((item) => item.kind === "file");
+  const entries = fileItems.map((item) => item.webkitGetAsEntry?.() || null);
+
+  if (entries.length === 0 || entries.some((entry) => entry === null))
+    return fromEvent(event);
+
+  const collected = entries.map((entry, index) => {
+    // Loose files dropped outside a folder keep their native File object so
+    // nonstandard platform properties survive — the desktop app reads
+    // Electron's file.path off uploads for its local file sync feature.
```
shatfield4, line 114: 👍

server/endpoints/api/document/index.js
```diff
@@ -305,29 +296,7 @@ function apiDocumentEndpoints(app) {
...
-              !isWithin(documentsPath, destinationPath)
-            )
-              throw new Error("Invalid file location");
-
-            fs.renameSync(sourcePath, destinationPath);
-            doc.location = path.join(folder, path.basename(doc.location));
-            doc.name = path.basename(doc.location);
-          }
-        }
+        const folder = moveProcessedDocsToFolder(documents, folderName);
```
shatfield4, line 299: 👍

server/utils/files/index.js
```diff
@@ -379,6 +379,53 @@ async function getWatchedDocumentFilenames(filenames = []) {
...
+ * processed collector documents into it, updating each document's `location`
+ * and `name` in-place. If the folder already exists, documents are merged
+ * into it so repeated uploads to the same folder are idempotent. Folder name
+ * handling matches the historical /v1/document/upload/:folderName behavior:
+ * normalizePath only, so existing API consumers keep their exact folders.
+ * @param {Array<{location: string, name: string}>} documents - documents returned by Collector.processDocument
+ * @param {string} folderName - target folder name (e.g. "my-notes")
+ * @param {string} basePath - base documents directory (overridable for testing)
+ * @returns {string} the normalized folder name the documents were moved into
+ */
```
shatfield4, line 700: Great idea moving this to a helper with tests. Helps clean up the endpoint it originally lived in.

shatfield4 (changes requested):
Great job! One small nit here otherwise, LGTM.

I ported this over to desktop and made sure the drag and drop was working fully in both docker and desktop versions. It does suck a little bit that we have to do this workaround to get the full path of files when a folder is dropped but I could not find another way to do this.

frontend/src/utils/folderUpload.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6027/commits/fcf78c52adf34cb229346b56569ed8b2d8704cab

shatfield4 (approved):
Nice, LGTM

timothycarambat:
Did a massive refactor while I was in there because some stuff was driving me insane.

Add folder upload support with a lazy-loaded document picker

Dragging a folder onto the uploader now creates a matching top-level
folder in the document store instead of dumping every file into
custom-documents. Files in subfolders are skipped for now — nested
folder support is deferred to a **future PR.**

Why: the picker previously loaded **every** document on disk up front,
which does not survive a large document store. It now loads folder
names and counts first and fetches a folder's files only when you
open it. (it will also eager load on hover of parent folder)

Uploading
- Drag-and-drop a folder to upload it (the separate "select a folder"
  button was removed — drag-and-drop covers it)
- Uploads no longer blank out the picker. The list stays on screen,
  hydrates in place when the upload finishes, and pre-selects the
  newly added files so you can embed them right away
- A folder drop of 50 files triggers one refresh, not 50
- Drop files directly onto a folder row in the picker to upload straight
  into that folder, instead of uploading and then moving. Rows highlight
  on drag-over and share the main uploader's progress list
- Dropping a folder onto a folder row is rejected rather than flattened,
  since storage is exactly folder/file.json and flattening would produce
  documents literally named "subfolder/report.pdf". A toast points you
  at the uploader, where a dropped folder becomes a folder of its own
- The link input no longer requires you to type a protocol. "example.com"
  and "example.com/docs" are accepted and sent as https

Document picker
- Folders show a name and file count immediately; files load on expand
- Hovering a folder prefetches it, so expanding usually feels instant
- Large folders page 100 at a time with a "Load more" row
- Clicking a folder row opens it. Selecting is the checkbox's job — a
  whole-row click was too easy to hit by accident for something that
  can stage hundreds of files for embedding
- Folder checkboxes now show a partial state, and selecting a folder
  also covers the files inside it that have not been fetched yet.
  Clicking a checked box clears the folder and clicking a partial one
  fills it in, whether the folder was selected as a whole or by way of
  every file in it being selected individually
- An empty folder can be selected, so it can be deleted
- Search runs on the server via ripgrep (filename and file contents)
  instead of filtering an in-memory copy of the whole tree
- While searching, the header counts matches rather than reporting the
  library-wide total next to a filtered list
- Deletes, moves and embeds show a small status strip instead of
  replacing the whole list with a spinner
- A selected folder's file count is readable in light mode against the
  selected-row background

Staying in sync
- Removing a document from a workspace puts it straight back in the
  picker. Fetched pages have embedded files filtered out of them, so
  they are now invalidated when a folder's embedded set changes and
  not just when its file count does — previously the only way to see
  an un-embedded file again was to leave the tab and come back
- Folder counts reflect what is actually listed. A folder with
  everything embedded reads as empty instead of still claiming its
  on-disk count

Picker internals
- All picker state (folder shells, fetched pages, expansion, selection,
  search) moved into a single useDocumentPicker reducer. It previously
  lived in three places that could disagree — parent state, a non-
  reactive ref, and per-row local state — which is what caused
  checkboxes not to highlight and rows to go stale
- FolderRow is now presentational and FileRow is memoized, so toggling
  a checkbox re-renders one row instead of the tree
- The upload queue moved into a useUploadQueue hook, so the main
  dropzone and every folder drop target feed one queue and one progress
  list

Server
- listFolders() returns folder shells (name + file count) without
  parsing any documents
- getDocumentsByFolder() takes offset/limit, parses only the requested
  page, and reads through the shared fileToPickerData helper so corrupt
  JSON and oversized documents are handled instead of throwing
- limit is clamped to 1000; limit=all opts out of paging entirely
- searchDocuments() shells out to ripgrep asynchronously and reads only
  the files ripgrep returns
- getDocumentsByDocPaths() resolves picker metadata for a workspace's
  embedded documents without walking the documents directory, and
  returns each document's docpath so callers can attribute it to the
  right folder
- moveProcessedDocsToFolder() is now shared by the GUI and API upload
  paths. Folder names must be a single path segment: document storage
  is exactly folder/file.json, and a nested name would create documents
  the picker cannot see and that cannot be embedded

Endpoints
- GET /system/local-files supports ?folder=&offset=&limit=
- GET /system/local-files/search?q= (new)
- POST /system/local-files/by-docpaths (new)
- GET /v1/documents gains an optional ?folder=&offset=&limit=
- GET /v1/documents/folder/:folderName gains optional offset/limit

Backward compatibility
- GET /v1/documents with no query params returns the full document tree
  exactly as before. It is slow on large instances, but returning empty
  items arrays to existing integrations would be worse
- GET /v1/documents/folder/:folderName returns every document by
  default rather than a first page
- POST /v1/document/upload/:folderName now rejects a URL-encoded path
  separator in the folder name. This is the one intentional behavior
  change; see moveProcessedDocsToFolder above

Tests
- Added coverage for moveProcessedDocsToFolder: folder creation and
  merging, in-place document updates, path traversal, and nested-name
  rejection

Resolves #2235
resolves #3135
resolves #5887
resolves #5000
resolves #4257
resolves #3561

PR #6029, opened 2026-07-21
fix: keep watched document re-syncs from breaking pinned document de-duplication

### LINKED ISSUE #5334: [BUG]: Pinned and Watched Documents without RAG

AnythingLLM desktop app

### What happened?

I would like to always add the entire, current, content of a file as context to my message, without embedding it through RAG. 

The content of the file changes, so I can't use any of the static ways like System Prompts.

Currently the watch feature doesn't update when a new thread is created or a new message sent.
I found out that toggling the watch icon, updates the status under `Watched documents` but doesn't actually update the document.  The documentation says that pinned documents are excluded from RAG, but have found that sometimes two versions of the document are added as context. Since pinned documents are excluded from RAG, therefore when watched **there is no reason to not always read the latest version of a file from disc.**

I suspect the various optimizations for pinned and for watched documents are not playing nicely together, hampering this feature.

Other options I considered that didn't work:
I was hoping to use system prompt variables, but those can't be dynamic (which would be a great feature on its own)

I also do not want to make two API calls each time in agent mode to have it read the file, doubling the cost.

Pin and Watch a text file.

ArslanRasheed60:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5334

### Description

**Reproduction**

1. Add a link document to a workspace and turn on "watch" for it.
2. Pin the same document in that workspace.
3. Let the watched-document sync job run once with changed source content.
4. Chat with the workspace: the document now comes back both as pinned full text and as
   RAG chunks, so the same content is injected into the context twice.

**Cause**

Pinned documents are de-duplicated out of similarity search results with
`sourceIdentifier()` (`server/utils/chats/index.js`), which keys on
`title` + `published`. The pinned side of that comparison is read from the document JSON on
disk; the RAG side is the metadata stored on each chunk.

`server/jobs/sync-watched-documents.js` wrote two different values for `published` during a
single sync: the payload handed to `addDocumentToNamespace` carried the **old** timestamp
from `currentDocumentData`, while `updateSourceDocument` immediately wrote a **new** one to
disk. From the first re-sync onwards the two identifiers can never match, the pin filter
silently stops filtering, and the document is injected twice.

**Fix**

Compute the timestamp once and stamp the workspace vectors, the source document and every
other workspace that references the same file with that same value. The primary
`addDocumentToNamespace` call runs with `skipCache: true` and is what writes the vector
cache file, so the bloomed workspaces (which read that cache) inherit the corrected
timestamp as well.

**Scope**

This is limited to the double-injection / de-duplication bug. The issue also raises a
secondary feature request — that a document which is both watched and pinned should always
have its latest version read from disk at chat time — which is deliberately **not**
included here, as that is a product design decision for the maintainers.

### Visuals (if applicable)

N/A — server-side only.

### Additional Information

Added `server/__tests__/jobs/sync-watched-documents.test.js`, which drives the sync job with
mocked collectors/models/vector DB and asserts that the vector payload, the on-disk payload
and the bloomed workspace payload all resolve to the same `sourceIdentifier()`. It fails on
current `master` (vector payload keeps the old timestamp, disk payload gets a new one) and
passes with this change.

### Developer Validations


angelplusultra:
LGTM

PR #6030, opened 2026-07-21
chore: improve agent tool routing between filesystem tools and rag-memory

### LINKED ISSUE #5968: [CHORE]: Update default system prompt for better tool use.

Docker (local)

## Title
  Agent chat never spontaneously selects `rag-memory` tool despite explicit "use workspace documents"
  instructions (generic-openai provider, tooled mode)

  ## Environment
  - AnythingLLM v1.15.0 (Docker, `mintplexlabs/anythingllm:latest`)
  - LLM provider: `generic-openai`, model `gemma-4-26B-A4B-it-QAT-MLX-4bit` (quantized, served via a local
  OpenAI-compatible endpoint)
  - Native tool calling confirmed active (`Provider.stream (tooled)` in logs — not the `UnTooled` prompted-JSON
  fallback)
  - `rag-memory` confirmed present in the agent's available tool set every time (selected by the tool reranker,
  logged explicitly each run)
  - oMLX
  - MacBook Pro M4 Max 64GB, macOS 26.5.1

  ## Summary
  Across 7 separate agent chat invocations in the same workspace — including prompts that explicitly said "using
  the information stored in this workspace" — the agent never once spontaneously called `rag-memory`, even when
  it was the only tool that could actually answer the request (a document that existed solely as a RAG-embedded
  document, not on the container's local filesystem). Instead it consistently reached for `filesystem-*` tools,
  which structurally cannot see RAG-embedded content, producing false "file not found" results and, in one
  observed case, misattributing content pulled from a different embedded document via `document-summarizer`
  instead.

## Steps to reproduce
  1. Embed a document into a workspace (confirmed via the `document_vectors` table — a real vector row exists).
  2. In agent chat, ask the model to review/analyze that document using workspace context (no need to reference
  any tool by name).
  3. Observe: the agent tries `filesystem-search-files` / `filesystem-read-text-file` (which only see a local
  sandbox directory, unrelated to RAG storage), not `rag-memory`.

  ## Evidence the tool itself is not broken
  - Called the same underlying `vectorDB.performSimilaritySearch` function directly, bypassing the LLM entirely —
  it returned correct, well-attributed, reranked results in under 300ms.
  - When the agent was explicitly told, by tool name, to call `rag-memory` with a specific query, it did so
  successfully and used the results correctly — including correctly attributing a specific claim to the actual
  source document rather than conflating it with the document under review.

  ## Expected behavior
  The agent should reliably choose `rag-memory` when a request references "this workspace's documents," or when a named file can't be located via filesystem tools, rather than defaulting to filesystem tools that have no
  visibility into RAG-embedded content at all.

  ## Secondary observation
  `PROVIDER_SUPPORTS_NATIVE_TOOL_CALLING` appears to have zero references anywhere in the current server source
  (checked via grep across `/app/server`). Setting it has no effect — `GenericOpenAiProvider` always resolves its
  tool-calling mode independently of this variable. Worth either honoring it or removing it from documentation
  to avoid confusion.

oMLX / gemma-4-26B-A4B-it-QAT-MLX-4bit (LMStudio)

Ollama / nomic-embed-text:latest

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)
- [x] 🔨 chore (Build, CI, maintenance)

### Relevant Issues

resolves #5968

### Description

Agents with both the filesystem and rag-memory skills enabled would often reach for filesystem tools when the user referenced embedded workspace documents, come up empty, and give up — embedded documents live in the vector database, not on disk.

This PR improves tool routing between the filesystem tools and the document tools (`rag-memory`, `document-summarizer`):

- Appends routing guidance to the default agent system prompt (only naming the document tools that are actually enabled) explaining that embedded documents are not on the filesystem.
- Rewrites the `rag-memory` tool description to clarify it searches embedded documents in chunks, and points to `document-summarizer` for full-document summaries.
- Updates `filesystem-search-files`, `filesystem-read-text-file`, `filesystem-read-multiple-files`, and `filesystem-list-directory` descriptions to redirect the agent to the document tools when a referenced document is not found on disk.
- Clarifies the `document-summarizer` description so it is clearly scoped to embedded workspace documents.

Prompt/description changes only — no behavioral logic changed.

### Visuals (if applicable)

N/A

### Additional Information

Added tests covering the conditional system prompt guidance: present when the document skills are enabled, partial when only one is enabled, and absent when both are disabled via `disabled_agent_skills`.

### Developer Validations

### DIFF
--- a/server/utils/agents/aibitat/plugins/memory.js
+++ b/server/utils/agents/aibitat/plugins/memory.js
@@ -55,6 +55,21 @@ const memory = {
                 content: "I am a robot, the user told me that i am.",
               }),
             },
+            {
+              prompt:
+                "Save this for later: my preferred deploy target is staging.",
+              call: JSON.stringify({
+                action: "store",
+                content: "The user's preferred deploy target is staging.",
+              }),
+            },
+            {
+              prompt: "What did I tell you about my deploy preferences?",
+              call: JSON.stringify({
+                action: "search",
+                content: "user's deploy target preferences",
+              }),
+            },
           ],
           parameters: {
             $schema: "http://json-schema.org/draft-07/schema#",
--- a/server/utils/agents/aibitat/plugins/memory.js
+++ b/server/utils/agents/aibitat/plugins/memory.js
@@ -19,13 +19,26 @@ const memory = {
           tracker: new Deduplicator(),
           name: this.name,
           description:
-            "Search your local documents and workspace files for relevant information, or store information to long-term memory. Use search to find answers in uploaded documents, embedded files, or previously stored memories. Use store only when explicitly asked to remember or save something.",
+            "Search the documents embedded in this workspace, or store information to long-term memory. " +
+            "Embedded workspace documents do NOT exist on the filesystem and cannot be found by any filesystem tool — " +
+            "this is the ONLY tool that can read them. " +
+            "Use search whenever the user references workspace documents, uploaded or embedded files, or previously stored memories, " +
+            "or when a document they mention cannot be found on the local filesystem. " +
+            "Use store only when explicitly asked to remember or save something.",
           examples: [
             {
               prompt: "Check my files for information about the project",
               call: JSON.stringify({
                 action: "search",
                 content: "<project information to search for>",
+              }),
+            },
+            {
+              prompt:
+                "Using the information stored in this workspace, summarize the quarterly report",
+              call: JSON.stringify({
+                action: "search",
+                content: "quarterly report",
               }),
             },
             {
--- a/server/utils/agents/defaults.js
+++ b/server/utils/agents/defaults.js
@@ -84,16 +84,23 @@ const WORKSPACE_AGENT = {
       role +=
         "\n\nWhen you need information from the user (URLs, file paths, preferences, choices, etc.), you MUST use the request-user-input tool. Do not ask questions in your text response - the user cannot reply to text. Only the tool can collect user input.";
 
-    return {
-      role,
-      functions: [
-        ...(await agentSkillsFromSystemSettings()),
-        ...clarifyingQuestionsSkills,
-        ...ImportedPlugin.activeImportedPlugins(),
-        ...AgentFlows.activeFlowPlugins(),
-        ...(await new MCPCompatibilityLayer().activeMCPServers()),
-      ],
-    };
+    const functions = [
+      ...(await agentSkillsFromSystemSettings()),
+      ...clarifyingQuestionsSkills,
+      ...ImportedPlugin.activeImportedPlugins(),
+      ...AgentFlows.activeFlowPlugins(),
+      ...(await new MCPCompatibilityLayer().activeMCPServers()),
+    ];
+
+    // When rag-memory is enabled, add explicit tool-routing guidance so the
+    // agent does not reach for filesystem tools when the user references
+    // workspace documents - embedded documents are not on the filesystem and
+    // are only reachable via the rag-memory tool. See issue #5968.
+    if (functions.includes(AgentPlugins.memory.name))
+      role +=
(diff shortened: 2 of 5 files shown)


server/utils/agents/defaults.js
```diff
@@ -84,16 +84,23 @@ const WORKSPACE_AGENT = {
...
+      ...ImportedPlugin.activeImportedPlugins(),
+      ...AgentFlows.activeFlowPlugins(),
+      ...(await new MCPCompatibilityLayer().activeMCPServers()),
+    ];
+
+    // When rag-memory is enabled, add explicit tool-routing guidance so the
+    // agent does not reach for filesystem tools when the user references
+    // workspace documents - embedded documents are not on the filesystem and
+    // are only reachable via the rag-memory tool. See issue #5968.
+    if (functions.includes(AgentPlugins.memory.name))
```
shatfield4: Can we add this here?

```js
    const hasFilesystemTools = functions.some(
      (f) => typeof f === "string" && f.startsWith("filesystem-")
    );
    if (hasFilesystemTools && functions.includes(AgentPlugins.memory.name))
      role +=
        "\n\nDocuments uploaded or embedded into this workspace live in a vector database - NOT on the filesystem. When the user references workspace documents or uploaded files, or a document they mention cannot be found on the local filesystem, use the rag-memory tool to search for it before trying any other tool.";
```

Reasoning behind this is that this adds to the prompt whenever rag-memory is enabled which is enabled by default. What we should do is check to make sure we only inject the prompt when filesystem tools are enabled (they are opt-in by default).

server/utils/agents/aibitat/plugins/memory.js
```diff
@@ -19,7 +19,12 @@ const memory = {
...
           description:
-            "Search your local documents and workspace files for relevant information, or store information to long-term memory. Use search to find answers in uploaded documents, embedded files, or previously stored memories. Use store only when explicitly asked to remember or save something.",
+            "Search the documents embedded in this workspace, or store information to long-term memory. " +
+            "Embedded workspace documents do NOT exist on the filesystem and cannot be found by any filesystem tool — " +
+            "this is the ONLY tool that can read them. " +
+            "Use search whenever the user references workspace documents, uploaded or embedded files, or previously stored memories, " +
+            "or when a document they mention cannot be found on the local filesystem. " +
+            "Use store only when explicitly asked to remember or save something.",
           examples: [
             {
```
shatfield4, line 29: Since we're trying to improve performance of the memory tool, could we add some examples of storing vs searching in the examples array?

```js
            {
              prompt: "Save this for later: my preferred deploy target is staging.",
              call: JSON.stringify({
                action: "store",
                content: "The user's preferred deploy target is staging.",
              }),
            },
```

shatfield4 (changes requested):
Looks pretty good here, just a few changes I noted that are worth investigating.

server/utils/agents/defaults.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6030/commits/938b82f7abf3c54b913ac90cb3e65367a3014835

server/utils/agents/aibitat/plugins/memory.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6030/commits/83a45946d6e0bf1cf6e55c4ba802ac24ff4ff2d1

shatfield4 (approved):
LGTM

timothycarambat:
CLosing this PR but going to keep the main issue open - I want to actually kill and wrap these skills together because yeah it is annoying the skills are conflated but have a huge amount of overlap. RAG-memory should also be able to search everything - not just uploaded docs.

Lets revisit this on internal board because this needs to be fixed as a full rework

PR #6038, opened 2026-07-23
Fix slash command presets not expanding into prompt text

### LINKED ISSUE #6036: [BUG]: slash-commands not automatically entering prompt text

Docker (local)

### What happened?

I've created quite a few slash-commands (11 total) that I use with models for various things, like weather, listing tools and benchmarking various facets of the model.  I have had the issue described below with various models.  

When I enter the slash command, the prompt that should be relayed to the model appears to not get relayed correctly.  What I find if I leave the conversation, when the slash-command has worked it will show the text that should be relayed as an entry from me, the full text.  When it has not worked it will just show the slash command.   The model will start asking questions oftentimes as well.  For instance I have one for /weather, which tells the model to find the location automatically, and the model will respond with a few different responses like: an incorrect timeframe for the weather, ask me what I would like to know about the weather,  or did I miss something about the weather, why did you just enter /weather.  

Is there a limit to slash-commands that we can have, or is it suggested that we not have more than so many active?   

This seems to happen randomly, of course it will happen deeper into the conversation.  Usually I start my conversations with a simple "hello" and then immediately after that it will often happen if I enter a slash-command.  

I've been using various Qwen models as of late from 9b to 35b, mostly, and not the standard Qwen models.  My backend is LM Studio.  One time a model did complain that intelligent tool use was an issue, when I was reproducing this, so I turned that off.  I've also restarted my docker container.  

Default embedder w/ mini selected.  Usually I have similarity set to Low.

shatfield4:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6036 

### Description

- Some users noticed their slash command (like `/weather`) would show up as-is in the chat instead of being replaced with its prompt
- This happened when the command wasn't at the very start of the message, or had punctuation around it (`ok, /weather`, `/weather?`)
- Fixed the matching so command expands in all cases, while still not matching inside a longer word (`/weather` in `/weatherman`)

### Developer Validations

### DIFF
--- a/server/__tests__/utils/chats/index.test.js
+++ b/server/__tests__/utils/chats/index.test.js
@@ -1,0 +1,98 @@
+/* eslint-env jest, node */
+const { grepCommand, grepAllSlashCommands } = require("../../../utils/chats");
+const { SlashCommandPresets } = require("../../../models/slashCommandsPresets");
+
+jest.mock("../../../models/slashCommandsPresets");
+
+// Helper to shape preset rows the way the model returns them.
+const preset = (command, prompt) => ({ command, prompt });
+
+describe("grepCommand", () => {
+  beforeEach(() => jest.clearAllMocks());
+
+  it("returns the built-in command when the message starts with it", async () => {
+    SlashCommandPresets.getUserPresets.mockResolvedValue([]);
+    expect(await grepCommand("/reset")).toBe("/reset");
+    expect(await grepCommand("/RESET now")).toBe("/reset"); // case-insensitive
+  });
+
+  it("returns the message unchanged when no command matches", async () => {
+    SlashCommandPresets.getUserPresets.mockResolvedValue([]);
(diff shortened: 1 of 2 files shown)


CutterSol:
Thanks, it's about time! Haha.  

I'll check the next version release.  In the meantime, I've updated my custom-skill and hope to post it to the ALLM hub soon, so that you guys might have better diagnostic data.  I hope to expand it to universal v1 endpoints to try to get settings, models, etc on the endpoint.  Does anything like this already exist?  I plan on checking the hub tonight.  

Hatfield made rain for LA!

angelplusultra (changes requested):
Seems solid. Do you think we can add thorough test suites for both `grepCommand` and `grepAllSlashCommands`? I think something like this would definitely benefit from it.

angelplusultra (approved):
LGTM

PR #6047, opened 2026-07-24
Start model router cooldown after inference completes

### LINKED ISSUE #6042: [FEAT]: Change/add toggle to model router cooldown timer to trigger AFTER inference has stopped.

Currently the cooldown timer appears to be triggered at the point where ALLM detects a model routing rule trigger. With the default set at 300s, some reasoning models doing multiple rounds of tool calls and COT outputs will extend the final output time to over 300s. This means that the router will fall back to the default model unless the user remembers to deliver criteria to reenable it when following up on the initial request. 

Yes, the most immediate action would be to set a longer cooldown timer. But making the timer trigger after the signal that inference has stopped is more user friendly IMO. Perhaps a toggle could be added to change when the timer start is triggered.

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6042 

### Description

- model router cooldown timer now starts when the response finishes instead of when the routing rule is detected
- prevents long-running inference (reasoning models, multi tool-call responses) from consuming the entire cooldown window before another chat can be sent
- re-stamps the sticky route at inference end via the existing `setStickyRoute`, so the timer always restarts for a routed message (fallback routes are not made sticky)
- `resolveProviderConnector` now also returns the router instance which each chat entry point calls `router?.onInferenceComplete()` once the routed model finishes responding, which re-stamps the sticky route from a fresh timestamp
- applied across every entry point that routes chat, not just the workspace UI: workspace chat stream, developer API (sync + stream), OpenAI-compatible endpoint (sync + stream), embed widget, and telegram

### Developer Validations


angelplusultra (changes requested):
The core logic makes sense, but I think its currently missing agentic paths. Does the `onInferenceComplete` method need to fire once the agent loop exits with a final response?

shatfield4:
> The core logic makes sense, but I think its currently missing agentic paths. Does the `onInferenceComplete` method need to fire once the agent loop exits with a final response?

Fixed to add the re-stamp after agent completes in d031e4c82

server/utils/agents/index.js
```diff
@@ -918,13 +919,15 @@ class AgentHandler {
...
-  startAgentCluster() {
-    return this.aibitat.start({
+  async startAgentCluster() {
+    const result = await this.aibitat.start({
       from: USER_AGENT.name,
       to: this.channel ?? WORKSPACE_AGENT.name,
       content: this.#stripAgentCommand(this.invocation.prompt),
       attachments: this.attachments,
     });
+    this._modelRouter?.onInferenceComplete();
```
angelplusultra: This seems to only refresh the sticky route on the first turn of the agent loop. Every subsequent prompt through the socket doesnt refresh the route. To repro throw a log in `onInferenceComplete()` after the sticky route refreshes and youll notice it doesn't fire on follow up prompts after the initial one. This needs to be called on the `onInterrupt` event on the aibitat instance because that's when the agent exits and waits for the user prompt on the socket level.

e.g
in `createAibitat`
```js
this.aibitat.onInterrupt(() => this._modelRouter?.onInferenceComplete());
this.aibitat.onTerminate(() => this._modelRouter?.onInferenceComplete());
```

After applying this you can get rid of the code above because its redundant and only fires after this promise resolves

angelplusultra (changes requested):

server/utils/agents/index.js
shatfield4: Resolved by adding a console log to `onInferenceComplete` for testing and was able to see the log triggered on every new message sent after inference completes.

angelplusultra (approved):
Nice job, this is pretty solid now. I would say maybe if you could add some test cases for `onInferenceComplete` just to be thorough that would be great. Otherwise LGTM

angelplusultra (approved):
LGTM

PR #6055, opened 2026-07-27
Add input_audio support for audio attachments on generic OpenAI provider

### LINKED ISSUE #6025: [FEAT]: Add support for audio attachments (input_audio) for API
## The problem

Currently, the GenericOpenAi provider (and #generateContent helper) assumes that all API attachments are images. When sending audio files via the API, AnythingLLM automatically wraps them in an image_url block along with the full data:audio/...;base64, prefix.

When using OpenAI-compatible backends that support audio (such as llama.cpp or native multimodal audio endpoints), this causes two issues:

1. Backends reject the request with 500 Invalid uri format because audio/... URIs are invalid inside an image_url block.
2. OpenAI's standardized schema for multimodal audio expects type: "input_audio" with a raw base64 string (without the data:audio/...;base64, Data URI prefix) and a format key (e.g., mp3, wav).

## The solution

Update the #generateContent method in /server/utils/AiProviders/genericOpenAi/index.js to inspect the MIME type of incoming attachments.

If an attachment's MIME type starts with audio/:

1. It should format the content object as type: "input_audio".
2. It should extract the raw base64 string by stripping the Data URI prefix (data:audio/...;base64,).
3. It should include the audio format (e.g., format: "mp3").

Thank you very much for your work!

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6025
closes #6053

### Description

- detect `audio/*` attachments in the generic OpenAI provider and send them as `input_audio` content blocks (raw base64 + format) instead of `image_url`
- handle both chat and agentic paths
- keep images and all other attachments on the existing `image_url` path (backward compatible)
- normalize the audio format field (mpeg/mp3 → mp3, wav variants → wav, strip `x-` otherwise)
- add tests covering image, audio, mixed, and no-attachment cases

### Developer Validations


angelplusultra (changes requested):
Few findings:

1. I couldn't repro what this is supposed to fix on `master`. I tested Ollama and OpenRouter through Generic OpenAI provider and both through agent chat and regular chat. Both worked with the LLM having the audio in context and able to chat about it. The users claim about providers throwing a 500 error when chat message has content type of `image_url` when it's an audio file was not verifiable on my end. It seems to not really care what the content "type" is honestly, it seems to just takes the bytes and gives it to the LLM. But this could definitely be a provider to provider kinda thing. If you know which exact provider will throw when a content type is tagged incorrectly that would be great so I can see the error for myself.

2. If we're going to fix this where we properly tag audio attachment files as `input_audio`, then we should do it across all providers right? All providers `generateContent` method assumes the attachment passed in is an image type and tags it as that.

3. I don't believe the current fix tags audio attachments as `input_audio` on the agentic paths. I think `generateContent` is just used via regular chat. Look up the function `formatMessageWithAttachments`

shatfield4:
> Few findings:
> 
> 1. I couldn't repro what this is supposed to fix on `master`. I tested Ollama and OpenRouter through Generic OpenAI provider and both through agent chat and regular chat. Both worked with the LLM having the audio in context and able to chat about it. The users claim about providers throwing a 500 error when chat message has content type of `image_url` when it's an audio file was not verifiable on my end. It seems to not really care what the content "type" is honestly, it seems to just takes the bytes and gives it to the LLM. But this could definitely be a provider to provider kinda thing. If you know which exact provider will throw when a content type is tagged incorrectly that would be great so I can see the error for myself.
> 2. If we're going to fix this where we properly tag audio attachment files as `input_audio`, then we should do it across all providers right? All providers `generateContent` method assumes the attachment passed in is an image type and tags it as that.
> 3. I don't believe the current fix tags audio attachments as `input_audio` on the agentic paths. I think `generateContent` is just used via regular chat. Look up the function `formatMessageWithAttachments`

1. In order to repro this, I had to use llama.cpp with llama server and a model that supports audio (ggml-org/ultravox-v0_5-llama-3_2-1b-GGUF). Ollama and OpenRouter are much more lenient with how you provide attachements but with llama.cpp that is how I was able to repro the 500 error.
2. Fixing this on all providers would be outside the scope of this issue and much riskier modifying it across all providers so we are holding off on that in this PR.
3. Fixed and pushed agentic path fixes

angelplusultra (changes requested):
Nice, but this is only a partial fix. `tooled.js` has its own module level implementation of `formatMessageWithAttachments` which fires if the provider supports native tool calling and goes through tooled completion handlers, which generic openai does by default. 

The provider level `formatMessageWithAttachments` only fires if it goes through untooled. Look at `formatMessageWithAttachments` inside `tooled.js`.

shatfield4:
> Nice, but this is only a partial fix. `tooled.js` has its own module level implementation of `formatMessageWithAttachments` which fires if the provider supports native tool calling and goes through tooled completion handlers, which generic openai does by default.
> 
> The provider level `formatMessageWithAttachments` only fires if it goes through untooled. Look at `formatMessageWithAttachments` inside `tooled.js`.

Went ahead and modified the `formatMessageWithAttachments` inside `tooled.js` with the same logic and also pulled this logic to a util so we can reuse this everywhere.

Something to note though is when we modify the `tooled.js` this applies to every other provider that uses this too. The only path it touches though is if a user was intentionally sending audio as image url to their provider (highly unlikely) but if they are sending audio, now it is in the correct format.

PR #6063, opened 2026-07-28
feat: abort agent sessions end-to-end when the user stops generation

### LINKED ISSUE #5905: [BUG]: Inference does not stop when clicking away from active session

AnythingLLM desktop app

### What happened?

This had been reported before, but closed, however I could repro this issue on 1.14.1 just today using LM Studio as the model provider.

https://github.com/Mintplex-Labs/anything-llm/issues/2640

If you move from an active chat it closes the connection (this in and of itself is a dubious design decision without real benefit) however the inference does not stop on the other end and it keeps generating tokens into thin air. While this may be a nuisance on a local LLM provider it can be a costly API issue using remote frontier API tokens.

This is a very serious bug.

set up LM Studio and use it as the model provider in AnythingLLM. Ask an agent question then click away and back. Notice that the question as well as the inference progress is gone in the chat. Go to the server side of LM Studio and observe that there is a active token process still generating tokens.

LM Studio with qwen/qwen3.6-35b-a3b

_No response_

### LINKED ISSUE #5969: [BUG]: `<StopGenerationButton />` Disappears on The First Prompt That Initiates an Agent Session

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5969
resolves #5905
closes #6006 

### Description

Builds on the stop-button fix by making the stop actually abort the agent session server-side, end to end:

- `AIbitat` now owns a session-wide `AbortController` plus an `_aborted` flag. `abort()` cancels in-flight provider requests and the loop bails at every boundary (`chat()`, `handleExecution`, `#handleExecutionStream`), so a mid-stream abort never acts on a truncated tool call or starts another turn.
- The websocket endpoint aborts the running session on socket close (stop button, tab close, disconnect) — not just on the explicit bail command — and skips starting the agent cluster if the socket closed while aibitat was still being built.
- `Provider` base class gains `attachAbortSignal()`; aibitat attaches the session signal to every provider it instantiates.
- Local providers (ollama, lmstudio, localai, koboldcpp, textgenwebui, nvidiaNim, dockerModelRunner, foundry, omlx, lemonade) thread the signal into their SDK calls via the shared `tooled.js` helpers, so aborting tears down the in-flight LLM request. Ollama merges the session signal with the SDK's own signal in a custom fetch wrapper.

Cloud providers are not wired to the signal yet — for those, an abort still stops the loop at the next boundary; the in-flight request just isn't cancelled. They can be wired in a follow-up or in this PR if the core logic is approved.

### Developer Validations


shatfield4 (approved):
LGTM

I was able to test the 3 different paths here that were modified. These 3 providers cover all the paths of other providers that were modified as well.

I tested:
- LM Studio
- Foundry
- Ollama

All 3 of these providers worked perfectly fine for when I aborted a streaming request mid-generation.

While testing, I saw this part of the code `if (error instanceof OpenAI.APIUserAbortError) throw error`. Good job handling this, I can see how this would be tricky. If this was wrapped in the `RetryError`, it would tell the provider that it was a failed retry and the provider would not treat it as a stop by would retry the same request.

timothycarambat:
Took a pass over this and restructured the provider plumbing, then extended the same
mechanism to the paths that weren't covered - there were some serious coverage gaps. 

Summary of what changed below:

### 1. The provider plumbing is now a class-level hoist

The original approach threaded `{ signal: this.abortSignal }` through each provider's
call sites plus `tooled.js`, and added an `APIUserAbortError` rethrow to every catch
block. Two problems with that:

- **It only covered 10 provider keys.** openai, anthropic, gemini, groq, bedrock,
  xai, cerebras and the rest were never wired, so pressing stop on any cloud provider
  did nothing upstream. Coverage was tied to whoever remembered to add the param.
- Every new provider is a chance to forget it and would be annoying to maintain long term

Replaced with a single binder. Every SDK we use funnels agent completions through
`chat.completions.create`, `responses.create`, or `messages.create`; `attachAbortSignal`
wraps whichever the client has, once per client, and the wrapper reads `abortSignal` at
call time. Attaching moved into `getProviderForConfig` so no provider can be handed out
unbound.

The per-provider `{ signal }` threading, the `tooled.js` option, and all
the `APIUserAbortError` rethrows were reverted - nothing retries `RetryError`, and
`#safeProviderCall` already rethrows as-is when `_aborted`, so those rethrows only
suppressed one `console.error` line.

Three implementation notes worth carrying forward:

- **The signal must go through the SDK's request options, not `fetch`.** Measured: via
  `options.signal` → `APIUserAbortError` immediately, 1 request. Via a `fetch` wrapper the
  SDK doesn't recognize it as a user abort, **retries twice**, and throws
  `APIConnectionTimeoutError`.
- **Ollama is bound at the fetch layer deliberately.** Its SDK has `ollama.abort()`, but
  that only walks `ongoingStreamedRequests`, and a request isn't pushed there until
  *after* response headers arrive — so aborting while a local model loads would do
  nothing. The fetch wrap cancels at any phase.
- **`abortableClients()` is a method, not a getter, on purpose.** `InheritMultiple` copies
  prototype props by assignment, which evaluates getters at class-build time.

Bedrock overrides `abortableClients()` since Anthropic models use a second client for claude models since we need to use messages there because of Mantle API issues.

### 2. Loop guards are load-bearing, not redundant

An abort mid-stream **resolves** with a _partial_ completion rather than throwing. I
confirmed this against the SDK. That partial can contain a truncated tool call, so the
`if (this._aborted) return null` checks in `handleExecution`/`handleAsyncExecution` are
the only thing preventing us from acting on it. Please don't remove them as "covered by
the signal."

### 3. Summarization had three separate bugs

Reproduced with a stubbed connector on a 14-section document, aborting at 1.5s:

- **The flow executor never passed `controllerSignal` at all**, so `if (controllerSignal?.aborted)`
  was always false and the loop ran to the end. 
- **The in-flight section was never cancelled** on any path  the signal was only checked
  *between* sections, and the LC connectors had no abort plumbing so document summary tool would keep chugging even after abort.
- **Tool approval waited for the full 2-minute timeout** after an abort. If the loop was
  sitting on "N sections left, continue?" when you hit stop, the socket is already closed
  so no response could ever arrive so it would wait for timeout.

`summarizeContent` now takes the signal off the aibitat instance instead of trusting
callers to pass one, and binds it to the connector so the in-flight request dies.

Both plugins dropped their `controller: new AbortController()` field and listener/cleanup. The approval promise got a single `settle()` that tears down all three waiters (user response, abort, timeout).

### 4. Regular (non-agent) chat was not covered in original PR

This is the bigger half of #5905 and wasn't in the original branch at all. `clientAbortedHandler` just did `resolve(fullText)` and returned the upstream `for await` kept iterating, so the provider kept generating tokens for a response nobody would read. This was the same issue that was solved in the original PR for agents, but did not extend to regular chat.

Fixed at the four streaming entry points with one line each 
`chats/stream.js` (workspace chat / stop button), 
`chats/embed.js`
`chats/apiChatHandler.js`
 `chats/openaiCompatible.js`

Plus two cleanups so a stop reads as a stop: the shared `handleDefaultStreamResponseV2`
catch routes abort errors to `clientAbortedHandler` instead of logging `[STREAMING ERROR]`
(same guard added to the 4 custom handlers that have their own error log), and
`writeResponseChunk` drops chunks queued after the socket closed.

### 5. Shared module

The abort primitives live in `server/utils/helpers/abortSignals.js`**, rather than in the agent code, so the chat layer doesn't depend on `agents/` or being an aibitat connection any more.

### Still not covered (deliberately)

- **Ephemeral agents** (developer API, telegram bot) construct AIbitat but never call
  `abort()` — request/response with no stop button, nothing to wire.
- **Sync (non-streaming) API handlers** — no `response` to hook a close event on.
- **In-flight tool calls** aren't cancelled. An executing MCP call or skill finishes, then
  the loop stops at the next guard. Only LLM requests are torn down.
- **LangChain models** in the summarize/web-scraping plugins keep their existing
  `emitter.on("abort")` listeners — they're not SDK clients.
- **The `streamingEnabled() !== true` branch** classifies the abort as a generic error:
  `genericOpenAi.getChatCompletion` does `.catch((e) => { throw new Error(e.message) })`,
  flattening `APIUserAbortError` so `isAbortError` misses it. The request *is* cancelled;
  it just logs an error instead of a clean `[STREAM ABORTED]`.

Now we just need to reverify the provider to ensure no side effects so we can safely ship this huge upgrade to the UX.

Issue #4166, opened 2025-07-17
[FEAT]: How to use GitHub Copilot's LLM


constCD:
### What would you like to see?

GitHub Copilot provides model usage (https://docs.github.com/en/copilot/how-tos/build-copilot-extensions/building-a-copilot-agent-for-your-copilot-extension/using-copilots-llm-for-your-agent), But anythingLLM doesn't seem to work, Cherry studio has provided an integrated way

timothycarambat:
> You can call Copilot's LLM deployment at https://api.githubcopilot.com/chat/completions with a POST request. Requests and responses should follow the format as the [OpenAI API](https://platform.openai.com/docs/api-reference/chat/create).

You should be able to do this with the `Generic OpenAI Provider`
Base URL of `https://api.githubcopilot.com`
Then use the token as the auth token
Then use whatever the `model` is you wish to use.

Have you tried this?

pjmartorell:
I can confirm that this setup works. Here’s how I configured it:

* **LLM Provider:** Generic OpenAI
* **Base URL:** `https://api.githubcopilot.com`
* **API Key:** I couldn’t get a GitHub Personal Access Token (PAT) or a Fine-grained token to work. Instead, I extracted a GitHub Copilot access token from my JetBrains IDE, following the steps described [here](https://aider.chat/docs/llms/github.html). I’m not sure if this is the recommended or fully compliant method according to Github ToS, but it does work for now.
* **Chat Model Name:** I used `gpt-5`. You can find other available models in the [GitHub Models Marketplace](https://github.com/marketplace?type=models) or via the [GitHub Models](https://github.com/github/gh-models) extension.
* **Token Context Window:** 4096 (adjustable depending on the model)
* **Max Tokens:** 1024 (also adjustable based on your needs and model limits)

ddeedev:
> I can confirm that this setup works. Here’s how I configured it:
> 
> * **LLM Provider:** Generic OpenAI
> * **Base URL:** `https://api.githubcopilot.com`
> * **API Key:** I couldn’t get a GitHub Personal Access Token (PAT) or a Fine-grained token to work. Instead, I extracted a GitHub Copilot access token from my JetBrains IDE, following the steps described [[here](https://aider.chat/docs/llms/github.html)](https://aider.chat/docs/llms/github.html). I’m not sure if this is the recommended or fully compliant method according to Github ToS, but it does work for now.
[...]

PR #6076, opened 2026-08-02
feat: Add GitHub Copilot provider using official Copilot SDK (TOS-compliant)

### LINKED ISSUE #6075: [FEAT]: GitHub Copilot compliant integration (current Generic OpenAI approach violates GitHub TOS)
## Context

Issue #4166 discussed using GitHub Copilot's LLM in AnythingLLM. The suggested workaround was to use the **Generic OpenAI Provider** pointing to `https://api.githubcopilot.com` with a token extracted from an IDE.

## Problem: This violates GitHub's Terms of Service

`https://api.githubcopilot.com` is **not a public API**. It is an internal endpoint intended exclusively for official GitHub Copilot clients (VS Code, JetBrains, etc.). Using it as a generic OpenAI-compatible backend through a third-party tool like AnythingLLM violates multiple provisions of GitHub's Acceptable Use Policies:

Accessing `api.githubcopilot.com` outside of an official Copilot client is unauthorized access.

Using the Copilot chat completions endpoint as a drop-in OpenAI replacement for a generic LLM chat UI exploits the service in a way that was never intended or authorized.

### 4. Users are already getting blocked

As reported by @ddeedev in #4166, GitHub is actively blocking this usage:
> "403 Access to this endpoint is forbidden. Please review our Terms of Service."

## The Compliant Path: GitHub Copilot SDK

As @MoazAlkharfan pointed out in #4166, the correct way to integrate is via the official **GitHub Copilot SDK**:
- https://github.com/github/copilot-sdk

This is the authorized, TOS-compliant way for third-party tools to interact with Copilot's LLM.

## Request

Please implement a **GitHub Copilot provider** in AnythingLLM that uses the official Copilot SDK (or the authorized API surface it exposes), rather than the current "Generic OpenAI" hack that violates GitHub's Terms of Service and is already being blocked by GitHub.

This would give AnythingLLM users a compliant, future-proof way to use their Copilot subscription.

igitur:
## Summary

This PR adds a **TOS-compliant GitHub Copilot provider** that uses the official [`@github/copilot-sdk`](https://github.com/github/copilot-sdk) instead of the unauthorized `api.githubcopilot.com` REST endpoint.

## Background

The current workaround documented in #4166 suggests using the **Generic OpenAI** provider pointing to `https://api.githubcopilot.com` with a token extracted from an IDE. This violates multiple provisions of GitHub's Acceptable Use Policies:

- **AUP §5 (Site Access and Safety)**: Unauthorized access to `api.githubcopilot.com`, which is an internal endpoint for official Copilot clients
- **AUP §6 (Services Usage Limits)**: Exploiting the service without express written permission
- **TOS §H (API Terms)**: API abuse

GitHub is actively blocking this usage with 403 errors.

The compliant path is the [Copilot SDK](https://github.com/github/copilot-sdk), which communicates with the Copilot CLI via JSON-RPC — the authorized integration channel.

### New files
- `server/utils/AiProviders/githubCopilot/index.js` — LLM chat provider using `@github/copilot-sdk`
- `server/utils/agents/aibitat/providers/githubCopilot.js` — Agent provider for the aibitat framework

### Modified files
- `server/utils/agents/aibitat/providers/index.js` — Register the new provider
- `server/utils/helpers/updateENV.js` — Add `GITHUB_COPILOT_MODEL_PREF` and `GITHUB_COPILOT_MODEL_TOKEN_LIMIT` env keys
- `server/package.json` — Add `@github/copilot-sdk` dependency

## How it works

The provider uses the official Copilot SDK which spawns and manages the Copilot CLI process. Chat completions go through `CopilotClient.createSession()` → `session.sendAndWait(prompt)` instead of hitting the REST API directly.

### Provider features
- ✅ Non-streaming chat (`getChatCompletion`)
- ✅ Streaming chat (`streamGetChatCompletion`)
- ✅ Token limit configuration
- ✅ Model selection via env
- ✅ Shared CopilotClient singleton (avoids spawning multiple CLI processes)

### Prerequisites
Users need the [Copilot CLI](https://github.com/features/copilot/cli) installed and authenticated (`copilot auth login`).

## Related

- Closes #6075
- Supersedes the TOS-violating workaround in #4166
- See also: [MoazAlkharfan's comment](https://github.com/Mintplex-Labs/anything-llm/issues/4166#issuecomment-4124882590) pointing to the official SDK

## Test Plan

- [ ] Provider loads without error when `LLM_PROVIDER=github-copilot`
- [ ] Non-streaming chat returns valid response
- [ ] Streaming chat returns valid response
- [ ] Provider falls back gracefully when Copilot CLI is not installed


frontend/src/components/LLMSelection/GithubCopilotOptions/index.jsx
```diff
@@ -0,0 +1,251 @@
...
+            <option value="claude-sonnet-5">Claude Sonnet 5</option>
+            <option value="claude-sonnet-4.6">Claude Sonnet 4.6</option>
+            <option value="claude-sonnet-4.5">Claude Sonnet 4.5</option>
+            <option value="claude-haiku-4.5">Claude Haiku 4.5</option>
+            <option value="claude-opus-4.8">Claude Opus 4.8</option>
+            <option value="claude-opus-4.7">Claude Opus 4.7</option>
+            <option value="claude-opus-4.6">Claude Opus 4.6</option>
+            <option value="claude-opus-4.5">Claude Opus 4.5</option>
+            <option value="gemini-3.1-pro-preview">Gemini 3.1 Pro</option>
+          </select>
```
timothycarambat, line 51: There is no way to do a `/models` call via the `getCustomModels.js` call path? I really want to avoid static model lists as they become a massive burden longterm that we have to maintain.

frontend/src/components/LLMSelection/GithubCopilotOptions/index.jsx
```diff
@@ -0,0 +1,251 @@
...
+      <div className="flex gap-[36px] flex-wrap">
+        <div className="flex flex-col w-60">
+          <label className="text-white text-sm font-semibold block mb-3">
+            Model context window
+          </label>
+          <input
+            type="number"
+            name="GithubCopilotTokenLimit"
+            className="border-none bg-theme-settings-input-bg text-white placeholder:text-theme-settings-input-placeholder text-sm rounded-lg focus:outline-primary-button active:outline-primary-button outline-none block w-full p-2.5"
+            placeholder="Content window limit (eg: 4096)"
```
timothycarambat, line 64: maybe 16K? 4K will be unusable

frontend/src/components/LLMSelection/GithubCopilotOptions/index.jsx
```diff
@@ -0,0 +1,251 @@
...
+      setError(res.error);
+      setPhase("error");
+      return;
+    }
+    setDeviceAuth(res);
+    setPhase("show-code");
+    pollForToken(res.device_code);
+  };
+
+  const pollForToken = async (deviceCode) => {
```
timothycarambat, line 103: This should all be a hook for readability and reuse since it seems like this would be used in a lot of places

frontend/src/components/LLMSelection/GithubCopilotOptions/index.jsx
```diff
@@ -0,0 +1,251 @@
...
+          Authentication
+        </label>
+        <button
+          type="button"
+          onClick={startDeviceAuth}
+          className="bg-primary-button text-white text-sm px-6 py-2.5 rounded-lg hover:bg-primary-button/90 transition-all flex items-center gap-x-2"
+        >
+          <svg viewBox="0 0 24 24" width="18" height="18" fill="currentColor">
+            <path d="M12 0C5.37 0 0 5.37 0 12c0 5.31 3.435 9.795 8.205 11.385.6.105.825-.255.825-.57 0-.285-.015-1.23-.015-2.235-3.015.555-3.795-.735-4.035-1.41-.135-.345-.72-1.41-1.23-1.695-.42-.225-1.02-.78-.015-.795.945-.015 1.62.87 1.845 1.23 1.08 1.815 2.805 1.305 3.495.99.105-.78.42-1.305.765-1.605-2.67-.3-5.46-1.335-5.46-5.925 0-1.305.465-2.385 1.23-3.225-.12-.3-.54-1.53.12-3.18 0 0 1.005-.315 3.3 1.23.96-.27 1.98-.405 3-.405s2.04.135 3 .405c2.295-1.56 3.3-1.23 3.3-1.23.66 1.65.24 2.88.12 3.18.765.84 1.23 1.905 1.23 3.225 0 4.605-2.805 5.625-5.475 5.925.435.375.81 1.095.81 2.22 0 1.605-.015 2.895-.015 3.3 0 .315.225.69.825.57A12.02 12.02 0 0024 12c0-6.63-5.37-12-12-12z" />
+          </svg>
```
timothycarambat, line 201: ? Can this not be an asset - even locally if it is required.

frontend/src/media/llmprovider/github-copilot.png
timothycarambat, line 1: This should be GH logo since that is the branding for it

frontend/src/models/system.js
```diff
@@ -587,6 +587,36 @@ const System = {
...
+      .catch((e) => {
+        console.error(e);
+        return { error: e.message };
+      });
+  },
+  completeCopilotDeviceAuth: async function (deviceCode) {
+    return fetch(`${API_BASE}/system/copilot/device-auth/complete`, {
+      method: "POST",
+      headers: baseHeaders(),
+      body: JSON.stringify({ deviceCode }),
```
timothycarambat, line 609: All of these methods should likely be on its own model to keep system cleaner since these are endpoints **only** used with Github copilot.

server/endpoints/system.js
```diff
@@ -1093,6 +1093,48 @@ function systemEndpoints(app) {
...
+          "../utils/AiProviders/githubCopilot"
+        );
+        const result = await GithubCopilotLLM.initiateDeviceAuth();
+        return response.status(200).json(result);
+      } catch (error) {
+        console.error("Copilot device auth initiate failed:", error.message);
+        response.status(500).json({ error: error.message });
+      }
+    }
+  );
```
timothycarambat, line 1112: Would like these moved out of system as well.

server/utils/agents/aibitat/providers/githubCopilot.js
```diff
@@ -0,0 +1,160 @@
...
+  model;
+
+  constructor(config = {}) {
+    super();
+    this.providerTag = "github-copilot";
+    const { model = "auto" } = config;
+
+    this.model = model;
+    this.verbose = true;
+    this._supportsToolCalling = false; // Use UnTooled fallback for now
```
timothycarambat, line 25: GH CoPilot does not support tool-calling? This should almost **certainly** be supported since that is the only way tools can be used in the IDE

server/utils/helpers/customModels.js
```diff
@@ -1249,6 +1252,29 @@ async function getCerebrasModels() {
   }
 }
 
+function getGithubCopilotModels() {
```
timothycarambat, line 1255: This list is different from the frontend list. If we must statically type it should be in the `MODELS` constant/util. I also would very much urge finding a `/models` endpoint as maintaining any static list is an absolute nightmare since models are added/removed randomly and if the list is not in sync people start getting errors.

server/utils/helpers/updateENV.js
```diff
@@ -222,6 +222,20 @@ const KEY_MAPPING = {
     checks: [nonZero],
   },
 
+  // GitHub Copilot InferenceSettings (uses official Copilot SDK)
+  GithubCopilotGhToken: {
+    envKey: "GH_TOKEN",
+    checks: [],
```
timothycarambat, line 228: `notEmpty`

server/utils/helpers/updateENV.js
```diff
@@ -222,6 +222,20 @@ const KEY_MAPPING = {
...
+  },
+  GithubCopilotModelPref: {
+    envKey: "GITHUB_COPILOT_MODEL_PREF",
+    checks: [],
```
timothycarambat, line 232: `notEmpty`

timothycarambat (changes requested):
There are some issues pointed out in comments below for model endpoints, GH specific tooling endpoints, and some other nits. The main issue is with the usage of `@github/copilot-sdk` - which requires access to inProcess or TCP connection. This fundamentally is an issue because the gh copilot running **does not run in the container** on Docker - so it would not be reachable since the process is on host. I think the only way for this to work in docker is the user run the GH Copilot as TCP and then AnythingLLM connect to it via the Generic OpenAI - which would be ToS compliant.

This means that the copilot endpoint method is the only way to use this service in Docker. This method would only function on Desktop, so it should be hidden as a provider in Docker to prevent confusion. This worked when you were doing this locally since the server was not in a container - it wont work in prod.

This repo and desktop at 99% the same, so this code may work on desktop but it needs guards to prevent crashing on Docker. Also given the nature of how this works, it also needs documentation on https://github.com/Mintplex-Labs/anythingllm-docs to prevent people misunderstanding/misconfiguration which is bound to happen given all of the above.

igitur:
Thanks for the thorough effort, @timothycarambat . I'll be transparent and admit the PR was mostly created using agentic coding, but guided by me, which would explain some of the points on which you have valid criticism. I will address properly.

But do I understand correctly that there's absolutely no way to get GH Copilot work in the Docker version? And if so, do you think it's still worth pursuing this PR for Desktop only? I would hope so.

PR #6085, opened 2026-08-04
fix: accumulate token usage across all completion calls in agent runs

### LINKED ISSUE #6084: [BUG]: Agent runs only record token usage from the final completion call, not the full run

All versions

### What happened?

Token usage metrics saved for agent runs (`response.metrics` on the workspace chat, scheduled job run results, and the `usageMetrics` stream event) only reflect the **final completion call** of the agent loop — not the accumulated usage of the whole run.

An agent run with N tool calls makes N+1 LLM completion calls (one per tool-call turn plus the final response). Every one of those calls is billed separately by the provider, and each turn re-sends the full growing conversation, so the recorded number can undercount the real usage of a run by an order of magnitude.

**Why it happens:**

1. The base agent provider (`server/utils/agents/aibitat/providers/ai-provider.js`) stores usage in a single `lastUsage` object. `resetUsage()` is called at the start of every `complete()`/`stream()` call and `recordUsage()` **overwrites** `lastUsage` at the end — there is no accumulation anywhere.
2. The agent loop (`handleExecution` / `handleAsyncExecution` in `server/utils/agents/aibitat/index.js`) recurses once per tool call. Each recursion runs another completion, wiping the previous turn's usage. The `usageMetrics` event is only emitted on the terminal turn (the completion that produced no tool call) using `providerInstance.getUsage()` — i.e. the last turn only.
3. Everything downstream inherits the wrong value: the `chat-history` plugin saves `getUsage()` into the chat's `response.metrics`, and the scheduled-job helper stores the event payload last-write-wins.

**Expected:** metrics for an agent run should be the sum of `prompt_tokens` / `completion_tokens` / `total_tokens` across all completion calls in the run, so the values can be trusted for reporting and (future) cost calculation.

**Related (can be a follow-up issue):** providers on the UnTooled path (LM Studio, LocalAI, KoboldCPP, TextGenWebUI, …) never call `recordUsage()` at all, so agent runs on those providers report zero usage regardless of accumulation. They also make two API calls per loop turn (tool-selection + text completion), both currently untracked.

1. Run an `@agent` prompt that triggers one or more tool calls (e.g. a web search followed by a summary).
2. Inspect `response.metrics` on the saved row in `workspace_chats`.
3. The recorded `prompt_tokens` / `completion_tokens` match only the final completion call as reported by the provider, not the sum of all calls made during the run.

Any provider that records usage (OpenAI, Anthropic, Gemini, Ollama, etc.)

n/a

angelplusultra:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6084

### Description

Agent runs saved token usage from only the final completion call of the agent loop — every tool-call turn before it was overwritten, so `response.metrics` could undercount a run by an order of magnitude.

**Changes:**

- `ai-provider.js`: added a run-level `cumulativeUsage` record alongside the per-call `lastUsage`. Both are written through a new shared `applyUsage()` that `recordUsage()` feeds, so every completion adds to the run totals. New `getCumulativeUsage()`/`resetCumulativeUsage()` methods.
- `aibitat/index.js`: the loop resets the accumulator at depth 0 and all four `usageMetrics` emission sites now report `getCumulativeUsage()` instead of `getUsage()`.
- `chat-history.js`: persists the accumulated totals into `response.metrics`.
- `cerebras.js`: its `recordUsage` override wrote `lastUsage` directly and would have bypassed accumulation — now routes through `applyUsage()`. Also guards `total_tokens` against `undefined` (which would poison run totals as `NaN` under `+=`).
- `gemini.js`: stream handler called `recordUsage()` per chunk, which was harmless under overwrite semantics but would multi-count under accumulation (Gemini attaches `usage` to multiple chunks). Now captures the latest and records exactly once per stream — same pattern as `tooled.js`. All other providers were audited and record once per completion.

The metrics object shape is unchanged, so downstream consumers (ephemeral handler, scheduled jobs, frontend, Telegram bot) pick up correct totals with no changes. `duration` now sums LLM time across the run and `outputTps` is the run-wide average.

New tests cover accumulation, the reset lifecycle, Anthropic-style usage key normalization, per-instance isolation, and the `InheritMultiple` mixin providers.

**Known gap (pre-existing, out of scope):** providers running the UnTooled fallback path never call `recordUsage()`, so those runs still report zeros.

### Visuals (if applicable)

N/A — metrics are stored in `workspace_chats.response.metrics`; the chat UI renders the same string as before.

### Additional Information

Usage tracking was introduced in #5197 with per-call overwrite semantics; this PR makes the run-level consumers sum instead.

### Developer Validations

### DIFF
--- a/server/__tests__/utils/agents/aibitat/providers/ai-provider.test.js
+++ b/server/__tests__/utils/agents/aibitat/providers/ai-provider.test.js
@@ -1,0 +1,149 @@
+const Provider = require("../../../../../utils/agents/aibitat/providers/ai-provider.js");
+const UnTooled = require("../../../../../utils/agents/aibitat/providers/helpers/untooled.js");
+const InheritMultiple = require("../../../../../utils/agents/aibitat/providers/helpers/classes.js");
+
+class TestProvider extends Provider {
+  model = "test-model";
+
+  constructor() {
+    super(null);
+  }
+}
+
+// Mirrors how the UnTooled providers (LM Studio, LocalAI, Cerebras, etc.) are
+// declared - Provider's fields and methods arrive via the InheritMultiple mixin
+// rather than a direct prototype chain.
+class MixinProvider extends InheritMultiple([Provider, UnTooled]) {
+  model = "mixin-model";
+}
+
+describe("Provider usage tracking", () => {
+  test("recordUsage accumulates tokens across multiple completions", () => {
+    const provider = new TestProvider();
+
+    provider.resetUsage();
+    provider.recordUsage({
+      prompt_tokens: 100,
+      completion_tokens: 20,
+      total_tokens: 120,
+    });
+
+    provider.resetUsage();
+    provider.recordUsage({
+      prompt_tokens: 250,
+      completion_tokens: 40,
+      total_tokens: 290,
+    });
+
+    provider.resetUsage();
+    provider.recordUsage({
+      prompt_tokens: 400,
+      completion_tokens: 60,
+      total_tokens: 460,
+    });
+
+    // getUsage only reflects the most recent completion
+    const last = provider.getUsage();
+    expect(last.prompt_tokens).toBe(400);
+    expect(last.completion_tokens).toBe(60);
+    expect(last.total_tokens).toBe(460);
+
+    // getCumulativeUsage reflects the sum of all completions
+    const totals = provider.getCumulativeUsage();
+    expect(totals.prompt_tokens).toBe(750);
+    expect(totals.completion_tokens).toBe(120);
+    expect(totals.total_tokens).toBe(870);
+    expect(totals.model).toBe("test-model");
+    expect(totals.provider).toBe("TestProvider");
+  });
+
+  test("resetUsage does not clear the accumulated totals", () => {
+    const provider = new TestProvider();
+
+    provider.resetUsage();
+    provider.recordUsage({
+      prompt_tokens: 100,
+      completion_tokens: 20,
+      total_tokens: 120,
+    });
+
+    provider.resetUsage();
+    expect(provider.getUsage().total_tokens).toBe(0);
+    expect(provider.getCumulativeUsage().total_tokens).toBe(120);
+  });
+
+  test("resetCumulativeUsage zeroes the accumulated totals", () => {
+    const provider = new TestProvider();
+
+    provider.resetUsage();
+    provider.recordUsage({
+      prompt_tokens: 100,
+      completion_tokens: 20,
+      total_tokens: 120,
+    });
+
+    provider.resetCumulativeUsage();
+    const totals = provider.getCumulativeUsage();
+    expect(totals.prompt_tokens).toBe(0);
+    expect(totals.completion_tokens).toBe(0);
+    expect(totals.total_tokens).toBe(0);
+    expect(totals.model).toBe(null);
+    expect(totals.provider).toBe(null);
+  });
+
+  test("recordUsage normalizes Anthropic-style input/output token keys", () => {
+    const provider = new TestProvider();
+
+    provider.resetUsage();
+    provider.recordUsage({ input_tokens: 30, output_tokens: 10 });
+
+    provider.resetUsage();
+    provider.recordUsage({ input_tokens: 50, output_tokens: 15 });
+
+    const totals = provider.getCumulativeUsage();
+    expect(totals.prompt_tokens).toBe(80);
+    expect(totals.completion_tokens).toBe(25);
+    expect(totals.total_tokens).toBe(105);
+  });
+
+  test("instances do not share an accumulator", () => {
+    const providerA = new TestProvider();
+    const providerB = new TestProvider();
+
+    providerA.resetUsage();
+    providerA.recordUsage({
+      prompt_tokens: 100,
+      completion_tokens: 20,
+      total_tokens: 120,
+    });
+
+    expect(providerA.getCumulativeUsage().total_tokens).toBe(120);
+    expect(providerB.getCumulativeUsage().total_tokens).toBe(0);
+  });
+
+  test("accumulation works through InheritMultiple mixin providers", () => {
+    const providerA = new MixinProvider();
+    const providerB = new MixinProvider();
+
+    providerA.resetUsage();
+    providerA.recordUsage({
+      prompt_tokens: 100,
+      completion_tokens: 10,
+      total_tokens: 110,
+    });
+
+    providerA.resetUsage();
+    providerA.recordUsage({
+      prompt_tokens: 300,
+      completion_tokens: 30,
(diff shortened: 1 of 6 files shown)


server/utils/agents/aibitat/providers/cerebras.js
```diff
@@ -181,17 +181,12 @@ class CerebrasProvider extends InheritMultiple([Provider, UnTooled]) {
...
     if (time_info?.completion_time) duration = time_info.completion_time;
 
-    this.lastUsage = {
+    this.applyUsage({
       prompt_tokens: promptTokens,
       completion_tokens: completionTokens,
-      total_tokens: usage.total_tokens,
-      outputTps:
-        completionTokens && duration > 0 ? completionTokens / duration : 0,
+      total_tokens: usage.total_tokens || promptTokens + completionTokens,
```
shatfield4: 👍 Good catch here, if we just += this it would be undefined and would never give us an accurate value.

server/utils/agents/aibitat/providers/ai-provider.js
```diff
@@ -86,6 +88,24 @@ class Provider {
...
+  cumulativeUsage = {
+    prompt_tokens: 0,
+    completion_tokens: 0,
+    total_tokens: 0,
+    duration: 0,
+    outputTps: 0,
+    model: null,
+    provider: null,
+    timestamp: null,
+  };
```
shatfield4: Looks like a duplicate here with lines 715.

Can we convert this into a static `#emptyUsage()` function for the initializer and reset to avoid the duplicate object?

server/utils/agents/aibitat/providers/ai-provider.js
```diff
@@ -639,16 +659,68 @@ class Provider {
...
+  resetCumulativeUsage() {
+    this.cumulativeUsage = {
+      prompt_tokens: 0,
+      completion_tokens: 0,
+      total_tokens: 0,
+      duration: 0,
+      outputTps: 0,
+      model: null,
+      provider: null,
+      timestamp: null,
```
shatfield4: See comment on line 98.

server/utils/agents/aibitat/index.js
```diff
@@ -1187,7 +1191,11 @@ https://docs.anythingllm.com/agent/intelligent-tool-selection
     };
 
     // Emit routing notification before the first completion so it appears above the response
-    if (depth === 0) this?.flushRoutingMetadata?.(msgUUID);
+    // and reset the usage accumulator so metrics only cover this run's completions.
+    if (depth === 0) {
+      this?.flushRoutingMetadata?.(msgUUID);
+      this.providerInstance?.resetCumulativeUsage?.();
```
shatfield4: Do we need option chaining here? Every provider already extends the `Provider` class so this might be a point for silent failures if we keep this here.

server/utils/agents/aibitat/index.js
```diff
@@ -1025,7 +1025,11 @@ https://docs.anythingllm.com/agent/intelligent-tool-selection
     };
 
     // Emit routing notification before the first completion so it appears above the response
-    if (depth === 0) this?.flushRoutingMetadata?.(v4());
+    // and reset the usage accumulator so metrics only cover this run's completions.
+    if (depth === 0) {
+      this?.flushRoutingMetadata?.(v4());
+      this.providerInstance?.resetCumulativeUsage?.();
```
shatfield4: Same thing as comment on line 1197

shatfield4 (changes requested):
Small nits here, otherwise LGTM.

server/utils/agents/aibitat/providers/ai-provider.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6085/commits/4f6d7b5d87c4c5ef71bbd89eb75d88296874d6a0

server/utils/agents/aibitat/providers/ai-provider.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6085/commits/4f6d7b5d87c4c5ef71bbd89eb75d88296874d6a0

server/utils/agents/aibitat/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6085/commits/e2dd6b6ac3746d6da735a8e4e67e266477b46954

server/utils/agents/aibitat/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6085/commits/e2dd6b6ac3746d6da735a8e4e67e266477b46954

server/__tests__/utils/agents/aibitat/providers/ai-provider.test.js
```diff
@@ -0,0 +1,149 @@
...
+}
+
+// Mirrors how the UnTooled providers (LM Studio, LocalAI, Cerebras, etc.) are
+// declared - Provider's fields and methods arrive via the InheritMultiple mixin
+// rather than a direct prototype chain.
+class MixinProvider extends InheritMultiple([Provider, UnTooled]) {
+  model = "mixin-model";
+}
+
+describe("Provider usage tracking", () => {
```
timothycarambat, line 20: This test is not adversarial enough. This mostly tests happy paths. Throw some weird, null, malformed, and other data payloads it so we know it is robust and will not crash since it will interact with every provider.

server/utils/agents/aibitat/index.js
```diff
@@ -1025,7 +1025,11 @@ https://docs.anythingllm.com/agent/intelligent-tool-selection
     };
 
     // Emit routing notification before the first completion so it appears above the response
-    if (depth === 0) this?.flushRoutingMetadata?.(v4());
+    // and reset the usage accumulator so metrics only cover this run's completions.
```
timothycarambat, line 1028: Does `ephemeral` also benefit from this change as well? Those tokens should be captured as well

angelplusultra: Yeah these changes are on the `Provider` level and `Aibitat` level which `AgentHandler` and `EphemeralAgentHandler` own.

server/__tests__/utils/agents/aibitat/providers/ai-provider.test.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6085/commits/8a48b5e4c67b2bbf134afcb2dac43de7a3a6e917

shatfield4 (approved):
LGTM

Issue #3744, opened 2025-04-30
[FEAT]: Add Gitea data connector


nicodv:
### What would you like to see?

I would love to see a connector for https://about.gitea.com/.

Self-hosted gitea + AnythingLLM for local embeddings & models would be a killer combination in terms of privacy for me.

manuelkamp:
that would be very nice

weixinrui:
Lots of appreciation if a connector added for Gitea.

nprime496:
@timothycarambat curious if Gitea support might be on the roadmap at some point. It’d be really handy for local setups. Appreciate all the work you’re doing on this project!

supersongssr:
I really need this feature because I host my code on a self-hosted Gitea instance.

Talung:
Any progress on this?  I see you mentioned in the discord that the feature was already in "Early Access" back on the 26th November 2025

timothycarambat:
I have never committed to Gitea (I dont even know what it is!) but it seems like a way to self-host a Git repo - so it would be something that I would presume we would support as a data connector?

nonplayer:
I came here from a web search looking for this feature today. This would be killer for my work flow needs, as I have long since ditched GitHub for a more self-hosted approach to version control.

PR #6087, opened 2026-08-05
feat: add Gitea data connector

ArslanRasheed60:
## What
Adds Gitea as a repository data connector, alongside the existing GitHub and GitLab connectors, closing #3744.

Gitea is always self-hosted, so the API base is derived from the origin of the repo URL entered in the connector form (no separate "base URL" field needed) — the same approach the GitLab connector uses for self-hosted GitLab instances.

## Implementation
- **Collector:** new `GiteaRepoLoader` (`collector/utils/extensions/RepoLoader/GiteaRepo/`) using the Gitea v1 REST API — repo metadata for default-branch resolution, paginated branch listing, recursive git-tree traversal, raw file content fetch. Supports an optional access token, `ignorePaths`, and 429 backoff, and works unauthenticated against public repos. Registered `gitea` in the loader resolvers.
- **Server:** `gitea` allowed as a repo platform and as a watchable document type, including a `resyncGitea` handler so watched Gitea documents can re-sync.
- **Frontend:** new Gitea connector UI (repo URL, optional access token, branch selector, file ignores) with its own icon, matching the existing GitHub/GitLab connector cards, plus `gitea://` citation support.

## Testing
- Full existing collector + server test suite passes (366 tests, 36 suites).
- New unit tests for the Gitea loader (URL parsing, branch resolution, ignore-path filtering, error handling).
- Live end-to-end test against the public `gitea.com` instance: branch listing, default-branch resolution, recursive tree traversal, real file content fetch, and 404/invalid-URL handling all verified working, unauthenticated.
- Manually verified in the running app: the Gitea Repo option appears in Data Connectors between GitHub and GitLab, the form renders correctly, the branch dropdown populates live from a real Gitea instance, and submitting a real repo (`gitea.com/gitea/tea`) successfully collected all 324 files into the document library.

## Note
While implementing this, I noticed the existing GitLab watched-document resync handler is missing upstream (only `github` is registered in the resync map) — a pre-existing gap. I added the Gitea equivalent so this connector doesn't repeat it, but left the GitLab gap itself untouched since it's out of scope for this PR.


ArslanRasheed60:
@timothycarambat this closes #3744 — following up on your comment there about supporting Gitea as a data connector. Implementation mirrors the existing GitHub/GitLab loaders. Happy to adjust anything to match your preferred conventions.

PR #6094, opened 2026-08-06
fix: [BUG]: Workspace LLM Temperature is not applied to agent sessions, and no agent-level temperature setting exists

### LINKED ISSUE #6091: [BUG]: Workspace LLM Temperature is not applied to agent sessions, and no agent-level temperature setting exists

Docker (local)

### What happened?

The workspace **LLM Temperature** setting (Chat Settings → LLM Temperature / `openAiTemp`) is applied to normal chat, but appears to have **no effect on agent (`@agent`) sessions**. The workspace Agent Configuration exposes provider and model, but no temperature field — so there is no way for a user to control sampling temperature for agent tool-calling turns at all.

This matters because some models are temperature-sensitive at the post-tool-call turn. In my case, Gemma 4 (E4B, Q4_K_M) intermittently returns an **empty final answer** after a successful tool call — the tool executes correctly, but `textResponse` comes back `""` with `completion_tokens: 1` (the model samples an immediate end-of-turn token). Lowering temperature eliminates this at the provider level, but there is no way to apply that fix to agent sessions in AnythingLLM.

**Environment:** AnythingLLM v1.15.0 in Docker (local), macOS 26.4, Apple Silicon (M5 Max). Agent skill used: a minimal custom skill `add_two_numbers` (also reproduces with other skills).

**Evidence that temperature controls the underlying behavior (provider-direct, AnythingLLM bypassed):** calling the provider directly and hand-feeding the tool result via an OpenAI-format `role:"tool"` message:

| Provider-direct temperature | Result over repeated runs |
|---|---|
| 0.7 | 3 spoke / 5 empty (8 runs) |
| 0 | 4 spoke / 0 empty (4 runs, identical output) |

**Evidence that the workspace setting does not reach agent sessions:** identical repeated request via the developer API (`POST /api/v1/workspace/{slug}/chat`, message `"@agent Use the add_two_numbers tool to add 100 and 200"`):

| Workspace LLM Temperature | Result over 8 runs |
|---|---|
| 0.7 | ~half empty (matches provider-direct 0.7 odds) |
| 0.3 (saved via Update workspace) | 4 spoke / 4 empty — **no change** |

If the workspace temperature were applied to the agent's LLM calls, the 0.3 runs should have shown a large reduction in empty responses (as the provider-direct tests demonstrate). They showed none. In every empty-response case the agent loop itself worked correctly: `thoughts` shows the tool executing with correct arguments, and `metrics` reports `completion_tokens: 1`.

Example empty-response payload:

```json
{"type":"textResponse","sources":[],"close":true,"error":null,
 "textResponse":"",
 "thoughts":["@agent is executing `add_two_numbers` tool {\n  \"a\": 100,\n  \"b\": 200\n}","add_two_numbers: 100 + 200"],
 "metrics":{"prompt_tokens":354,"completion_tokens":1,"total_tokens":355}}
```

**Expected behavior:** either agent sessions inherit the workspace `openAiTemp`, or Agent Configuration exposes its own temperature setting.

**Related suggestion (optional):** when the final post-tool-call completion returns empty content, the agent loop could retry once (optionally at reduced temperature) before returning an empty `textResponse` — currently the user just sees a blank reply under the tool indicator, which reads as a broken app even though the tool ran fine

1. Generic OpenAI provider pointed at any llama.cpp-based server running Gemma 4 E4B Q4_K_M (or another model whose post-tool-call turn is temperature-sensitive).
2. Workspace with one custom agent skill enabled; set workspace LLM Temperature to 0.3, Update workspace.
3. Send `@agent <use the skill>` repeatedly (chat window or `/api/v1/workspace/{slug}/chat`).
4. Observe intermittent empty `textResponse` with `completion_tokens: 1` at the same rate as temperature 0.7 — the 0.3 setting has no effect on agent runs.

Generic OpenAI → Docker Model Runner v1.2.6 (llama.cpp b9879-metal, rev 72874f5) / docker.io/ai/gemma4:latest (Gemma 4 E4B-It, GGUF Q4_K_M)

AnythingLLM built-in embedder (all-MiniLM-L6-v2) — not involved in this bug.

ojassharma7:
Fixes #6091.

## What changed
- `server/__tests__/utils/agents/aibitat/providers/helpers/tooled.test.js`
- `server/utils/agents/aibitat/index.js`
- `server/utils/agents/aibitat/providers/ai-provider.js`
- `server/utils/agents/aibitat/providers/helpers/tooled.js`
- `server/utils/agents/ephemeral.js`
- `server/utils/agents/index.js`

## Verification
The project's own test suite was run before and after this change; it introduces no new test failures or lint violations.

### DIFF
--- a/server/__tests__/utils/agents/aibitat/providers/helpers/tooled.test.js
+++ b/server/__tests__/utils/agents/aibitat/providers/helpers/tooled.test.js
@@ -1,5 +1,7 @@
 const {
   formatMessagesForTools,
+  tooledStream,
+  tooledComplete,
 } = require("../../../../../../utils/agents/aibitat/providers/helpers/tooled.js");
 
 describe("formatMessagesForTools attachment content (native tool path)", () => {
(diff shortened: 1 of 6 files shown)


timothycarambat:
Moving this to a core-team issue. There is a reason why we havent supported this and while seemingly trivial it can cause some problems if we add it like it is in this PR where models throw on the mere existences of temp being in the payload

Issue #1945, opened 2024-07-24
[FEAT]: Adjust Max Output Tokens


luketech85:
### What would you like to see?

I would like to be able to adjust the max output tokens. It seems to be quite restricted at the moment when I know it can output more very effectively (for example with Claude Sonnet 3.5).

PR #6097, opened 2026-08-07
feat: add configurable max_tokens for Together AI

### LINKED ISSUE #6071: [FEAT]: Add the option to set max_tokens in Together AI config

Currently, when using Together AI, there is no way to set `max_tokens`. Together AI's default for `max_tokens` is 2048. This typically results in truncation of the responses, which **makes Together AI essentially unusable with thinking models** (and, if you use the application specifically in the described scenario, as I do, unusable altogether).

It is possible to register Together AI as a "Generic OpenAI" provider. One can then use the UI to set `max_tokens` to the desired value and thus prevent truncation, but the functionality gets severely degraded (fetching the model list does not work so no way to select the model in chat; also no way to see the thinking trace).

I feel the best way to address this would be to expose `max_tokens` in the Together AI's config UI in the same way it's exposed for Generic OpenAI.

This issue is essentially #1945 but I'm describing it here because it's much more severe specifically for Together AI (and, possibly, some other providers too).

#6048 is related - it may or may not be a part of the solution.

richboyneedcash:
Closes #6071

### Problem
Together AI's API defaults `max_tokens` to 2048, which truncates responses and makes thinking models effectively unusable via the native Together AI provider. Today the only workaround is to register Together AI as a **Generic OpenAI** provider to get the `max_tokens` field — but that loses model-list fetching (no model dropdown) and the thinking trace.

### Change
Bring the same `max_tokens` option the Generic OpenAI provider already has to the native Together AI provider:

- **Provider**: read `TOGETHER_AI_MAX_TOKENS` (default `1024`, via `toValidNumber`) and pass `max_tokens` on both the sync (`getChatCompletion`) and streaming (`streamGetChatCompletion`) calls
- **Config**: register `TogetherAiMaxTokens` in `updateENV` (`nonZero` check) and expose it in `systemSettings`
- **UI**: add a **Max Tokens** input to the Together AI config (mirrors the Generic OpenAI options)
- **Docs**: document `TOGETHER_AI_MAX_TOKENS` in `server/.env.example` and `docker/.env.example`

### Notes
Mirrors the existing `GENERIC_OPEN_AI_MAX_TOKENS` pattern for consistency. Backend files pass `node --check`. This is essentially #1945 but, as the reporter notes, far more impactful for Together AI specifically.

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -51,6 +51,7 @@ GID='1000'
 # LLM_PROVIDER='togetherai'
 # TOGETHER_AI_API_KEY='my-together-ai-key'
 # TOGETHER_AI_MODEL_PREF='mistralai/Mixtral-8x7B-Instruct-v0.1'
+# TOGETHER_AI_MAX_TOKENS=1024
 
 # LLM_PROVIDER='mistral'
 # MISTRAL_API_KEY='example-mistral-ai-api-key'
--- a/frontend/src/components/LLMSelection/TogetherAiOptions/index.jsx
+++ b/frontend/src/components/LLMSelection/TogetherAiOptions/index.jsx
@@ -25,7 +25,24 @@ export default function TogetherAiOptions({ settings }) {
         />
       </div>
       {!settings?.credentialsOnly && (
-        <TogetherAiModelSelection settings={settings} apiKey={apiKey} />
+        <>
+          <TogetherAiModelSelection settings={settings} apiKey={apiKey} />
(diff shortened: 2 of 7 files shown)


timothycarambat:
https://github.com/Mintplex-Labs/anything-llm/issues/6071#issuecomment-5134433201

We dont want to provider-level patch each of these possible configs as it bloats the ENV needlessly - that comment links to a PR we are going to merge this across all 20+ providers.

PR #6102, opened 2026-08-07
feat: track per-message LLM cost in chat metrics

angelplusultra:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

Translations #6103

### Description

Adds per-message USD cost tracking to chat and agent metrics. This is plumbing for upcoming cost features — costs are recorded in the metrics of every chat and agent run, but intentionally **not displayed anywhere in the UI yet**, since individual chat/run costs are usually too small to be useful on their own.

**Server**

- New `ModelPricing` helper (`server/utils/helpers/modelPricing/`) that sources pricing from [models.dev](https://models.dev), cached on disk for 3 days with ETag revalidation. Falls back to the pricing snapshot bundled in `@opencode-ai/models` when offline/air-gapped.
- Cost (`inputCost`/`outputCost`/`totalCost`) is added to metrics in all chat flows (workspace stream, developer API, embed, OpenAI-compatible endpoint, Telegram bot) and agent runs. Agent runs price each completion call individually, so totals stay correct even if the model changes mid-run.
- Local/self-hosted providers (Ollama, LM Studio, etc.) report $0. Unknown providers/models omit the cost fields entirely — pricing degrades to "unknown", never a wrong number.
- Works with the model router: cost is calculated against the resolved delegate provider/model, not the router itself.
- New `display_currency` system setting (admin/manager editable, validated against supported currencies) and a `GET /system/exchange-rates` endpoint. Exchange rates come from [Frankfurter](https://frankfurter.dev), cached on disk server-side for 30 days with a 10s fetch timeout — air-gapped installs degrade to USD.

**Display currency (built, not yet wired in)**

The display currency setting and cost conversion pipeline are fully built and ready to go — they just need to be wired in once we have a feature we want to build around cost:

- `CurrencyPreference` dropdown (Settings → Interface) is complete but commented out of the page with a note.
- `useCurrency` hook fetches the instance currency + rates and formats stored USD costs for display; currently unused, documented as the entry point for future cost surfaces.

Includes #6085 (cumulative agent token usage), which this builds on — agent cost is derived from the accumulated per-call usage that PR introduces.

### Visuals (if applicable)

N/A — no user-facing changes. Cost lives only in stored metrics for now.

### Additional Information

Pricing data adds no hard runtime dependency: if models.dev is unreachable and no cache exists, the bundled snapshot is used; if that also fails, chats simply record no cost.

### Developer Validations


server/utils/agents/aibitat/providers/ai-provider.js
```diff
@@ -75,16 +81,47 @@ class Provider {
...
+   * numeric strings, nulls, negative or non-finite values), so anything that
+   * does not resolve to a usable number becomes 0.
+   * @param {unknown} value
+   * @returns {number}
+   */
+  static #toSafeMetric(value) {
+    const number = Number(value);
+    if (!Number.isFinite(number) || number < 0) return 0;
+    return number;
+  }
```
shatfield4: See previous comment on `#toSafeMetric` to dedupe and form into a util

server/utils/helpers/modelPricing/index.js
```diff
@@ -0,0 +1,434 @@
...
+   * provider reporting a negative or non-finite count must not produce a
+   * negative or infinite cost.
+   * @param {unknown} value
+   * @returns {number}
+   */
+  static #safeTokenCount(value) {
+    const number = Number(value);
+    if (!Number.isFinite(number) || number < 0) return 0;
+    return number;
+  }
```
shatfield4: This looks like the same logic as the `#toSafeMetric` function in ai-provider.js. Let's move this to a util and reuse it in both place.

shatfield4 (changes requested):
Small nit here with deduping 2 functions with the same logic in the quick review that I did.

The only other thing that I would investigate is if adding the new fields to the metrics/usage object causes any side effects (specifically in the openai compat responses). I'm not exactly sure if this can cause errors across other providers but lets be sure to test this.

server/utils/agents/aibitat/providers/ai-provider.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6102/commits/0c2bd2e47dd5dffe1fc082c743a47c676ffc798a

server/utils/helpers/modelPricing/index.js
angelplusultra: https://github.com/Mintplex-Labs/anything-llm/pull/6102/commits/0c2bd2e47dd5dffe1fc082c743a47c676ffc798a

PR #6105, opened 2026-08-07
Fix generic openai thinking parsers

### LINKED ISSUE #6099: [BUG]: Generic OpenAI provider does not parse vLLM reasoning field

Docker (local)

### What happened?

When a reasoning model served by vLLM is used through the Generic OpenAI provider, vLLM returns reasoning text in `message.reasoning` for non-streaming responses and `delta.reasoning` for streaming chunks. The provider currently reads only `reasoning_content`, so the reasoning text is omitted and the Thinking panel is not populated.

The Generic OpenAI provider should accept both fields while preserving the existing `reasoning_content` behavior. When both fields are present, `reasoning_content` should continue to take precedence.

1. Configure an OpenAI-compatible vLLM endpoint with a reasoning-capable model.
2. Send a prompt that produces reasoning, using both streaming and non-streaming responses.
3. Observe that vLLM returns `reasoning` / `delta.reasoning`, but AnythingLLM omits the reasoning text.
4. Compare with a response using `reasoning_content`, which is parsed correctly.

Generic OpenAI-compatible provider / vLLM with a reasoning-capable model

_No response_

timothycarambat:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6099
closes #6100

### Description

Supports both async and agentic chat thinking parsers for Generic OpenAI connector. Testing with deepseek (reasoning_content) LMStudio (reasoning) and Ollama (thinking) payload responses. Falls back to `undefined` if not found or parseable.

### Developer Validations

### DIFF
--- a/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
+++ b/frontend/src/components/WorkspaceChat/ChatContainer/ChatHistory/HistoricalMessage/index.jsx
@@ -22,6 +22,20 @@ import HistoricalOutputs from "./HistoricalOutputs";
 import HistoricalClarifyingQuestions from "./HistoricalClarifyingQuestions";
 import { openImageLightbox } from "@/components/ImageLightbox";
 
+function hasVisibleContent(message) {
+  if (!message) return false;
+  const stripped = message
+    .replace(new RegExp(THOUGHT_REGEX_COMPLETE, "g"), "")
+    .trim();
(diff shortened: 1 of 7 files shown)


shatfield4 (approved):
LGTM

I was able to test a few models using docker model runner and lemonade. Everything is working as expected and with both providers, I used a thinking model and non-thinking.

PR #6131, opened 2026-08-16
fix: select the physical LAN adapter for the mobile pairing QR code

### LINKED ISSUE #6129: [BUG]: Mobile pairing QR code points to the wrong network interface on Windows

AnythingLLM desktop app

### What happened?

The generated pairing QR code for AnythingLLM Mobile points to the IP address of a VPN network interface of my Windows machine instead of the Ethernet interface. This results in "Network request failed" error in the mobile app when trying to connect using the QR code.

_No response_

_No response_

_No response_

ArslanRasheed60:
resolves #6129

On Windows, the mobile pairing QR code can point at the wrong network interface (e.g. a VPN or virtual adapter) instead of the physical LAN adapter, so the phone can't reach the desktop app.

Adds `getLocalNetworkAddress()` to pick the physical LAN interface instead of taking the first address returned by `ip.address()`, and uses it when building the mobile pairing URL.

### DIFF
--- a/frontend/src/pages/GeneralSettings/MobileConnections/ConnectionModal/index.jsx
+++ b/frontend/src/pages/GeneralSettings/MobileConnections/ConnectionModal/index.jsx
@@ -87,13 +87,13 @@ export default function MobileConnectModal({ isOpen, onClose }) {
  */
 function processConnectionUrl(url) {
   /*
-   * In dev mode, the connectionURL() method uses the `ip` module
+   * In dev mode, the connectionURL() method uses the local LAN address
    * see server/models/mobileDevice.js `connectionURL()` method.
    *
    * In prod mode, this method returns the absolute path since we will always want to use
    * the real instance hostname. If the domain changes, we should be able to inherit it from the client side
    * since the backend has no knowledge of the domain since typically it is run behind a reverse proxy or in a container - or both.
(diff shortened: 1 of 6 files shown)


ArslanRasheed60:
@timothycarambat could you take a look when you have a chance?

timothycarambat:
The docker (this) repo does not need this - only desktop does. There is an easier fix though:
```node
const dgram = require("dgram");

/**
 * Finds the LAN address of this machine by asking the OS routing table
 * which source IP it would use to reach an external destination.
 * No packet is actually sent - the UDP socket is connected but not written to.
 * @returns {Promise<string|null>}
 */
function getLocalNetworkAddress() {
  return new Promise((resolve) => {
    const sock = dgram.createSocket("udp4");
    sock.connect(53, "8.8.8.8", () => {
      const addr = sock.address().address;
      sock.close();
      resolve(addr && addr !== "0.0.0.0" ? addr : null);
    });
    sock.on("error", () => {
      sock.close();
      resolve(null);
    });
  });
}

module.exports = { getLocalNetworkAddress };
```

This will always resolve the correct address space since this util is only used when AnythingLLM is already attached to `0.0.0.0` - so this uses the dgram built-in to resolve the routing table internally without guessing namespace or hardware names.

PR #6138, opened 2026-08-17
New AnythingLLM wordmark + fix logo theme resolution

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #

### Description

- replace AnythingLLM wordmark with the new brand mark in both light and dark variants
- updates README and Open Computer AnythingLLM logos
- fix the default logo lookup pointing at a filename that does not exist on disk (made every light mode logo request fall through to the bundled frontend copy)
- fix logo endpoint keying off a theme value the app stopped sending, so it served the wrong variant for every theme
- send the resolved theme when requesting the logo so "system" picks the right variant instead of being passed through unresolved
- make sure logo is the same size on home page and in settings

### Developer Validations


frontend/src/models/system.js
```diff
@@ -464,7 +464,7 @@ const System = {
     const url = new URL(`${fullApiUrl()}/system/logo`);
     url.searchParams.append(
       "theme",
-      localStorage.getItem("theme") || "default"
```
timothycarambat, line 467: Why did we need to change this - was it not working properly prior?

server/endpoints/system.js
```diff
@@ -711,8 +711,7 @@ function systemEndpoints(app) {
 
   app.get("/system/logo", async function (request, response) {
     try {
-      const darkMode =
-        !request?.query?.theme || request?.query?.theme === "default";
+      const darkMode = request?.query?.theme !== "light";
```
timothycarambat, line 714: Did this need to be bumped? Should just be replacing the images

server/utils/files/logo.js
```diff
@@ -5,7 +5,7 @@ const { v4 } = require("uuid");
 const { SystemSettings } = require("../../models/systemSettings");
 const { normalizePath, isWithin } = require(".");
 const LOGO_FILENAME = "anything-llm.png";
-const LOGO_FILENAME_DARK = "anything-llm-dark.png";
+const LOGO_FILENAME_DARK = "anything-llm-invert.png";
```
timothycarambat, line 8: Was `-dark` not a real file or something? Why `invert` now?

timothycarambat (commented):
quick questions - see above

server/utils/files/logo.js
timothycarambat: > fix the default logo lookup pointing at a filename that does not exist on disk (made every light mode logo request fall through to the bundled frontend copy)

This was fixing that, yeah?

shatfield4: Correct it was fixing that part where the dark file didn't exist but it was falling back to the bundled image in the frontend.

server/endpoints/system.js
shatfield4: Default theme no longer exists, it's a legacy theme that was leftover from before. Only theme types that exist now are: system, light, dark

frontend/src/models/system.js
shatfield4: Yes, this was broken in a weird way actually. `localStorage` only holds `system`, `light`, or `dark` now, so `theme === "default"` never matches and `darkMode` was always false and the endpoint resolved `LOGO_FILENAME_DARK` every time.

That was hidden because `LOGO_FILENAME_DARK` pointed at `anything-llm-dark.png`, which didn't exist in `server/storage/assets`. Every request 204'd and `LogoContext` fell back to the bundled logo.

We're now using the `data-theme` now because localstorage stores the preference which could be `system` and the backend wasn't able to resolve that to `light` or `dark` but `data-theme` always resolves to `light` or `dark` so it updates on its own when users switch from light or dark mode on the system level.

PR #6140, opened 2026-08-18
Add generate-image agent skill

### LINKED ISSUE #6139: [FEAT]: Image generation agent skill

We should add an agent skill for `generate-image` which will take parameters:
- prompt
- steps
- dimensions (eg: 512x512, 1024x1024)
- edit feature to pass images into the skill?

Needs to support:
- Status response events
- Response states for agent mode

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6139 

### Description

- add a `generate-image` agent skill so images can be generated via a tool call
- skill accepts a prompt, optional dimensions, and an edit flag for transforming an image already in the conversation
- edits use the images attached to the last user message, falling back to the image generated earlier in the session
- show a placeholder card while the image is generating that swaps for the result, or clears if generation fails or is cancelled
- send the generated image inline over the socket so it renders immediately, and persist it to the chat so it rerenders on reload
- when no image generation provider is configured the skill tells the agent to point the user at Settings > Image Generation
- hide the skill in the in-chat tools menu until a provider is configured, while always listing it on the agent skills settings page

### Additional Information

- Will create translations PR after review is approved

### Developer Validations


frontend/src/pages/Admin/Agents/GenericSkillPanel/index.jsx
timothycarambat, line 1: Using one of the images we have for headers already plop it into Gemini or something to make an image one **or ask Tiff** and I am sure we can get an image supported easily.

frontend/src/utils/chat/agent.js
```diff
@@ -314,16 +315,40 @@ export default function handleSocketResponse(socket, event, setChatHistory) {
...
+      );
+      if (data.failed) return history;
+      return [
+        ...history,
         {
           uuid: v4(),
           type: "textResponse",
           content: data.content,
           outputs: data.outputs || [],
-          chatId: data.chatId || null,
```
timothycarambat, line 326: we dont need this anymore no?

server/endpoints/agentFileServer.js
```diff
@@ -179,10 +179,10 @@ async function findInWorkspaceChats(storageFilename, { user, isMultiUser }) {
   const workspaceIds = workspaces.map((w) => w.id);
   if (workspaceIds.length === 0) return null;
 
-  // DB-level filter so we don't load every chat into memory.
+  // DB-level filter so we don't load every chat into memory. `include` is skipped
+  // since an agent's chat row is still hidden while it generates files mid-turn.
   const chats = await WorkspaceChats.where({
     workspaceId: { in: workspaceIds },
-    include: true,
```
timothycarambat, line 185: Why does this need to go? We only want image that are finalized and attached to a real chat? If the workspace is cleared then we dont need the image anymore?

server/utils/agents/aibitat/plugins/websocket.js
```diff
@@ -152,8 +152,8 @@ const websocket = {
         // expose function for sockets across aibitat
         // type param must be set or else msg will not be shown or handled in UI.
         aibitat.socket = {
-          send: (type = "__unhandled", content = "") => {
-            socket.send(JSON.stringify({ type, content }));
+          send: (type = "__unhandled", content = "", extras = {}) => {
+            socket.send(JSON.stringify({ type, content, ...extras }));
```
timothycarambat: What is the pattern for us to send like FileDownloadCards? It should work something like that and not need to send extras over send - will that pattern not work?

timothycarambat (changes requested):
Just some questions, nit, and potentially a refactor opp for frontend for detecting if imgen is supported.

server/endpoints/agentFileServer.js
shatfield4: Reverted this since it was more of a band-aid fix. The problem was that as an agent is generating an image it could not serve it since the agent handler marks chats as complete after the agent stops working. The state would flip from pending -> done but was never able to serve that file until we refreshed the page since the agent still keeps streaming more after the image is generated.

Fixed this now by only letting chats that are still pending through. I did this by checking if the row has text since rows that have no text will be pending generations.

frontend/src/utils/chat/agent.js
shatfield4: Undid this since chatId is still sent by the /img path inside an agent session and it is what the card's edit/delete/feedback actions key off of. Moved to data.content now to use the same pattern as FileDownloadCards.

server/utils/agents/aibitat/plugins/websocket.js
shatfield4: Removed this as well and matched the same pattern as FileDownloadCard. The payload now uses data.content and switched /img command to use the same shape so both match.

frontend/src/components/WorkspaceChat/ChatContainer/PromptInput/ToolsMenu/Tabs/AgentSkills/useAgentSkillsState.js
```diff
@@ -48,6 +50,7 @@ export default function useAgentSkillsState(defaultSkills) {
           AgentFlows.listFlows(),
           System.isFileSystemAgentAvailable(),
           System.isMultiUserMode(),
+          System.keys(),
```
timothycarambat: Instead of keys can we do the same as we do for `isFileSystemAgentAvailable` and just make a simple dedicated endpoint? I would then think we can migrate the current conditional for this on the slash command modal to also use this endpoint to determine if `/img` is allowed - unless that is handled dramatically different.

timothycarambat (changes requested):

PR #6152, opened 2026-08-20
Fix agent backend crash from oversized tool reranker query

### LINKED ISSUE #6145: [BUG]: Backend helper process crashes during agent sessions; ~830 MB leaked per agent exchange
On AnythingLLM Desktop 1.16.0 (macOS), agent-mode sessions intermittently end
immediately with **"Agent session complete."** and no output. The UI stays up, so
it looks like the agent simply gave up. It is not — **the backend helper process
has crashed.** After it happens, any subsequent action fails ("Could not respond
to message… Failed to fetch"), and opening the settings pane spins forever.
Quitting and reopening the app always fixes it, because that starts a new
backend.

macOS wrote a crash report for **every** occurrence. I had seven in one day.

### Two findings, which may or may not be the same bug

**1. The backend leaks about 830 MB per agent exchange, and never gives it back.**

Sampled the backend process (the one listening on 127.0.0.1:3001) every two
minutes:

```
BACKEND #1
17:50   71.4 MB   ┐
  …               ├ 14 minutes idle          +0.3 MB
18:04   71.7 MB   ┘
18:06   96.7 MB   ← one agent exchange begins
18:08  907.5 MB                              +835.8 MB
18:10–18:22  ~893 MB  (holds; never released)
18:26  process gone (crashed 18:25:57)

BACKEND #2  (after restarting the app)
18:34  189.5 MB   ← fresh process
18:36 1012.2 MB   ← one agent exchange       +822.7 MB
18:38 1003.6 MB
18:40  process gone (crashed 18:38:49)
```

Idle costs nothing. Each agent exchange costs ~830 MB, whether the exchange
succeeds or fails. Both measured exchanges were ordinary: a short question, one
tool call, a short answer.

This resembles #5801 / PR #5804 ("100–650 MB per RAG chat message"), but that
fix merged 2026-06-11, well before 1.16.0, so either it does not cover this path
or this is a second leak.

**2. The crash itself.** From the macOS crash reports:

```
exception    EXC_BREAKPOINT (SIGTRAP)
termination  SIGNAL code=5, byProc: "exc handler", byPid: <the backend pid>
frames       Electron Framework … libsystem_malloc.dylib _malloc_zone_memalign
```

The process aborts itself. I searched the reports specifically for
"JavaScript heap out of memory" and found **no such string**, so despite the
leak I cannot claim this is a heap-exhaustion crash.

### A reliable marker in your own data

Every crash is preceded, within 3–249 seconds, by an agent session left
**`closed = 0`** in `workspace_agent_invocations`. Seven crashes, seven unclosed
sessions, one-to-one, no exceptions. Sessions that close normally never crash.

Sessions per backend before it died: 3, 4, 2, 2, 4, 2, 1 — so it is not a fixed
count.

### Where it dies

In `storage/logs/backend-YYYY-MM-DD.log`, the crashing session's last line is
always the same, and there is never an error line after it:

```
[AgentHandler] Injecting N parsed file(s) into user message
```

N was 3 on one crash and 1 on another, so it is not caused by the number of
attached files.

### The LLM provider is not involved

The provider is a local OpenAI-compatible server (`mlx_lm.server`) on another
machine. During a failure I queried it directly with curl and it answered in
**0.868 s**, same process, four hours uptime, 0.1% CPU. The settings pane also
hangs after a crash, and that path never touches the LLM at all.

### Environment

- AnythingLLM Desktop **1.16.0**, macOS 15.7.7, Apple Silicon, 16 GB
- LLM provider: **Local AI** (OpenAI-compatible), remote host over a private network
- Embedder: built-in. Vector DB: LanceDB. `rag-memory` disabled.
- Agent skills on: `filesystem-agent`, `web-browsing`, `create-files-agent`,
  plus one MCP server (6 tools). ~31 tools attached per session.
- Workspace had 1–3 small text files attached (~1,000–8,200 tokens total).

### Steps to reproduce

1. Desktop app, workspace in agent mode, Local AI provider pointed at any
   OpenAI-compatible server.
2. Enable the filesystem, web-browsing and file-creation skills, plus an MCP
   server.
3. Attach one small text file to the workspace.
4. Send an agent request that uses a tool. Watch the backend process memory
   (`lsof -nP -tiTCP:3001 -sTCP:LISTEN`, then `ps -o rss= -p <pid>`).
5. Memory jumps several hundred MB and stays. Repeat. Within a few exchanges a
   session ends instantly with "Agent session complete.", a crash report appears
   in `~/Library/Logs/DiagnosticReports/`, and the app is unusable until
   restarted.

### What would help

Whether the PR #5804 singleton caching covers the agent path with attached
parsed files, and whether anything downstream of `Injecting N parsed file(s)`
can abort the process rather than throwing an error that the session can report.

shatfield4:
### Pull Request Type

- [ ] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6145 

### Description

- previous behavior was injecting attached/pinned workspace documents into the tool reranker filling up the entire 512 context window size before injecting the tools
- moved `extractUserPrompt` to be called before injecting attached/pinned documents
- change reranker to no longer inject attached/pinned documents into the tool reranker and only inject the last user message for tool reranking
- change reranker to batch calls instead of doing 1 single call with all tools/context to prevent large spikes in RAM
- onnxruntime allocated the amount of memory it needs for each reranking call and when doing 1 large call, it allocates a large spike of ~1gb of memory and does not release it back to the OS
- batching calls to the reranking model allocates only enough memory for each call since each call after reuses the allocated memory and only increases it if required by the model so the overall memory usage is lower and the batched calls actually finish much quicker than 1 large call
- removed `chunkedRerank` function since our modified rerank function now does this in batches

### Developer Validations

### DIFF
--- a/server/__tests__/utils/agents/toolReranker.test.js
+++ b/server/__tests__/utils/agents/toolReranker.test.js
@@ -1,0 +1,85 @@
+const mockRerank = jest.fn();
+
+jest.mock("../../../utils/EmbeddingRerankers/native", () => ({
+  NativeEmbeddingReranker: jest.fn().mockImplementation(() => ({
+    initClient: jest.fn().mockResolvedValue(),
+    rerank: mockRerank,
+  })),
+}));
+
+const {
+  ToolReranker,
+} = require("../../../utils/agents/aibitat/utils/toolReranker");
+
+function makeTools(count) {
+  return Array.from({ length: count }, (_, i) => ({
+    name: `tool_${i}`,
+    description: `does thing ${i}`,
+  }));
+}
+
+describe("ToolReranker.rerank", () => {
+  beforeEach(() => {
+    ToolReranker.instance = null;
+    mockRerank.mockReset();
+    mockRerank.mockImplementation(async (_query, documents, { topK }) =>
+      documents
+        .slice(0, topK)
+        .map((doc, i) => ({ ...doc, rerank_corpus_id: i, rerank_score: 1 }))
+    );
+    jest.spyOn(console, "log").mockImplementation(() => {});
+  });
+
+  afterEach(() => jest.restoreAllMocks());
+
+  it("truncates the query so injected document context cannot reach the cross-encoder", async () => {
+    // Agent turns append the full text of every attached file onto the user
+    // message. Passing that through verbatim blows up the (batch x sequence^2)
+    // cost of the cross-encoder and crashes the process.
+    const injectedContext = "lorem ipsum dolor sit amet ".repeat(4000);
+    const userPrompt = `summarize the attached file\n\n<attached_documents>\n${injectedContext}\n</attached_documents>`;
+    expect(userPrompt.length).toBeGreaterThan(100_000);
+
+    await new ToolReranker().rerank(userPrompt, makeTools(31));
+
+    const [queryUsed] = mockRerank.mock.calls[0];
+    expect(queryUsed.length).toBeLessThanOrEqual(1000);
+    expect(queryUsed.startsWith("summarize the attached file")).toBe(true);
+  });
+
+  it("leaves a normal length prompt untouched", async () => {
+    const userPrompt = "search the web for today's weather";
+
+    await new ToolReranker().rerank(userPrompt, makeTools(31));
+
+    expect(mockRerank.mock.calls[0][0]).toBe(userPrompt);
+  });
+
+  it("hands the full tool set to the reranker in a single call", async () => {
+    await new ToolReranker().rerank("find me a chart", makeTools(60));
+
+    expect(mockRerank).toHaveBeenCalledTimes(1);
+    expect(mockRerank.mock.calls[0][1]).toHaveLength(60);
+  });
+
+  it("maps reranked corpus ids back to the original tools", async () => {
+    const tools = makeTools(20);
+    mockRerank.mockImplementation(async () => [
+      { rerank_corpus_id: 7, rerank_score: 0.9 },
+      { rerank_corpus_id: 2, rerank_score: 0.8 },
+    ]);
+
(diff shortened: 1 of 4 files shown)


timothycarambat:
Putting numbers to this:
## Benchmark: Unbatched (pre-PR) vs Batched (post-PR) Reranking

**Environment:** Docker `--cpus=2 --memory=4g` (simulating 2 vCPU / 4GB VMs), `linux arm64`, Node 20, model `Xenova/ms-marco-MiniLM-L-6-v2`

### Single Call

| Docs | Unbatched Time | Unbatched RSS Δ | Batched Time | Batched RSS Δ | Peak RSS |
|------|---------------|-----------------|-------------|---------------|----------|
| 25   | 486ms         | +10.5 MB        | 403ms       | +0.0 MB       | 221.4 MB |
| 50   | 685ms         | +16.0 MB        | 1086ms      | +0.0 MB       | 237.4 MB |
| 100  | 1099ms        | +52.0 MB        | 2003ms      | +3.9 MB       | 293.5 MB |
| 200  | 2216ms        | +60.9 MB        | 4095ms      | +0.0 MB       | 354.4 MB |

### Concurrent (3 simultaneous calls, 50 docs each)

| Mode       | Time   | RSS Δ  | Peak RSS |
|------------|--------|--------|----------|
| Unbatched  | 2104ms | +0.1 MB | 354.5 MB |
| Batched    | 2692ms | +0.0 MB | 354.7 MB |

> **Takeaway:** Batched reranking eliminates the permanent RSS growth from ONNX (16-60 MB per unbatched call, never released to the OS) at the cost of ~1.5x wall-clock time on constrained hardware. At the production cap of 50 docs, that is ~400ms additional latency per query vs. avoiding the cumulative memory pressure can crash a VM

PR #6157, opened 2026-08-21
fix: [BUG]: Leaving chat while model is thinking makes the chat go away.

### LINKED ISSUE #6156: [BUG]: Leaving chat while model is thinking makes the chat go away.

AnythingLLM desktop app

### What happened?

If I leave the chat window while the model is thinking the chat vanishes and there's no way to get back to it. Checking in LM Studio it also looks like the model has stopped working.

Start a chat with a thinking model, go to another tab or screen, then go back.

LM Studio Bionic, any thinking model

_No response_

Roshan931:
Fixes #6156

## Root cause

Aborted stream during reasoning phase discards buffered thinking text, so no chat record is persisted

## Changes

- In handleDefaultStreamResponseV2, flush the buffered reasoning text into fullText (closing the <think> block) when the client aborts or when the stream finishes having produced reasoning only, so partial reasoning-only responses are persisted and the chat survives the disconnect. Added a regression test for the stream handler.

### DIFF
--- a/server/utils/helpers/chat/responses.js
+++ b/server/utils/helpers/chat/responses.js
@@ -51,13 +51,20 @@ function handleDefaultStreamResponseV2(response, stream, responseProps) {
     let fullText = "";
     let reasoningText = "";
 
+    function flushPendingReasoningToFullText() {
+      if (!reasoningText) return fullText;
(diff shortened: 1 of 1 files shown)


timothycarambat:
superceded by #6190

PR #6164, opened 2026-08-23
Fix broken relative links in READMEs

simpleqt:
### What does this PR do?

Fixes broken markdown links in the main README and the four locale READMEs (fa-IR, ja-JP, tr-TR, zh-CN). All of these 404 today when clicked on GitHub:

- **Root-relative links**: links written with a leading slash, e.g. `/server/storage/models/README.md` or `/images/youtube.png`, resolve against `github.com` instead of the repo, so they 404. Converted to proper relative paths (`./server/...` from the root README, `../server/...` from `locales/`). Targets verified to exist, including the `#text-generation-llm-selection` anchor.
- **`../embed/README.md` links** (fa-IR, ja-JP): the `embed/` directory is empty in the repo, so the README is gone. The embed widget now lives in [Mintplex-Labs/anythingllm-embed](https://github.com/Mintplex-Labs/anythingllm-embed); the main README already links there, so the locale READMEs now use the same target.
- **Dead localized anchors** (fa-IR, tr-TR): the feature list linked the English anchor `#supported-llms-embedder-models-speech-models-and-vector-databases`, but those files translate the heading. Updated to the slug of the actual localized heading (verified with the GitHub anchor algorithm).

### What it doesn't touch

Two other dead links exist (`docker/HOW_TO_USE_DOCKER.md` → deleted `ollama/README.md`, `open-computer/README.md` → deleted `BUILD-QEMU.md`); their intended replacement isn't obvious from the current tree, so I left them out of this PR.

### Review checklist

- [x] Docs / README only — no functional changes
- [x] Every changed link target verified to exist

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,7 @@
 <a name="readme-top"></a>
 
 > [!NOTE]
-> We are also working on [Open Computer](/open-computer) which gives an entire computer environment for AI Agents to use.
+> We are also working on [Open Computer](./open-computer) which gives an entire computer environment for AI Agents to use.
 >
 > This will bring AnythingLLM's agent capabilities to a new level and a novel UX paradigm for AI Agent use.
 >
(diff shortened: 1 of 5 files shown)


README.md
timothycarambat: This file can be unchecked - none of these links are actually broken and resolve properly

timothycarambat:
None of these links are broken on GitHub and resolve as expected - images and links all go to the right place practically speaking

simpleqt:
Fair enough — thanks for checking. The audit tool I used flags root-relative links that some renderers (editors, offline previews) don't resolve, but you're right that they work as expected where it matters on GitHub. Won't pursue this pattern further here.

PR #6165, opened 2026-08-23
fix: four server-side bugs (privatemode provider, prompt-history delete, success flag, event log)

simpleqt:
### What does this PR do?

Four small, independently verifiable server-side bug fixes found during a code review pass. No behavior changes beyond the fixes.

**1. `getLLMProviderClass` returns `undefined` for the `privatemode` provider** — `server/utils/helpers/index.js`

`getLLMProviderClass` destructures `PrivateModeLLM`, but `server/utils/AiProviders/privatemode/index.js` exports `PrivatemodeLLM` (capital M only), so the class resolves to `undefined`. `getLLMProvider` in the same file already uses the correct name. Impact: `Workspace._getContextWindow` returns `null` for privatemode workspaces and the aibitat `contextLimit` falls back to a hard-coded 8000.

**2. Single prompt-history delete silently no-ops** — `server/endpoints/workspaces.js`

`DELETE /workspace/prompt-history/:id` runs the `validWorkspaceSlug` middleware, but the route has no `:slug` param. The middleware executes `Workspace.get({ slug: undefined })`; Prisma ignores the undefined filter and returns an arbitrary workspace, so the delete is scoped to that workspace's id. For any other workspace `deleteMany` matches zero rows while the endpoint still returns `success: true` — the UI removes the entry and it reappears on reload. Fix: drop the middleware and the bogus scoping, delete by primary key. The `admin`/`manager` role gate is unchanged. (Sibling routes GET/DELETE `/workspace/:slug/prompt-history` are unaffected.)

**3. Error path reports `success: true`** — `server/endpoints/workspaces.js`

The catch block of the suggested-messages POST returns a 500 whose body says `success: true`. Aligned with `success: false` as used everywhere else on error paths.

**4. `document_embedded` event log always shows `documentName: "unknown"`** — `server/endpoints/workspacesParsedFiles.js`

The handler logs `document?.name`, but `Document.get` returns a `workspace_documents` row whose column is `filename` (`server/prisma/schema.prisma`) — `name` does not exist, so the fallback hit every time. Now logs `filename`.

### How to verify

1. Set `LLM_PROVIDER=privatemode` and inspect `GET /workspace/:slug` → `contextWindow` no longer `null`.
2. Create two workspaces, generate prompt history in the second, delete one entry from settings → it stays deleted after reload (previously reappeared).
3. Force an error in suggested-messages save → response body now `success: false`.
4. Embed a parsed file → admin event log shows the real document name.

### Review checklist

- [x] Bug fixes only, each verified against the code paths involved
- [x] `node --check` passes on all three modified files


simpleqt:
Answering the question from the #6168 review thread: yes, this one is a real bug. `server/utils/AiProviders/privatemode/index.js` exports the class as `PrivatemodeLLM` (and `getLLMProvider` in the same helpers file already destructures that exact name at line 237), but `getLLMProviderClass` destructures `PrivateModeLLM` — a key that does not exist — so the `case "privatemode"` branch always returned `undefined` and any caller resolving that provider's class broke. The fix just aligns the name. Happy to split the four fixes into separate PRs if you prefer reviewing them one at a time.

PR #6166, opened 2026-08-23
fix: require auth on env-dump endpoints

simpleqt:
Adds missing authentication to the two env-dump routes, found during a security review (per your SECURITY.md, low-severity issues with an obvious fix are welcome as public PRs).

1. GET /v1/system/env-dump (server/endpoints/api/system/index.js) had no [validApiKey] middleware, even though its own swagger doc declares a 403 InvalidAPIKey response and every sibling route in the file is gated.
2. GET /env-dump (server/endpoints/system.js) had no middleware at all.

Both call dumpENV(), which in production rewrites server/.env to the frozen protectedKeys subset - an unauthenticated, repeatable config-integrity action that can silently drop hand-added env keys on restart. The console route now requires validatedRequest + admin/manager role, matching other privileged system routes in the same file.

Verify: default docker deployment, curl -i http://localhost:3001/api/env-dump returns 200 before and 401/403 after; curl -i http://localhost:3001/api/v1/system/env-dump without an API key returns 403 after.

Two-line change, no behavior change for authorized callers.

### DIFF
--- a/server/endpoints/api/system/index.js
+++ b/server/endpoints/api/system/index.js
@@ -13,7 +13,7 @@ const { validApiKey } = require("../../../utils/middleware/validApiKey");
 function apiSystemEndpoints(app) {
   if (!app) return;
 
-  app.get("/v1/system/env-dump", async (_, response) => {
+  app.get("/v1/system/env-dump", [validApiKey], async (_, response) => {
     /*
    #swagger.tags = ['System Settings']
    #swagger.description = 'Dump all settings to file storage'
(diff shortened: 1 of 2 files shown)


timothycarambat:
There is a reason this exists and this is not a valid concern

simpleqt:
Understood, thanks for taking a look. I couldn't find a caller in this repo, which is why it read as an oversight — but if the unauthenticated route is intentional for a flow that runs before credentials exist, gating it would break that. Noted for future reference; the other PRs (#6164/#6165/#6168) are unaffected by this.

PR #6171, opened 2026-08-24
feat: add LanceDB Cloud as a separate vector database provider

### LINKED ISSUE #3940: [FEAT]: Support for External LanceDB Instances in AnythingLLM

**Description:**
Currently, AnythingLLM appears to rely on a local instance of LanceDB for vector storage. I would like to request support for connecting to an external/self-hosted LanceDB instance.

**Use Case:**
This would be especially useful for users or organizations who:

Want to centralize vector storage across multiple AnythingLLM deployments

Run LanceDB in a scalable, cloud-hosted environment (e.g., on a dedicated server or as a containerized service)

Need more control over storage, backups, or performance optimization

Require integration with existing infrastructure that already uses LanceDB

Suggested Implementation:

Add configuration options in the .env file or settings UI to define a remote LanceDB connection (host, port, credentials, etc.)

Ensure that the AnythingLLM backend uses this connection instead of initializing a local instance

**Benefits:**

Increased flexibility for deployment

Better separation of concerns between LLM logic and vector DB

Aligns with enterprise-grade use of AnythingLLM in scalable architectures

**References:**

LanceDB documentation

Possible inspiration from how AnythingLLM supports other remote vector DBs like Pinecone or Weaviate

Thanks for considering this feature!

ArslanRasheed60:
resolves #3940

## What
Per review feedback, this no longer touches the on-device `lancedb` provider at all. LanceDB Cloud is now its own vector database provider, `lancedb_cloud`, in the same shape as Chroma Cloud: a class that extends the local provider and overrides only what differs.

`server/utils/vectorDbProviders/lance/` is byte-identical to `master`.

## Implementation
- **Provider** (`server/utils/vectorDbProviders/lancedb_cloud/index.js`): `LanceDbCloud extends LanceDb`. It overrides `uri` (from `LANCEDB_CLOUD_URI`), `connect()` (forwards `apiKey`/`region` for `db://`, and keys its cached connection on URI + credentials so a settings change reconnects without a restart), and `reset()` (the parent removes its folder from disk, which is meaningless or destructive for an external store, so this drops every table instead). Everything else is inherited.
- **Registration**: `getVectorDbClass()` case, `supportedVectorDB()`, `systemSettings.currentSettings`, `ProviderPrivacy`, and the vector DB provider list in Settings, mirroring `chromacloud` one-for-one.
- **ENV** (`server/utils/helpers/updateENV.js`): `LANCEDB_CLOUD_URI` (required, validated) plus optional `LANCEDB_CLOUD_API_KEY` / `LANCEDB_CLOUD_REGION`. Object-store credentials (`AWS_ACCESS_KEY_ID` etc.) are read from the process environment by LanceDB itself, so they are documented as container-level env rather than new AnythingLLM keys.
- **Reset on switch**: nothing new. Because this is a distinct `VECTOR_DB` value, the existing `handleVectorStoreReset` postUpdate hook fires on `lancedb <-> lancedb_cloud` and on any embedder change while on it. Note the ChromaCloud-style `if (process.env.VECTOR_DB !== ...) throw` guard is deliberately *not* mirrored here: `updateENV` assigns the new `VECTOR_DB` before running `postUpdate`, so that guard would stop the outgoing provider from cleaning up its own namespaces.

## Testing
`npx jest server/__tests__` - **39 suites / 520 tests pass**.

- `server/__tests__/utils/vectorDbProviders/lancedb_cloud/index.test.js` - inheritance, URI resolution, connect with/without each credential, connection cache reuse and re-dial, `reset()` dropping tables instead of touching the filesystem, and a case asserting the on-device provider still connects to `STORAGE_DIR/lancedb` with a single argument while all three cloud vars are set.
- `server/__tests__/utils/helpers/updateENV.test.js` - URI validation through `updateENV`, and reset-on-switch: `lancedb -> lancedb_cloud` resets `lancedb`, the reverse resets `lancedb_cloud`, an embedder or embedding-model change while on `lancedb_cloud` resets `lancedb_cloud`, and an unchanged provider resets nothing.

End-to-end on a running instance (server + collector + frontend):
- Local LanceDB unchanged - pane still shows only the "no configuration needed" line, embeds into `server/storage/lancedb`.
- "LanceDB Cloud" appears as its own entry in the provider list next to LanceDB, with URI / API Key / Region fields.
- Switching to it shows the reset warning, logs `Vector configuration changed from lancedb to lancedb_cloud - resetting lancedb namespaces`, empties the local store, and re-embeds into the external URI while the local directory stays empty. Switching back drops the external tables. Changing the embedder while on `lancedb_cloud` resets it too.
- Invalid URIs are rejected with the validation message.
- `s3://` against MinIO: writes land in the bucket.


ArslanRasheed60:
@timothycarambat when you have a moment - this one is scoped to keep the default LanceDB path byte-identical (single-argument `connect()`, no options object) so existing installs and the desktop app are unaffected. Verified end-to-end against a real S3 store in the docker image as well as the local default. Happy to adjust the approach if you'd rather this stay env-only without the settings fields.

timothycarambat (changes requested):
We should leave the regular `lancedb` untouched totally and put this under `lancedb_cloud` instead as a separate provider who can inherit and override the regular lancedb methods.

The reason is that the default lanceDB **must** be on device - it should not be messed with at all and we should enforce using lancedb cloud as a new provider totally as this will force an embedder/db wipe across the system which is required when swapping providers anyway.

This would cover the current gap in this pr where swapping from local embedder to cloud endpoint does not trigger a vectordb reset and it should since the vectordbs are now different. 

There is also an implicit scope confusion introducing this as the main default as well since only empty values seem to imply "use on device vs cloud" and it would be more clear someone intended to use lanceDB on S3 vs lanceDB on device.

This is a welcome change for sure, it just needs to be its own provider instead of crossing scope with the on device provider. We did the same for Chroma/Chroma Cloud.

ArslanRasheed60:
@timothycarambat reworked along the lines you asked for.

`lancedb` is completely untouched now - `server/utils/vectorDbProviders/lance/` is byte-identical to `master`. LanceDB Cloud is its own `lancedb_cloud` provider that extends it and overrides only `uri`, `connect()` and `reset()`, same shape as Chroma/Chroma Cloud, and it shows up as its own entry in the provider list.

Because it's a distinct `VECTOR_DB` value, the existing `handleVectorStoreReset` hook covers the gap you pointed out: switching either direction, and changing the embedder while on it, all trigger the reset. Unit tests cover that plus the provider itself (520 tests green), and I verified the switch/reset behaviour live in the app.

Ready for another look when you have a minute.

ArslanRasheed60:
@timothycarambat rebased onto latest master — 47 suites / 645 tests green.

Worth flagging: the three workflows on this PR are sitting in `action_required`, so CI has never actually run against it. It needs an approval click on your side.

Recap of the rework since your review, in case it got lost: `server/utils/vectorDbProviders/lance/` is byte-identical to master now. LanceDB Cloud is its own `lancedb_cloud` provider that extends it and overrides only `uri`, `connect()` and `reset()` — same shape as Chroma/Chroma Cloud. Since it's a distinct `VECTOR_DB` value, the existing `handleVectorStoreReset` hook covers the swap gap you pointed out.

Your review is still anchored to the pre-rework commit, so the PR reads as changes-requested. Happy to keep adjusting if anything's still off.

PR #2632, opened 2024-11-14
feat: support for iam roles for bedrock client

dannysteenman:
### Pull Request Type

- [x] ✨ feat
- [x] 🐛 fix
- [ ] ♻️ refactor
- [ ] 💄 style
- [ ] 🔨 chore
- [ ] 📝 docs

### Relevant Issues

resolves https://github.com/Mintplex-Labs/anything-llm/issues/2588

### What is in this change?

This feature includes the ability to authenticate with bedrock on aws using your iam role instead of using long term iam credentials. It should be backwards compatible with people who have setup: 

```
AWS_BEDROCK_LLM_ACCESS_KEY_ID=
AWS_BEDROCK_LLM_ACCESS_KEY=
```

### Additional Information

In the UI it looks like this now, default:

SCR-20241114-jaaq

when you enable the iam role:

SCR-20241114-jacj

Please verify it and let me know if you have some feedback, so I can improve it, then I can also update the docs once you're happy with the result.

### Developer Validations

- [ ] I ran `yarn lint` from the root of the repo & committed changes
- [ ] Relevant documentation has been updated
- [ x ] I have tested my code functionality
- [ x] Docker build succeeds locally

PR #6173, opened 2026-08-24
feat: Entra ID / managed identity auth for Azure OpenAI (Azure AI Foundry)

### LINKED ISSUE #6172: [FEAT]: Entra ID / managed identity auth for Azure OpenAI (Azure AI Foundry)

**Support Microsoft Entra ID (managed identity) authentication for the Azure OpenAI / Azure AI Foundry provider** — the Azure equivalent of what #2632 did for AWS Bedrock with IAM roles.

#### Problem

Today the Azure OpenAI provider requires a static API key everywhere:

- `server/utils/AiProviders/azureOpenAi/index.js:15` — throws without `AZURE_OPENAI_KEY`
- `server/utils/EmbeddingEngines/azureOpenAi/index.js:8` — same
- `server/utils/agents/aibitat/providers/azure.js:16` and `server/utils/agents/aibitat/providers/ai-provider.js:303-312` — agent clients
- `server/utils/agents/index.js:155` — agent preflight check
- `server/utils/helpers/updateENV.js:33` — `AzureOpenAiKey` is validated with `isNotEmpty`
- `frontend/src/components/LLMSelection/AzureAiOptions/index.jsx:33` and `frontend/src/components/EmbeddingSelection/AzureAiOptions/index.jsx:27` — `required` inputs

That's a problem when AnythingLLM runs *on* Azure (Container Apps, App Service, AKS with workload identity, or a VM):

1. It forces a long-lived secret to be stored and rotated by hand, when the platform can hand out short-lived tokens automatically.
2. Many organizations set `disableLocalAuth: true` on their Azure AI Foundry / Azure OpenAI resource, so key auth is **not available at all** — the provider is simply unusable there.

#### Precedent

#2632 solved the same problem for Bedrock by adding a connection-method switch: `AWS_BEDROCK_LLM_CONNECTION_METHOD=iam_role` skips the explicit-credential requirement and lets the AWS SDK's default credential chain resolve credentials from the environment. The Azure side has a direct analogue in `@azure/identity` (`DefaultAzureCredential` / `ManagedIdentityCredential`).

#### Proposal

Add an auth-method switch alongside the existing key:

| Env var | Values | Notes |
| --- | --- | --- |
| `AZURE_OPENAI_CONNECTION_METHOD` | `api_key` (default) \| `managed_identity` | `api_key` keeps today's behaviour exactly |
| `AZURE_OPENAI_MANAGED_IDENTITY_CLIENT_ID` | optional | client ID of a user-assigned managed identity; omit for system-assigned |

When the method is `managed_identity`, `AZURE_OPENAI_KEY` is no longer required and the clients get a bearer token for the `https://cognitiveservices.azure.com/.default` scope instead.

The LLM path currently uses the base `OpenAI` client against Azure's v1 surface (`/openai/v1`, see `AzureOpenAiLLM.formatBaseUrl`), which sends `Authorization: Bearer <apiKey>` — exactly the header Entra ID auth needs. Since Entra tokens expire (~1h) while `apiKey` is a fixed string, the token has to come from a provider on each request rather than being baked in at construction:

```js
// server/utils/AiProviders/azureOpenAi/credentials.js (new)
const { DefaultAzureCredential, getBearerTokenProvider } = require("@azure/identity");
const AZURE_COGNITIVE_SCOPE = "https://cognitiveservices.azure.com/.default";

function azureAuth() {
  if (process.env.AZURE_OPENAI_CONNECTION_METHOD !== "managed_identity")
    return { apiKey: process.env.AZURE_OPENAI_KEY };

  const credential = new DefaultAzureCredential({
    managedIdentityClientId: process.env.AZURE_OPENAI_MANAGED_IDENTITY_CLIENT_ID || undefined,
  });
  // getBearerTokenProvider caches the token and refreshes it before expiry
  const getToken = getBearerTokenProvider(credential, AZURE_COGNITIVE_SCOPE);
  return {
    apiKey: "managed-identity", // placeholder; the fetch wrapper below sets the real header
    fetch: async (url, init = {}) => {
      const headers = new Headers(init.headers);
      headers.set("Authorization", `Bearer ${await getToken()}`);
      return fetch(url, { ...init, headers });
    },
  };
}
```

The embedder uses the SDK's `AzureOpenAI` class, which takes `azureADTokenProvider` directly, so it can use the same credential without the `fetch` wrapper.

Full set of touchpoints:

- new `@azure/identity` dependency in `server/package.json`
- shared credential helper used by the LLM, the embedder, `providers/azure.js`, and the LangChain client in `ai-provider.js`
- relax the key requirement in the four constructors and in the agent preflight (`server/utils/agents/index.js:155`)
- `updateENV.js`: register the two new keys, and make the `AzureOpenAiKey` `isNotEmpty` check conditional on the connection method
- `systemSettings.js:879-885`: expose `AzureOpenAiConnectionMethod` so the UI can render the right fields
- both `AzureAiOptions` components: a connection-method selector that hides/unrequires the API key field when managed identity is selected
- `docker/.env.example:22-24` and `:196-197`: document the new vars

#### Backwards compatibility

`AZURE_OPENAI_CONNECTION_METHOD` defaults to `api_key`, so existing installs are untouched — same as the `iam` default in the Bedrock change.

#### Notes / questions for maintainers

- `DefaultAzureCredential` also covers `az login` and workload identity, which makes local development and AKS work without extra code. Would you prefer a single `managed_identity` value backed by `DefaultAzureCredential`, or two explicit values (`managed_identity` → `ManagedIdentityCredential`, `default_credential` → `DefaultAzureCredential`)?
- `@azure/identity` is a new runtime dependency for the server (~ the same weight as the AWS SDK already pulled in for Bedrock). Any objection?
- Should the same switch apply to the Azure embedder in the same PR, or would you rather keep the PR to the LLM provider only?

#### Offer

I'm happy to implement this — I'd fork the repo, build it on a branch, and open a PR following `CONTRIBUTING.md` (PR against `master`). Just let me know your preference on the two naming questions above before I start, so the PR doesn't need a redesign in review.

LeTuR:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6172

### Description

The Azure OpenAI (Azure AI Foundry) provider can only authenticate with a static API key today. That forces a long-lived secret onto installs running on Azure, and it makes the provider **unusable entirely** against a resource configured with `disableLocalAuth: true`, which a lot of organizations require.

This adds Microsoft Entra ID / managed identity as an alternative — the Azure counterpart to the IAM-role option #2632 once added for Bedrock.

**New settings**

| Env var | Values | Notes |
| --- | --- | --- |
| `AZURE_OPENAI_CONNECTION_METHOD` | `api_key` (default) \| `managed_identity` | The default preserves today's behavior exactly |
| `AZURE_OPENAI_MANAGED_IDENTITY_CLIENT_ID` | optional | Client ID of a user-assigned managed identity; leave empty for the system-assigned one |

With `managed_identity`, `AZURE_OPENAI_KEY` is no longer required and every Azure client authenticates with a token for the `https://cognitiveservices.azure.com/.default` scope from `DefaultAzureCredential`.

**How the token gets onto the request**

`server/utils/AiProviders/azureOpenAi/credentials.js` is a small shared module that all four Azure client construction sites now go through:

- The LLM, the agent provider and the LangChain client use the plain `OpenAI` client against Azure's v1 surface, which already sends `Authorization: Bearer <apiKey>` — the exact header Entra needs. But an Entra token expires while `apiKey` is fixed at construction, so these get a `fetch` wrapper that sets the header from the token provider on every request. `getBearerTokenProvider` caches the token and refreshes it ahead of expiry, and the credential itself is built once per process (rebuilt only if the configured client ID changes).
- The embedder uses the SDK's `AzureOpenAI` class, which accepts `azureADTokenProvider` directly and handles refresh itself.

`DefaultAzureCredential` also resolves environment credentials, AKS workload identity and a developer's `az login` session, so the one setting works in production and locally without a second code path.

**Backwards compatibility**

`AZURE_OPENAI_CONNECTION_METHOD` defaults to `api_key`. Existing installs are untouched — the key stays required, and the "No Azure API key was set." error is unchanged in that mode.

**Files touched**

- new `server/utils/AiProviders/azureOpenAi/credentials.js` — the shared credential logic
- `server/utils/AiProviders/azureOpenAi/index.js`, `server/utils/EmbeddingEngines/azureOpenAi/index.js`, `server/utils/agents/aibitat/providers/azure.js`, `server/utils/agents/aibitat/providers/ai-provider.js` — build clients through it
- `server/utils/agents/index.js` — agent preflight no longer demands a key under managed identity
- `server/utils/helpers/updateENV.js`, `server/models/systemSettings.js` — register and expose the two new settings
- both `AzureAiOptions` components — a Connection Method selector that swaps the API Key field for an optional Managed Identity Client ID field
- `frontend/src/locales/en/common.js`, `docker/.env.example` — labels and documentation
- `server/package.json` / `server/yarn.lock` — adds `@azure/identity`

### Visuals (if applicable)

The LLM and embedder settings panes gain a **Connection Method** dropdown (`API Key` / `Managed Identity`) next to the Azure Service Endpoint. Selecting `Managed Identity` replaces the API Key input with an optional **Managed Identity Client ID** input; selecting `API Key` restores the original form unchanged.

### Additional Information

Two design questions from the issue that I resolved one way to keep the PR concrete — happy to change either:

1. **One `managed_identity` value backed by `DefaultAzureCredential`**, rather than separate `managed_identity` / `default_credential` values. It keeps the UI to a single choice and makes local development work without a second setting. Easy to split if you would rather be explicit.
2. **The embedder is included**, since it shares `AZURE_OPENAI_ENDPOINT` / `AZURE_OPENAI_KEY` with the LLM and would otherwise still demand a key against a resource that has none.

Also worth flagging: #2632's `AWS_BEDROCK_LLM_CONNECTION_METHOD` no longer exists — Bedrock was since refactored onto the Mantle API with a plain key. I cited it as the design precedent, not as live code to match.

`@azure/identity` is a new server dependency (~4.13.2). `server/yarn.lock` is updated with its subtree; the diff is additions only.

### Developer Validations

On the last two, so you know exactly what has and has not been exercised:

- I added `server/__tests__/utils/AiProviders/azureOpenAi/credentials.test.js` (11 tests, passing) covering both connection methods, the bearer header being set per request, caller headers/`signal`/body surviving the fetch wrapper, system- vs user-assigned identity selection, and credential rebuild when the client ID changes. The pre-existing `azureOpenAiModelPref` suite still passes.
- **I have not tested this end to end against a live Azure resource with a managed identity**, and I have not run a Docker build. If that verification is a merge requirement, tell me and I will run it against a real deployment before you spend review time on this.
- On lint: I verified Prettier formatting over every changed file rather than running the full `yarn lint`, which needs a complete workspace install.

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -20,7 +20,9 @@ GID='1000'
 
 # LLM_PROVIDER='azure'
 # AZURE_OPENAI_ENDPOINT=
-# AZURE_OPENAI_KEY=
+# AZURE_OPENAI_CONNECTION_METHOD='api_key' # 'api_key' (default) or 'managed_identity' to authenticate with Microsoft Entra ID.
+# AZURE_OPENAI_KEY= # Not needed when using 'managed_identity'.
+# AZURE_OPENAI_MANAGED_IDENTITY_CLIENT_ID= # Only for a user-assigned managed identity - leave empty for the system-assigned one.
 # AZURE_OPENAI_MODEL_PREF='my-gpt35-deployment' # This is the "deployment" on Azure you want to use. Not the base model.
 # EMBEDDING_MODEL_PREF='embedder-model' # This is the "deployment" on Azure you want to use for embeddings. Not the base model. Valid base model is text-embedding-ada-002
 
@@ -194,7 +196,9 @@ GID='1000'
 
 # EMBEDDING_ENGINE='azure'
 # AZURE_OPENAI_ENDPOINT=
-# AZURE_OPENAI_KEY=
+# AZURE_OPENAI_CONNECTION_METHOD='api_key' # 'api_key' (default) or 'managed_identity' to authenticate with Microsoft Entra ID.
+# AZURE_OPENAI_KEY= # Not needed when using 'managed_identity'.
+# AZURE_OPENAI_MANAGED_IDENTITY_CLIENT_ID= # Only for a user-assigned managed identity - leave empty for the system-assigned one.
 # EMBEDDING_MODEL_PREF='my-embedder-model' # This is the "deployment" on Azure you want to use for embeddings. Not the base model. Valid base model is text-embedding-ada-002
 
 # EMBEDDING_ENGINE='localai'
--- a/frontend/src/components/EmbeddingSelection/AzureAiOptions/index.jsx
+++ b/frontend/src/components/EmbeddingSelection/AzureAiOptions/index.jsx
@@ -1,4 +1,11 @@
+import { useState } from "react";
+
 export default function AzureAiOptions({ settings }) {
+  const [connectionMethod, setConnectionMethod] = useState(
+    settings?.AzureOpenAiConnectionMethod || "api_key"
+  );
+  const usesManagedIdentity = connectionMethod === "managed_identity";
+
   return (
     <div className="w-full flex flex-col gap-y-4">
       <div className="w-full flex items-center gap-[36px] mt-1.5">
@@ -20,20 +27,56 @@ export default function AzureAiOptions({ settings }) {
 
         <div className="flex flex-col w-60">
           <label className="text-white text-sm font-semibold block mb-3">
-            API Key
+            Connection Method
           </label>
-          <input
-            type="password"
-            name="AzureOpenAiKey"
+          <select
+            name="AzureOpenAiConnectionMethod"
+            value={connectionMethod}
+            onChange={(e) => setConnectionMethod(e.target.value)}
             className="border-none bg-theme-settings-input-bg text-white placeholder:text-theme-settings-input-placeholder text-sm rounded-lg focus:outline-primary-button active:outline-primary-button outline-none block w-full p-2.5"
-            placeholder="Azure OpenAI API Key"
-            defaultValue={settings?.AzureOpenAiKey ? "*".repeat(20) : ""}
             required={true}
-            autoComplete="off"
-            spellCheck={false}
-          />
+          >
+            <option value="api_key">API Key</option>
+            <option value="managed_identity">Managed Identity</option>
+          </select>
         </div>
 
+        {usesManagedIdentity ? (
+          <div className="flex flex-col w-60">
+            <label className="text-white text-sm font-semibold block mb-3">
+              Managed Identity Client ID
+            </label>
+            <input
+              type="text"
+              name="AzureOpenAiManagedIdentityClientId"
+              className="border-none bg-theme-settings-input-bg text-white placeholder:text-theme-settings-input-placeholder text-sm rounded-lg focus:outline-primary-button active:outline-primary-button outline-none block w-full p-2.5"
+              placeholder="Optional - only for a user-assigned identity"
+              defaultValue={settings?.AzureOpenAiManagedIdentityClientId}
+              required={false}
+              autoComplete="off"
+              spellCheck={false}
+            />
+          </div>
+        ) : (
+          <div className="flex flex-col w-60">
+            <label className="text-white text-sm font-semibold block mb-3">
+              API Key
+            </label>
+            <input
+              type="password"
+              name="AzureOpenAiKey"
+              className="border-none bg-theme-settings-input-bg text-white placeholder:text-theme-settings-input-placeholder text-sm rounded-lg focus:outline-primary-button active:outline-primary-button outline-none block w-full p-2.5"
+              placeholder="Azure OpenAI API Key"
+              defaultValue={settings?.AzureOpenAiKey ? "*".repeat(20) : ""}
+              required={true}
+              autoComplete="off"
+              spellCheck={false}
+            />
+          </div>
+        )}
+      </div>
+
+      <div className="w-full flex items-center gap-[36px]">
         <div className="flex flex-col w-60">
           <label className="text-white text-sm font-semibold block mb-3">
             Embedding Deployment Name
--- a/frontend/src/components/LLMSelection/AzureAiOptions/index.jsx
+++ b/frontend/src/components/LLMSelection/AzureAiOptions/index.jsx
@@ -1,9 +1,14 @@
 import { Info } from "@phosphor-icons/react";
+import { useState } from "react";
 import { useTranslation } from "react-i18next";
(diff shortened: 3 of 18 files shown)


LeTuR:
Added deployment selection for this provider, since testing the managed-identity
path surfaced how easy the free-text field is to get wrong: a connection label
pasted where the deployment name belongs comes back only as `404 The API
deployment for this resource does not exist`, with nothing in the UI to check it
against.

**Chat Deployment Name** is now a dropdown populated from the resource itself:

azure-model-picker

Notes on the approach:

- The data-plane `models` route is not usable for this. It advertises the whole
  regional catalog — 414 entries against my test resource — almost none of which
  the resource can serve, so the dropdown would be full of models that 404. The
  `deployments` route reflects the resource, and it is only served on
  api-version `2023-03-15-preview`; the current versions dropped it.
- Deployments are classified against the capability catalog so embedding
  deployments stay out of the chat dropdown. A deployment whose model the catalog
  does not list is **kept**, not hidden — a resource can serve models absent from
  it (a partner model published through AI Foundry, in my case `Kimi-K2.6`), and
  those work normally. They appear under a separate group.
- Listing authenticates with whatever the form currently holds, managed identity
  included, and deliberately does not touch the cached token provider the live
  clients use.
- The name can still be typed in. Listing needs read access to the resource, and
  Azure does not report a deployment created in the last few minutes, so the
  dropdown must not be the only way in. It also falls back to the text input
  automatically when listing fails.

Verified end to end against a real Azure AI Foundry resource over managed
identity with no API key set: the dropdown lists the resource's deployments, and
chat completions succeed against the selected one. `models.test.js` adds 12 tests
covering classification, the provisioning-status filter, both auth headers,
endpoint normalisation, and the catalog-unavailable and listing-failed paths.

One thing I did not change, flagged in case it is worth a separate issue: the
**Chat Model Token Limit** options stop at 200,000 and then jump to 1,047,576,
so there is no correct value for a model with a 922,000-token window. Neither
neighbour is right — one wastes most of the context, the other overstates it.

PR #6178, opened 2026-08-25
feat(localai): support multiple concurrent connections

### LINKED ISSUE #6162: [FEAT]: Allow multiple Local AIs

I can switch the model at any time, but the given context window is fixed in my settings, not configurable for each model on my Local AI endpoint.

On top of that, I host two endpoints for two different AI routers (CLIProxyAI and llama.cpp), but I cannot add both endpoints, I have to pick one.

Max-Reisinger:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6162

### Description

- add admin-managed, named LocalAI connections with an endpoint, optional encrypted API key, default model, and context window
- allow workspaces and the chat model selector to bind to a saved LocalAI connection while retaining the existing system LocalAI settings as the default
- allow model-router fallbacks and individual routing rules to target different LocalAI connections concurrently
- propagate the selected connection through chat, agent, and router provider resolution without mutating global environment settings
- prevent deletion while a connection is referenced and add regression coverage for validation, secret handling, and per-instance LocalAI isolation

## Screenshots

Multiple named LocalAI connections can be saved independently, each with its own base URL, model, API key, and context window.

A model router can target a specific saved LocalAI connection. In this example, prompts containing `code` are routed to `LocalAI GPU 2` using `qwen2.5-14b-instruct`.

exec-3e9d6e3a-82e6-46c7-bcae-7b343f765cde

exec-cfef17ae-d074-48a1-82f2-6f56f28672b6

### Additional Information

- Existing `LOCAL_AI_*` settings remain fully supported as the fallback when no saved connection is selected.
- Prisma schema validation, migration SQL execution against a temporary SQLite database, frontend lint, and the frontend production build pass locally.
- The feature-focused test suites pass (4 suites, 70 tests). The root test run reaches 504 passing tests, but the collector FFmpeg suite cannot run in this environment because it uses Node 24 and has no FFmpeg binary; the contribution guide requires Node 18.18.
- Root lint is currently blocked by five pre-existing Prettier errors in three untouched server files. Targeted lint for every changed server file and full frontend lint pass.

### Developer Validations

### DIFF
--- a/frontend/src/components/LLMSelection/LocalAiConnectionSelector/index.jsx
+++ b/frontend/src/components/LLMSelection/LocalAiConnectionSelector/index.jsx
@@ -1,0 +1,43 @@
+import { useEffect, useState } from "react";
+import LocalAiConnection from "@/models/localAiConnection";
+
+export default function LocalAiConnectionSelector({
+  name = "chatConnectionId",
+  value = "",
+  onChange,
+  onConnectionChange,
+  required = false,
+  className = "border-none bg-theme-settings-input-bg text-white text-sm rounded-lg block w-full p-2.5",
+}) {
+  const [connections, setConnections] = useState([]);
+
+  useEffect(() => {
+    LocalAiConnection.all().then(setConnections);
+  }, []);
+
+  function handleChange(event) {
+    const nextValue = event.target.value;
+    onChange?.(nextValue);
+    onConnectionChange?.(
+      connections.find((connection) => String(connection.id) === nextValue) ||
+        null
+    );
+  }
+
+  return (
+    <select
+      name={name}
+      value={value || ""}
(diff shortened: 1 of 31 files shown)


timothycarambat:
The issue linked is marked `core-team` only so we cannot merge this. The reason is because what is being proposed is very complex and this PR doesnt consider a lot of other conditions like pre-existing connections, migrations, and a bunch of other stuff that if merged would explode lots of peoples instances. 

This issue just is too critical path for non-core solution even if this specific change does work for you specifically. That being said you are always welcome to have it on a fork while we work on roadmapping how to get this functionality live.

PR #6181, opened 2026-08-25
Route Bedrock cross-region inference profiles via bedrock-runtime

### LINKED ISSUE #6179: [BUG]: AWS ISO Regions for Bedrock

All versions

### What happened?

Specialized Regions specifically ISO regions are not supported in the Bedrock UI configurations. Adding Regions that are not in the dropdown list via Helm install causes the UI to crash. 

Here is an example of specialized regions.
https://docs.okd.io/4.22/installing/installing_aws/ipi/installing-aws-specialized-region.html

Requesting that we add support for specialized regions.

Scenario 1: Add A specialized region via helm install - navigate to the model select for bedrock.
Scenario 2: There is no way to modify or add specialized regions via UI

Bedrock - any model

N/a

timothycarambat:
### Pull Request Type

- [x] ✨ feat (New feature)
- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6179

### Description

Mantle rejects cross-region inference profile IDs (eu./us./global.*), which are the only way to run Claude in many regions (all of EU). The bedrock-runtime endpoint serves the same Anthropic Messages API with the same bearer API key and accepts profile IDs, so Claude models are now routed by ID shape: geo-prefixed profile IDs go to bedrock-runtime, plain catalog IDs keep the Mantle path unchanged.

Model listing now merges the Mantle catalog with Anthropic inference profiles from the Bedrock control plane (same bearer auth), so profile IDs appear in the model dropdown even though Mantle's listing omits them.

The region selector gains a manual-entry fallback so regions outside the hardcoded list (GovCloud, specialized partitions set via ENV/Helm) no longer render an empty select that clobbers the saved region on save. Optional endpoint host override ENVs added for air-gapped partitions whose domains don't follow commercial naming.

### Developer Validations

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -110,6 +110,11 @@ GID='1000'
 # AWS_BEDROCK_LLM_MODEL_PREFERENCE=global.anthropic.claude-haiku-4-5-20251001-v1:0
 # AWS_BEDROCK_LLM_MODEL_TOKEN_LIMIT='128000'
 # AWS_BEDROCK_STREAMING_DISABLED=1
+# Optional endpoint host overrides for air-gapped/specialized AWS partitions
+# (eg: ISO/C2S) whose domains do not follow the commercial naming scheme.
+# AWS_BEDROCK_LLM_MANTLE_ENDPOINT='https://bedrock-mantle.us-west-2.api.aws'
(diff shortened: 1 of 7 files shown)


jessethorn13:
Thanks @timothycarambat I regularly install to air-gapped environments. This helps so much! I will follow up by the end of the week.

Issue #6142, opened 2026-08-18
[FEAT]: Token Estimation Improvements for better chunking


kihei088blue:
### How are you running AnythingLLM?
AnythingLLM desktop app

### AnythingLLM Version
v1.16.0

### LLM Provider
Generic OpenAI (local llama.cpp server, Qwen3.5-4B-UD-Q4_K_XL.gguf)
- Model context window: 32768 (matches `-c 32768` on the llama-server side)
- Max Tokens: 20480

### Embedder
Generic OpenAI (local llama.cpp server, bge-m3-q4_k_m.gguf, `--embedding`)

### What happened?

When attaching a document to the chat window (via drag-and-drop or the `+` icon, per the "Attach" flow described in the docs: https://docs.useanything.com/chatting-with-documents/introduction), the model only received approximately the **first 3.6k tokens** of the document, regardless of the document's actual size or the configured context window (32768). The remainder of the document was silently dropped — no warning, no truncation notice, and no context-overflow prompt was shown.

The token counter next to the `+` icon confirmed this: it consistently showed ~3.6-3.7k tokens used, even though the source file was ~78,000 characters / 853 lines / well within the configured 32768-token context window.

As a direct consequence, any question about content located past this cutoff point returned "not found in the document" even though the content was clearly present in the file.

I also observed a related, less consistent symptom with the workspace "Pin" feature (embed a document, then pin it to bypass RAG) — see "Note on Pin feature" below.

**I identified the root cause: half-width katakana characters** (Unicode range U+FF61–U+FF9F), commonly used in Japanese government/procurement documents for list markers like `(ｱ)`, `(ｲ)`, `(ｳ)`. See "Root cause identified" below for the confirming test.

### Expected behavior

Per the documentation, "By default, AnythingLLM will insert the full text of your documents into the chat window" when it fits within the model's context window. Since my document (~20-30k tokens estimated) is smaller than the configured 32768-token context window, I expected the full text to be included.

### Steps to Reproduce

1. Start a local llama.cpp server (`llama-server`) with `-c 32768` and a Generic OpenAI-compatible endpoint.
2. Configure AnythingLLM's LLM Preference: Generic OpenAI, BaseURL pointing to the local server, Model context window = 32768.
3. Start a **new** chat thread in a workspace.
[...]

PR #6183, opened 2026-08-26
fix: use available prompt window for document context

SidheshwarSarangal:
## Summary

- Replaces the fixed 15% system-prompt cap in `messageArrayCompressor` with a budget based on the remaining prompt-window capacity.
- Reserves space for the response, user prompt, and maximum permitted chat history.
- Allows document context to reuse system-instruction capacity that was not needed.
- Adds a focused regression test for large document context.

## Problem

When prompt compression was triggered, document context was limited to 75% of the system allowance. Since the system allowance was 15% of the model context window, document context could use only 11.25% of the available window—even when substantial capacity remained unused.

This caused large attached or pinned documents to be truncated much more aggressively than necessary.

## Implementation

The message-array compressor now calculates the system and document-context allowance from:

- The model's prompt-window limit
- The 600-token response buffer
- The current user-prompt size
- The maximum permitted chat-history allowance

The existing user-prompt and history priorities remain unchanged.

This change intentionally does not modify the tokenizer fallback behavior, which is the other concern discussed in the related issue.

## Validation

- [x] Added a regression test for dynamic document-context budgeting
- [x] Focused compressor test passes
- [x] Generic OpenAI provider tests pass
- [x] 10 relevant tests pass
- [x] Production file passes ESLint
- [x] Both changed files pass Prettier
- [x] No whitespace errors

Connects #6142

### DIFF
--- a/server/__tests__/utils/helpers/chat.test.js
+++ b/server/__tests__/utils/helpers/chat.test.js
@@ -1,0 +1,55 @@
+jest.mock("../../../utils/helpers/tiktoken", () => ({
+  TokenManager: jest.fn().mockImplementation(() => ({
+    countFromString: (input = "") => String(input).length,
+    statsFrom: (input) =>
+      input.reduce(
+        (total, message) => total + String(message.content).length,
+        0
+      ),
+    tokensFromString: (input = "") => [...String(input)],
+    bytesFromTokens: (tokens = []) => tokens.join(""),
+  })),
+}));
+
+const { messageArrayCompressor } = require("../../../utils/helpers/chat");
+
+describe("messageArrayCompressor", () => {
+  it("uses available prompt-window capacity for document context", async () => {
+    const llm = {
+      model: "test-model",
+      promptWindowLimit: () => 4_000,
+      limits: {
+        system: 600,
+        history: 600,
+        user: 2_800,
+      },
+    };
+
+    const context = "c".repeat(3_500);
+    const userPrompt = "u".repeat(200);
+
+    const result = await messageArrayCompressor(
+      llm,
+      [
+        {
+          role: "system",
+          content: `Follow the system instructions.\nContext:${context}`,
+        },
+        {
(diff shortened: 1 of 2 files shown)


SidheshwarSarangal:
Hi maintainers, just following up on this PR when you have a chance. It addresses the document-context budgeting concern from #6142 and includes a focused regression test. Please let me know if any changes are needed. Thanks.

SidheshwarSarangal:
Hi @timothycarambat, gentle ping when you have a chance. This PR addresses the document-context budgeting concern from #6142 and includes a regression test. I’d appreciate any feedback or requested changes.

timothycarambat:
Thanks for the PR, but I'm not going to merge this one.

This compressor is critical-path code that was tuned by back-testing, and the header comment states the design goal explicitly: a compressed prompt that fits "with a lot of room for response." This change moves every compressed prompt to within 600 tokens of the window limit. For the exact setup in #6142 (32k window, Max Tokens 20480 on llama.cpp) a ~27k compressed prompt plus the reply budget would be rejected or truncated by the server.

It also fills the window using the cl100k estimate that we already know is wrong for CJK text. Compression only fires for non-OpenAI models when that estimate is off, so removing the cushion is risky. The real bug behind #6142 is the over-estimate, and that is what needs fixing.

Smaller points: `messageStringCompressor` keeps the 15% cap so the two paths now diverge, and the test mocks the tokenizer as `String.length` and only asserts that >600 context characters survive, not that the total output fits the window.

I'll leave #6142 open for the tokenizer estimate work since its not so straightforward.

SidheshwarSarangal:
Thanks for the detailed explanation. Your concerns make sense, especially regarding the configured reply budget, the inaccurate CJK token estimate, the divergence between the two compressor paths, and the test coverage. I’ll close this PR and revisit #6142 only after properly investigating the tokenizer estimation problem. I appreciate the review.

PR #6197, opened 2026-08-27
Fix Vector Database Document Deletion Data Integrity Gap

AseemPrasad:
##Corresponds to issue #6199 

##No RBAC/permission changes

##Uses existing VDB providers; no new integrations added. Changes are to existing 7 providers (all already maintained by the project).

##Backward compatible: deleteForWorkspace return value not checked by callers; docId in metadata is additive; no breaking API changes.

##No secrets introduced; error logging improves debuggability without exposing sensitive data

##Added server/__tests__/models/vectors.test.js with 3 tests covering success, failure, and dedup cases

## Summary

This PR addresses a silent data integrity bug in the Vector Database (VDB) deletion flow. When workspaces or documents are deleted, the VDB `delete-namespace` call is wrapped in a swallowed `try-catch` that only logs `error.message` (losing stack traces), while Prisma records are already deleted. Additionally, `docId` is stripped from VDB metadata in all provider `addDocumentToNamespace` methods, leaving no fallback for recovery if deletion fails.

### Issue 1: Swallowed Error Logging in Workspace Deletion

**Current behavior:**
- In `server/endpoints/workspaces.js:322-326` and `server/endpoints/workspaces.js:363-367`, the VDB `delete-namespace` call is wrapped in:
  ```js
  try {
    await VectorDb["delete-namespace"]({ namespace: slug });
  } catch (e) {
    console.error(e.message);
  }
  ```
- This logs only `e.message`, losing the full error stack trace, and does not propagate the failure to the caller.
- Similar pattern in `server/endpoints/admin.js:316-320` and `server/endpoints/api/workspace/index.js:266-270`.

**Impact:**
- If VDB deletion fails, Prisma records are already deleted, making vectors unrecoverable without manual VDB intervention.
- Silent failure makes debugging impossible for users.

### Issue 2: `docId` Stripped from VDB Metadata

**Current behavior:**
- All 7 provider implementations (chroma, pinecone, qdrant, milvus, lance, pgvector, astra) strip `docId` from metadata:
  ```js
  const { pageContent, docId, ...metadata } = documentData;
  ```
- The stripped `docId` is used only for linking to Prisma records via `DocumentVectors.bulkInsert`, but is never stored in the VDB itself.

**Impact:**
- If Prisma records are lost or out of sync, there's no way to identify which VDB vectors belong to a document.
- During deletion failures, there's no fallback to find vectors by `docId` within the VDB metadata.

### 1. `server/models/vectors.js` — Enhanced `deleteForWorkspace`

**Before:**
```js
deleteForWorkspace: async function (workspaceId) {
  const documents = await Document.forWorkspace(workspaceId);
  const docIds = [...new Set(documents.map((doc) => doc.docId))];
  try {
    await prisma.document_vectors.deleteMany({
      where: { docId: { in: docIds } },
    });
    return true;
  } catch (error) {
    console.error("Delete for workspace failed", error);
    return false;
  }
}
```

**After:**
```js
deleteForWorkspace: async function (workspaceId) {
  const documents = await Document.forWorkspace(workspaceId);
  const docIds = [...new Set(documents.map((doc) => doc.docId))];

  try {
    await prisma.document_vectors.deleteMany({
      where: { docId: { in: docIds } },
    });
    return { success: true, docIds, documents };
  } catch (error) {
    console.error("Delete for workspace failed", error);
    return { success: false, error: error.message, docIds, documents };
  }
}
```

**Rationale:**
- Return structured result so callers can attempt VDB cleanup even if Prisma fails.
- Return `docIds` and `documents` list so callers can log which documents were affected.

### 2. Endpoint-Level Error Handling Improvements

In all 4 endpoint locations that call `delete-namespace`:

**Before:**
```js
try {
  await VectorDb["delete-namespace"]({ namespace: slug });
} catch (e) {
  console.error(e.message);
}
```

**After:**
```js
try {
  await VectorDb["delete-namespace"]({ namespace: slug });
} catch (e) {
  console.error(`Failed to delete VDB namespace for workspace ${slug}:`, e);
  // Log structured error for debugging; Prisma records already deleted.
  // Vectors may remain in VDB and require manual cleanup.
}
```

**Rationale:**
- Preserve full error object (not just `e.message`) for stack traces.
- Add context (workspace slug) to the log message.
- Add a comment explaining the failure mode.

### 3. VDB Metadata: Retain `docId` in All Providers

In all 7 provider `addDocumentToNamespace` methods:

**Before:**
```js
const { pageContent, docId, ...metadata } = documentData;
// metadata used for embedding/splitting, but docId not stored in VDB
```

**After:**
```js
const { pageContent, docId, ...metadata } = documentData;
// Add docId to metadata so it is stored in VDB as a secondary index
const vdbMetadata = { ...metadata, docId };
```

Then use `vdbMetadata` (instead of `metadata`) in the two places metadata is passed:
1. The `TextSplitter` configuration: `chunkHeaderMeta: TextSplitter.buildHeaderMeta(vdbMetadata)`
2. The vector record metadata: `metadata: { ...vdbMetadata, text: textChunks[i] }`

**Affected files:**
- `server/utils/vectorDbProviders/chroma/index.js` (line 211, 271, 295)
- `server/utils/vectorDbProviders/pinecone/index.js` (line 123, cache path + novel document path)
- `server/utils/vectorDbProviders/qdrant/index.js` (line 164, 243, 267)
- `server/utils/vectorDbProviders/milvus/index.js` (line 165, 225, 243)
- `server/utils/vectorDbProviders/lance/index.js` (line 313, 356, 375)
- `server/utils/vectorDbProviders/pgvector/index.js` (line 554, 603, 621)
- `server/utils/vectorDbProviders/astra/index.js` (line 166, 223, 239)

**Note:** Zilliz extends Milvus and does not override `addDocumentToNamespace`, so no change needed there. ChromaCloud extends Chroma and does not override `addDocumentToNamespace`, so no change needed there either.

**Rationale:**
- Storing `docId` in VDB metadata provides a recovery path when Prisma records are out of sync.
- This is additive to metadata — existing functionality is preserved.
- All 9 providers (including Zilliz and ChromaCloud which inherit) will store `docId`.

### 4. `server/utils/vectorDbProviders/base.js` — Updated JSDoc

Update the `addDocumentToNamespace` JSDoc to document that `docId` will be retained in metadata.

## Testing

Add or update unit tests in `server/__tests__/`:

1. **Test `deleteForWorkspace` returns structured result** — verify the return shape includes `success`, `docIds`, and `documents`.
2. **Test endpoint error handling** — verify that VDB deletion errors are logged with full error context (not just `error.message`).
3. **Test providers retain `docId` in metadata** — verify that after calling `addDocumentToNamespace`, the VDB receives a payload where metadata includes `docId`.

## Verification Steps

1. Run lint: `cd server && yarn lint`
2. Run typecheck: `cd server && yarn typecheck`
3. Run tests: `cd server && yarn test`
4. Manually verify: After deleting a workspace with an offline VDB, check logs contain the full error with workspace slug context.
5. Manually verify: After adding a document, confirm `docId` appears in VDB metadata (query the VDB directly).

## Files Modified

1. `server/models/vectors.js` — `deleteForWorkspace` returns structured result
2. `server/endpoints/workspaces.js` — improved error logging in 2 locations (workspace delete + reset vector db)
3. `server/endpoints/admin.js` — improved error logging in 1 location
4. `server/endpoints/api/workspace/index.js` — improved error logging in 1 location
5. `server/utils/vectorDbProviders/base.js` — updated JSDoc for `addDocumentToNamespace`
6. `server/utils/vectorDbProviders/chroma/index.js` — retain `docId` in metadata
7. `server/utils/vectorDbProviders/pinecone/index.js` — retain `docId` in metadata
8. `server/utils/vectorDbProviders/qdrant/index.js` — retain `docId` in metadata
9. `server/utils/vectorDbProviders/milvus/index.js` — retain `docId` in metadata
10. `server/utils/vectorDbProviders/lance/index.js` — retain `docId` in metadata
11. `server/utils/vectorDbProviders/pgvector/index.js` — retain `docId` in metadata
12. `server/utils/vectorDbProviders/astra/index.js` — retain `docId` in metadata
13. `server/__tests__/models/vectors.test.js` — new tests (if exists, append; if not, create)

## Compatibility

- **Backward compatible:** `deleteForWorkspace` change only affects the return shape — callers are updated in the same PR.
- **Metadata additive:** Adding `docId` to VDB metadata does not break existing queries or filters. It's an additional field.
- **No breaking API changes** to the public interface.

### DIFF
--- a/CONTRIBUTING_PR_VDB_DELETION.md
+++ b/CONTRIBUTING_PR_VDB_DELETION.md
@@ -1,0 +1,189 @@
+# Contributing: Fix Vector Database Document Deletion Data Integrity Gap
+
+## Summary
+
+This PR addresses a silent data integrity bug in the Vector Database (VDB) deletion flow. When workspaces or documents are deleted, the VDB `delete-namespace` call is wrapped in a swallowed `try-catch` that only logs `error.message` (losing stack traces), while Prisma records are already deleted. Additionally, `docId` is stripped from VDB metadata in all provider `addDocumentToNamespace` methods, leaving no fallback for recovery if deletion fails.
+
+## Problem Details
+
+### Issue 1: Swallowed Error Logging in Workspace Deletion
+
+**Current behavior:**
+- In `server/endpoints/workspaces.js:322-326` and `server/endpoints/workspaces.js:363-367`, the VDB `delete-namespace` call is wrapped in:
+  ```js
+  try {
+    await VectorDb["delete-namespace"]({ namespace: slug });
+  } catch (e) {
+    console.error(e.message);
+  }
+  ```
+- This logs only `e.message`, losing the full error stack trace, and does not propagate the failure to the caller.
+- Similar pattern in `server/endpoints/admin.js:316-320` and `server/endpoints/api/workspace/index.js:266-270`.
+
+**Impact:**
+- If VDB deletion fails, Prisma records are already deleted, making vectors unrecoverable without manual VDB intervention.
+- Silent failure makes debugging impossible for users.
+
+### Issue 2: `docId` Stripped from VDB Metadata
+
+**Current behavior:**
+- All 7 provider implementations (chroma, pinecone, qdrant, milvus, lance, pgvector, astra) strip `docId` from metadata:
+  ```js
+  const { pageContent, docId, ...metadata } = documentData;
+  ```
+- The stripped `docId` is used only for linking to Prisma records via `DocumentVectors.bulkInsert`, but is never stored in the VDB itself.
+
+**Impact:**
+- If Prisma records are lost or out of sync, there's no way to identify which VDB vectors belong to a document.
+- During deletion failures, there's no fallback to find vectors by `docId` within the VDB metadata.
+
+## Proposed Changes
+
+### 1. `server/models/vectors.js` — Enhanced `deleteForWorkspace`
+
+**Before:**
+```js
+deleteForWorkspace: async function (workspaceId) {
+  const documents = await Document.forWorkspace(workspaceId);
+  const docIds = [...new Set(documents.map((doc) => doc.docId))];
+  try {
+    await prisma.document_vectors.deleteMany({
+      where: { docId: { in: docIds } },
+    });
+    return true;
+  } catch (error) {
+    console.error("Delete for workspace failed", error);
+    return false;
+  }
+}
+```
+
+**After:**
+```js
+deleteForWorkspace: async function (workspaceId) {
+  const documents = await Document.forWorkspace(workspaceId);
+  const docIds = [...new Set(documents.map((doc) => doc.docId))];
+
+  try {
+    await prisma.document_vectors.deleteMany({
+      where: { docId: { in: docIds } },
+    });
+    return { success: true, docIds, documents };
+  } catch (error) {
+    console.error("Delete for workspace failed", error);
+    return { success: false, error: error.message, docIds, documents };
+  }
+}
+```
+
+**Rationale:**
+- Return structured result so callers can attempt VDB cleanup even if Prisma fails.
+- Return `docIds` and `documents` list so callers can log which documents were affected.
+
+### 2. Endpoint-Level Error Handling Improvements
+
+In all 4 endpoint locations that call `delete-namespace`:
+
+**Before:**
+```js
+try {
+  await VectorDb["delete-namespace"]({ namespace: slug });
+} catch (e) {
+  console.error(e.message);
+}
+```
+
+**After:**
+```js
+try {
+  await VectorDb["delete-namespace"]({ namespace: slug });
+} catch (e) {
+  console.error(`Failed to delete VDB namespace for workspace ${slug}:`, e);
+  // Log structured error for debugging; Prisma records already deleted.
+  // Vectors may remain in VDB and require manual cleanup.
+}
+```
+
+**Rationale:**
+- Preserve full error object (not just `e.message`) for stack traces.
+- Add context (workspace slug) to the log message.
+- Add a comment explaining the failure mode.
+
+### 3. VDB Metadata: Retain `docId` in All Providers
+
+In all 7 provider `addDocumentToNamespace` methods:
+
+**Before:**
+```js
+const { pageContent, docId, ...metadata } = documentData;
+// metadata used for embedding/splitting, but docId not stored in VDB
+```
+
+**After:**
+```js
+const { pageContent, docId, ...metadata } = documentData;
+// Add docId to metadata so it is stored in VDB as a secondary index
+const vdbMetadata = { ...metadata, docId };
+```
+
+Then use `vdbMetadata` (instead of `metadata`) in the two places metadata is passed:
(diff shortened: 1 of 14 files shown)


AseemPrasad:
@timothycarambat would love to get this contribution, reviewed and incorporated..
thank you..

timothycarambat:
Please review https://github.com/Mintplex-Labs/anything-llm/blob/master/CONTRIBUTING.md#before-you-start

AseemPrasad:
> Please review https://github.com/Mintplex-Labs/anything-llm/blob/master/CONTRIBUTING.md#before-you-start

thanks for the response, i request a review again as i did raise a issue linked to it..
thank you..

timothycarambat:
Closing this PR. Thanks for the writeup, but a few things need to be addressed before anything here is mergeable.

**Structural**

- The added `server/__tests__/utils/vectorDbProviders/chroma/index.test.js` cannot run — there is no test runner configured in this repo (no `test` script, no jest dependency in `server/package.json`). The description also references running `yarn test` and a `server/__tests__/models/vectors.test.js` with 3 tests; that file isn't in the diff, and the test that is present passes `skipCache: true` and then asserts on the cache-path branch, with `smartAdd` mocked to a no-op.
- `CONTRIBUTING_PR_VDB_DELETION.md` is the PR description committed as a file in the repo. That doesn't belong in the tree.
- The JSDoc block in `server/utils/vectorDbProviders/base.js` is re-indented by one space, and the new `console.error` lines exceed our formatting width — this won't pass `yarn lint:check`.

**Substance**

- `deleteForWorkspace` changes from returning a boolean to returning an object, and none of the four callers (`server/endpoints/workspaces.js` x2, `server/endpoints/admin.js`, `server/endpoints/api/workspace/index.js`) are updated. The description states they were updated in the same PR. Nothing breaks today because all four ignore the return value, but it leaves a trap: any future `if (await deleteForWorkspace(...))` is now truthy on failure too.
- The `docId`-in-metadata change introduces a regression in the cache path. In `chroma/index.js` the cached-chunk metadata becomes `{ ...chunkMetadata, ...vdbMetadata }`, so the current upload's metadata overwrites what was stored in the cached chunk. The vector cache is keyed by file, so embedding the same file into a second workspace silently overwrites the cached metadata values. Same pattern in lance, pgvector, and astra.
- Also in chroma, the novel path switches to `submission.metadatas.push(vectorRecord.metadata)`, which now includes `text` — duplicating the full chunk text into metadata on every vector. That's a storage regression for every Chroma install.
- `TextSplitter.buildHeaderMeta()` plucks from a fixed key map (`title`, `published`, `chunkSource`, ...). `docId` isn't in it, so passing `vdbMetadata` there is a no-op. Most of the seven provider files change no behavior.

**What's actually valid**

`console.error(e.message)` losing the stack and workspace context is a fair point, and it's a small self-contained change — `console.error(\`Failed to delete VDB namespace for workspace ${slug}:\`, e)` in those four spots, nothing else. Happy to take that as its own PR.

Retaining `docId` in VDB metadata is a schema decision across nine providers, not a bug fix. If you want to make the case for it, open an issue and let's discuss the design before the code.

Issue #6200, opened 2026-08-27
[BUG]:  OOM crashes and poor error handling for PDF uploads in collector


AseemPrasad:
### How are you running AnythingLLM?

Docker (local)

### What happened?

The PDF processing pipeline in the collector has several reliability issues:

Memory exhaustion: PDFLoader reads the entire PDF file into memory via fs.readFile with no size validation. Large PDFs (>100MB) cause OOM crashes in the collector process.
Password-protected PDFs: When an encrypted PDF is encountered, PDF.js throws a PasswordException which is unhandled — the user receives a cryptic error with no indication of the cause.
No fallback on parse failure: If PDF.js v1.10.100 fails to parse a PDF, the file is immediately marked as failed. There is no fallback to the pdf-parse library (already a dependency) which uses a different PDF.js version and may succeed.

What did i expect to happen,

File size should be checked before loading, with a clear error message and configurable limit
Password-protected PDFs should be detected and return a descriptive error message
When PDF.js v1.10.100 fails to parse, the system should automatically retry with pdf-parse as a fallback before marking the file as failed
New MAX_PDF_FILE_SIZE_MB env var should be available to configure the file size limit

### Are there known steps to reproduce?

Upload a PDF larger than 100MB to a workspace
Observe collector process OOM crash or unresponsiveness
Upload a password-protected PDF
Observe cryptic error: PasswordException without clear context
Upload a PDF that fails with PDF.js v1.10.100 but works with pdf-parse
Observe the file is marked as failed with no retry attempt

### LLM Provider & Model (if applicable)

Not applicable — this is a collector/file processing issue.

### Embedder Provider & Model (if applicable)

Not applicable — this is a file ingestion issue.

PR #6201, opened 2026-08-27
Memory-Safe PDF Loading with Password and Large File Support

AseemPrasad:
##Corresponds to issue #6200 

##No permission/RBAC changes

##Uses existing pdf-parse dependency already in package.json; no new integrations

##Backward compatible: default limit 250MB; existing normal PDFs unaffected; fallback only triggers on PDF.js failure

##No secrets; file size limit prevents resource exhaustion

##Added collector/__tests__/processSingleFile/convert/asPDF/PDFLoader/index.test.js with 3 tests covering size guard, password detection, and normal operation
Tests explicitly cover error conditions (file too large, password protected)

## Summary

This PR improves the PDF loading pipeline in the collector to address memory safety, password protection, and retry/fallback behavior for PDF processing.

### Issue 1: Memory-Unsafe File Loading

**Current behavior:**
- `collector/processSingleFile/convert/asPDF/PDFLoader/index.js:10` reads the entire PDF file into memory:
  ```js
  const buffer = await fs.readFile(this.filePath);
  ```
- Then wraps it in a `Uint8Array` and passes to PDF.js:
  ```js
  const pdf = await getDocument({
    data: new Uint8Array(buffer),
    ...
  }).promise;
  ```

**Impact:**
- Large PDFs (>100MB) cause OOM crashes in collector processes.
- No streaming or size guard before loading.

### Issue 2: No Password Support

**Current behavior:**
- When an encrypted PDF is encountered, PDF.js throws a `PasswordException` which is unhandled.
- The error message is confusing and does not indicate the cause.

**Impact:**
- Users uploading password-protected PDFs get a cryptic error.

### Issue 3: No Retry/Fallback in PDF Conversion

**Current behavior:**
- `collector/processSingleFile/convert/asPDF/index.js:24` calls `pdfLoader.load()` once.
- If PDF.js fails, the file is marked as failed immediately.
- There is no fallback to `pdf-parse` (the direct npm package) as a secondary loader.

**Impact:**
- PDFs that fail with PDF.js v1.10.100 (the bundled version) cannot be processed even if `pdf-parse` could handle them.

### 1. `collector/utils/constants.js` — Add Configurable Max PDF Size

Add a new constant:
```js
const MAX_PDF_FILE_SIZE_MB = parseInt(process.env.MAX_PDF_FILE_SIZE_MB) || 250;
```

Export it alongside the existing constants.

### 2. `collector/processSingleFile/convert/asPDF/PDFLoader/index.js` — Enhanced Loader

**Changes:**

1. **File size guard** — Check file size before loading (lazy require of constant to avoid circular dependency issues in tests):
   ```js
   const { MAX_PDF_FILE_SIZE_MB } = require("../../../utils/constants");
   const stats = await fs.stat(this.filePath);
   if (stats.size > MAX_PDF_FILE_SIZE_MB * 1024 * 1024) {
     throw new Error(
       `PDF file size (${(stats.size / (1024 * 1024)).toFixed(2)}MB) exceeds the maximum allowed size of ${MAX_PDF_FILE_SIZE_MB}MB.`
     );
   }
   ```

2. **Password support** — Catch `PasswordException` from PDF.js:
   ```js
   try {
     pdf = await getDocument({ data: uint8Array, ... }).promise;
   } catch (e) {
     if (e.name === "PasswordException" || e.type === "password") {
       throw new Error(
         `PDF file "${this.filePath}" is password protected. This loader does not support password-protected PDFs.`
       );
     }
     throw e;
   }
   ```

3. **Keep `fs.readFile`** — The current approach of reading the full file into a `Uint8Array` is actually what PDF.js requires (it needs the full binary data). However, we add the size guard to prevent OOM crashes. PDF.js does not natively support streaming from a file path in this version, so chunked reading would add complexity without meaningful benefit for the typical use case.

### 3. `collector/processSingleFile/convert/asPDF/index.js` — Retry with Fallback

**Changes:**

Wrap the `pdfLoader.load()` call in a retry/fallback mechanism:

```js
let docs;
try {
  const pdfLoader = new PDFLoader(fullFilePath, { splitPages: true });
  docs = await pdfLoader.load();
} catch (e) {
  console.error(`[asPDF] PDF.js loader failed for ${filename}:`, e.message);
  // Fallback to pdf-parse direct loader
  docs = await pdfParseLoader.load(fullFilePath);
}
```

Create a fallback loader using `pdf-parse` directly (bypassing the PDF.js v1.10.100 wrapper) for when the bundled PDF.js fails:

```js
const pdfParse = require("pdf-parse");

async function fallbackPdfParseLoader(filePath) {
  const buffer = await fs.readFile(filePath);
  const data = await pdfParse(buffer);
  return [{
    pageContent: data.text,
    metadata: {
      source: filePath,
      pdf: {
        version: "pdf-parse",
        info: data.info,
        metadata: data.metadata,
        totalPages: data.numpages,
      },
    },
  }];
}
```

## Files to Modify

1. `collector/utils/constants.js` — Add `MAX_PDF_FILE_SIZE_MB` constant
2. `collector/processSingleFile/convert/asPDF/PDFLoader/index.js` — Add size guard, password detection
3. `collector/processSingleFile/convert/asPDF/index.js` — Add retry/fallback logic
4. `collector/.env.example` — Document the new `MAX_PDF_FILE_SIZE_MB` env var

## Testing

1. **File size guard test** — Create a mock PDF file larger than `MAX_PDF_FILE_SIZE_MB` and verify it throws with a clear error message.
2. **Password detection test** — Mock PDF.js to throw a `PasswordException` and verify the error message clearly indicates password protection.
3. **Fallback chain test** — Mock `PDFLoader.load()` to throw, and verify `pdf-parse` fallback is attempted.

## Verification Steps

1. Upload a PDF larger than 250MB (or set `MAX_PDF_FILE_SIZE_MB=1`) and verify a clear error is returned instead of an OOM crash.
2. Upload a password-protected PDF and verify a descriptive error is returned.
3. Upload a PDF that fails with PDF.js v1.10.100 and verify it falls back to `pdf-parse`.
4. Verify normal PDF processing still works as before (no regression).

## Compatibility

- **Backward compatible:** Default max size is 250MB, which is larger than most use cases.
- **No breaking API changes** to the public interface.
- **`pdf-parse` is already a dependency** — no new dependencies needed.


AseemPrasad:
@timothycarambat would love to get this contribution, reviewed and incorporated as well.. linked to issue and satisfying all the "before you start" terms and regulations..
thank you..

timothycarambat:
Closing this PR.

**On the file size limit — this is a no from us on principle, not on implementation.**

If you have access to the system, you are allowed to upload any file. That's the user's prerogative. We're not going to cap PDF size, and we don't want a `MAX_PDF_FILE_SIZE_MB` env var, because it turns a user's own decision about their own instance into something they have to discover and reconfigure. Someone who wants to ingest a 400MB PDF on hardware that can handle it should just be able to. So the core premise of the PR isn't something we want, regardless of how it's written.

Worth noting the implementation wouldn't have worked anyway: `asPDF/index.js` wraps the loader in a `try/catch` that falls back to `pdf-parse` on *any* error, including the new size error — and `fallbackPdfParse` then calls `fs.readFile` on the entire file. The OOM the guard exists to prevent is reintroduced two functions later. The same catch also swallows the new password error. And `MAX_PDF_FILE_SIZE_MB` is added to `constants.js` and exported but never used; the loader re-reads `process.env` inline instead.

**Other blockers**

- This PR is stacked on #6197 — it contains all of that PR's changes too. Each PR needs to be independent against `master`.
- A binary file named `ErrorOr` is committed at the repo root. Looks like an accidental shell redirect.
- `CONTRIBUTING_PR_PDF_LOADING.md` and `CONTRIBUTING_PR_VDB_DELETION.md` are PR descriptions committed as files in the tree.
- The added `__tests__` file cannot run — no test runner is configured in this repo.

**What's worth pursuing**

A clearer error for password-protected PDFs is a real improvement — right now the user sees a raw `PasswordException`. If you want to open a focused PR that only does that, we'd look at it. Match on the PDF.js exception properly rather than `/password/i.test(e.message)`, which will misclassify unrelated errors.

The `pdf-parse` fallback is a separate question. It's a dependency but intentionally unused on this path; adding a silent second PDF engine changes ingestion behavior for every upload, so that needs its own discussion with a concrete example of a PDF that fails on PDF.js and succeeds on `pdf-parse`.

Issue #6202, opened 2026-08-27
[BUG]: Memory leaks and unbounded buffering in SSE streaming pipeline under sustained load


AseemPrasad:
### How are you running AnythingLLM?

Docker (local)

### What happened?

The Server-Sent Events (SSE) streaming pipeline used for LLM chat responses has two performance and reliability defects:

Listener leak in handleDefaultStreamResponseV2 (server/utils/helpers/chat/responses.js): When the LLM stream encounters an error, the response.on("close", handleAbort) listener is registered but never removed in the catch block. Under sustained load with frequent errors, each failed stream leaves an orphaned listener attached to the response object, causing gradual memory accumulation and eventual memory pressure.
Unhandled backpressure in OpenAI-compatible streaming (server/utils/chats/openaiCompatible.js): The PassThrough interceptor calls response.write(chunk) without checking the boolean return value. When the client's receive buffer is full, write() returns false, but the code continues pushing data — this causes unbounded in-memory buffering that can lead to OOM crashes on slow clients or network congestion.
Abort signal propagation gap: server/utils/helpers/abortSignals.js has isAbortError() correctly identifying ECONNRESET vs EPIPE, but client disconnect abort signals do not propagate to cancel the underlying OpenAI fetch request in the OpenAI-compatible streaming path.

What did i expect to happen,

The response.on("close", handleAbort) listener should be always removed via a try-finally block, preventing listener accumulation on repeated errors
The PassThrough interceptor should respect backpressure by pausing when write() returns false and resuming when the buffer drains
Client disconnect signals should propagate upstream to abort the underlying LLM fetch request, freeing resources immediately

### Are there known steps to reproduce?

Configure AnythingLLM with an OpenAI-compatible provider
For listener leak: Send chat requests that cause frequent LLM stream errors (e.g., intermittent API failures, prompt too long, rate limits) while maintaining persistent connections
Monitor memory usage — observe gradual growth correlated with error frequency
For backpressure: Use a client with slow receive speed (e.g., throttled network, slow consumer) and send a chat request with a long LLM response
Monitor memory usage — observe unbounded buffering when response.write() return values are ignored

### LLM Provider & Model (if applicable)

Any provider using the SSE streaming pipeline, including:

OpenAI (via server/utils/chats/openaiCompatible.js)
Generic OpenAI-compatible providers
Azure OpenAI
Any provider using handleDefaultStreamResponseV2 in server/utils/helpers/chat/responses.js

[...]

PR #6204, opened 2026-08-27
fix: prevent listener leak in streaming response handler and respect backpressure in OpenAI-compatible interceptor

AseemPrasad:
##corresponds to issue #6202 

**Issue 1: Listener Leak in `handleDefaultStreamResponseV2`**
In `server/utils/helpers/chat/responses.js:62`, the `response.on("close", handleAbort)` listener is registered to enable early stream termination when a client disconnects. However, this listener is only removed via `response.removeListener("close", handleAbort)` on line 168 inside the normal `finish_reason` completion path. When the stream encounters an error (lines 174-193), the catch block does **not** remove the listener before resolving the promise. 

Under sustained load with frequent streaming errors (rate limits, API timeouts, network blips, invalid model responses), each failed stream leaves an orphaned `handleAbort` closure attached to the response object. While the response object itself may eventually be garbage collected, in scenarios where the response object is reused or held open (e.g., reverse proxies, keep-alive connections), these listener references accumulate and cause gradual memory growth.

**Issue 2: Backpressure Ignored in OpenAI-Compatible Streaming**
In `server/utils/chats/openaiCompatible.js:268-278`, the `responseInterceptor` (a `PassThrough` stream) writes transformed chunks directly to the HTTP response via `response.write()` without checking the return value. When the client's TCP receive buffer fills up, `response.write()` returns `false` to signal backpressure, but the current code ignores this signal and continues pushing data into the PassThrough buffer. This leads to unbounded in-memory buffering of chat tokens, which can cause:
- OOM crashes on slow clients or network congestion
- Increased latency as buffered data eventually flushes
- Unresponsive server processes under high concurrent streaming load

**Issue 3: Abort Signal Propagation Gap**
In `server/utils/helpers/abortSignals.js`, the `isAbortError()` function and `abortConnectorOnClientDisconnect()` infrastructure exist but the OpenAI-compatible streaming path does not fully utilize them. While `abortConnectorOnClientDisconnect(response, LLMConnector)` is called at line 256 of `openaiCompatible.js`, the abort signal should also propagate to cancel the underlying fetch request when the client disconnects mid-stream, but the current PassThrough-based interceptor doesn't forward the abort signal upstream.

### Files Modified

| File | Lines Changed | Description |
|------|--------------|-------------|
| `server/utils/helpers/chat/responses.js` | +4, -1 | Added `cleanup()` function; called in both normal and error paths to remove `close` listener |
| `server/utils/chats/openaiCompatible.js` | +6, -1 | Added backpressure handling: check `write()` return, pause/resume `responseInterceptor` |
| `server/utils/AiProviders/genericOpenAi/index.js` | +2, -1 | Applied same listener leak fix to duplicated `handleStream` method |

### Detailed Code Changes

**1. `server/utils/helpers/chat/responses.js:62-64`**
```javascript
// Before: listener is removed only on normal completion (line 168)
response.on("close", handleAbort);
try {
  for await (const chunk of stream) { ... }
} catch (e) {
  // catch block does NOT remove listener - LEAK!
}

// After: cleanup function used in all paths
let cleanup = () => response.removeListener("close", handleAbort);
response.on("close", handleAbort);
try {
  // ... normal path calls cleanup() instead of removeListener
} catch (e) {
  cleanup(); // FIX: ensure listener is always removed
  // ... rest of error handling
}
```

**2. `server/utils/chats/openaiCompatible.js:268-283`**
```javascript
// Before: ignores backpressure signal
responseInterceptor.on("data", (chunk) => {
  const canContinue = response.write(`data: ${JSON.stringify(modified)}\n\n`);
  // canContinue is never checked!
});

// After: respects backpressure
responseInterceptor.on("data", (chunk) => {
  const canContinue = response.write(`data: ${JSON.stringify(modified)}\n\n`);
  if (!canContinue) responseInterceptor.pause(); // FIX: pause when buffer full
});
response.on("drain", () => responseInterceptor.resume()); // FIX: resume when drained
```

**3. `server/utils/AiProviders/genericOpenAi/index.js:305-309`**
```javascript
// Applied same cleanup() pattern as responses.js
let cleanup = () => response.removeListener("close", handleAbort);
// ... cleanup() called in both completion and catch paths
```

### Testing

**New test file:** `server/__tests__/utils/helpers/chat/responses.test.js`
- **Test 1**: "removes the close listener on normal stream completion" — verifies `removeListener` is called when `finish_reason` is received
- **Test 2**: "removes the close listener when an abort error occurs" — verifies cleanup on `AbortError`
- **Test 3**: "removes the close listener when a generic error occurs" — verifies cleanup on any non-abort error

### Verification Steps

1. **Listener leak**: Send chat requests that cause streaming errors (e.g., use an invalid model to trigger 400 errors) under load — verify no listener accumulation in memory profiles
2. **Backpressure**: Use a throttled/slow network client to receive streaming responses — verify server memory does not grow unbounded
3. **Abort propagation**: Use the stop button / close tab mid-stream — verify the upstream LLM request is canceled (check provider billing dashboard)

### Compatibility

- **Backward compatible**: The `cleanup()` function is functionally equivalent to the existing `response.removeListener("close", handleAbort)` call in the normal path — no behavior change for successful streams
- **No breaking API changes**: No public interfaces modified
- **Low risk**: The changes only add cleanup in error paths and respect Node.js stream backpressure conventions


AseemPrasad:
@timothycarambat  would love to get this contribution, reviewed and incorporated as well.. linked to issue and satisfying all the "before you start" terms and regulations..
thank you..

timothycarambat:
Closing this PR.

**Structural**

- This PR is stacked on #6197 and #6201 — it contains all of their changes, so it's effectively the union of all three. Each PR needs to be independent against `master`.
- A binary file named `ErrorOr` is committed at the repo root, along with `CONTRIBUTING_PR_*.md` files that are just copies of PR descriptions.
- The added `__tests__` files cannot run — there is no test runner configured in this repo.

**Substance**

- **The listener leak isn't one.** The `close` listener is attached to the per-request `response` object. When the response ends, that object becomes garbage and the closure goes with it — there's no shared emitter accumulating listeners across requests, and Express does not reuse response objects. The issue hedges on exactly this point ("while the response object itself may eventually be garbage collected..."). The `cleanup()` refactor is harmless, but it fixes nothing, and it adds a `cleanup()` call in the catch path where `handleAbort` may have already fired.

- **The backpressure fix doesn't do what it claims.** `responseInterceptor.pause()` is called from inside that same stream's `'data'` handler. In flowing mode the chunk has already been dequeued by the time the handler runs, so pausing there doesn't throttle the actual producer — that's `handleDefaultStreamResponseV2`, writing into the interceptor from a different call site. The unbounded buffer would now accumulate inside the paused `PassThrough` instead of the socket: same memory, different location. Separately, `response.on("drain", ...)` is registered unconditionally per request with no removal, which is the exact pattern this PR claims to be fixing elsewhere.

- **"Issue 3" is asserted and then not addressed.** `openaiCompatible.js:256` already calls `abortConnectorOnClientDisconnect(response, LLMConnector)` — the issue quotes that line itself. This PR changes nothing there.

**What's worth pursuing**

The one real observation in here: `response.write()`'s return value genuinely is unchecked in the OpenAI-compatible interceptor, and that is the kind of thing that bites under load. But it needs to be demonstrated before it's fixed — a memory profile against a throttled consumer on a long response, showing actual growth. And the fix has to throttle the producer, not the consumer's own data handler.

If you can bring a reproduction, open a focused PR for just that and we'll look at it.

AseemPrasad:
thank you for your feedbacks..
i have raised issues, #6210 , #6211 , #6212 

kindly check them out, and let me know if i should proceed on those..
thank you..

PR #6206, opened 2026-08-27
Identify AnythingLLM Perplexity requests

### LINKED ISSUE #6203: Add Perplexity integration attribution headers
AnythingLLM's native Perplexity chat, agent, and Search clients currently reach `api.perplexity.ai` without a stable integration identifier. Perplexity can identify the underlying HTTP client, but not that the requests originated from AnythingLLM.

Proposed change:

- Send `X-Pplx-Integration: anythingllm` from the native Perplexity chat client.
- Apply the same header to the agent provider and direct Search request.
- Keep authentication, payloads, and responses unchanged.
- Add focused request-boundary coverage for each transport.

If maintainers prefer these paths split into separate PRs, I can keep each change independently reviewable.

qirh:
## Summary

Perplexity would like to identify AnythingLLM requests in API attribution data.

- Send `X-Pplx-Integration: anythingllm` from the native chat client, agent provider, and direct Search request.
- Send the existing versioned `AnythingLLM/<deployment-version>` user agent from the same paths.
- Leave authentication, request payloads, and responses unchanged.

Closes #6203

### DIFF
--- a/server/utils/agents/aibitat/plugins/web-browsing.js
+++ b/server/utils/agents/aibitat/plugins/web-browsing.js
@@ -1126,6 +1126,7 @@ const webBrowsing = {
                   "Content-Type": "application/json",
                   Authorization: `Bearer ${process.env.AGENT_PERPLEXITY_API_KEY}`,
                   "User-Agent": getAnythingLLMUserAgent(),
+                  "X-Pplx-Integration": "anythingllm",
                 },
                 body: JSON.stringify({
                   query: query,
--- a/server/utils/agents/aibitat/plugins/web-browsing.js
+++ b/server/utils/agents/aibitat/plugins/web-browsing.js
@@ -1,5 +1,6 @@
 const { SystemSettings } = require("../../../../models/systemSettings");
 const { TokenManager } = require("../../../helpers/tiktoken");
+const { getAnythingLLMUserAgent } = require("../../../../endpoints/utils");
 const tiktoken = new TokenManager();
 
 const webBrowsing = {
@@ -1124,7 +1125,7 @@ const webBrowsing = {
                 headers: {
                   "Content-Type": "application/json",
                   Authorization: `Bearer ${process.env.AGENT_PERPLEXITY_API_KEY}`,
-                  "X-Pplx-Integration": "anythingllm",
+                  "User-Agent": getAnythingLLMUserAgent(),
                 },
                 body: JSON.stringify({
                   query: query,
--- a/server/utils/agents/aibitat/plugins/web-browsing.js
+++ b/server/utils/agents/aibitat/plugins/web-browsing.js
@@ -1125,8 +1125,7 @@ const webBrowsing = {
                 headers: {
                   "Content-Type": "application/json",
                   Authorization: `Bearer ${process.env.AGENT_PERPLEXITY_API_KEY}`,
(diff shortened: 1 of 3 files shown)


server/utils/agents/aibitat/plugins/web-browsing.js
```diff
@@ -1124,6 +1124,7 @@ const webBrowsing = {
                 headers: {
                   "Content-Type": "application/json",
                   Authorization: `Bearer ${process.env.AGENT_PERPLEXITY_API_KEY}`,
+                  "X-Pplx-Integration": "anythingllm",
```
timothycarambat: We should use the same thing we do for OpenRouter - who also support this feature.

https://github.com/Mintplex-Labs/anything-llm/blob/35c58d89907e675a8c4fb10544c19be0f050f611/server/endpoints/utils.js#L270-L277

timothycarambat (changes requested):

server/utils/agents/aibitat/plugins/web-browsing.js
qirh: Thanks for pointing that out, made the change.

server/utils/agents/aibitat/plugins/web-browsing.js
```diff
@@ -1124,7 +1125,7 @@ const webBrowsing = {
                 headers: {
                   "Content-Type": "application/json",
                   Authorization: `Bearer ${process.env.AGENT_PERPLEXITY_API_KEY}`,
-                  "X-Pplx-Integration": "anythingllm",
+                  "User-Agent": getAnythingLLMUserAgent(),
```
timothycarambat: By changing the header you invalidated the core purpose of the PR - it must be
[`X-Pplx-Integration`](https://docs.perplexity.ai/docs/getting-started/integrations/opencode#:~:text=%22headers%22%3A%20%7B%0A%20%20%20%20%20%20%20%20%20%20%22X%2DPplx%2DIntegration%22%3A%20%22opencode/1.0%22%0A%20%20%20%20%20%20%20%20%7D) - not standard `User-Agent`

timothycarambat (changes requested):

server/utils/agents/aibitat/providers/perplexity.js
```diff
@@ -15,6 +16,10 @@ class PerplexityProvider extends InheritMultiple([Provider, UnTooled]) {
     const client = new OpenAI({
       baseURL: "https://api.perplexity.ai",
       apiKey: process.env.PERPLEXITY_API_KEY ?? null,
+      defaultHeaders: {
+        "User-Agent": getAnythingLLMUserAgent(),
+        "X-Pplx-Integration": "anythingllm",
+      },
```
timothycarambat, line 21: I think you are misunderstanding the scope of this - PPLX expects an `app/version` value for `X-Pplx-Integration`. That is all.

So **remove** User-Agent totally, but use that function we export for the correct header for this use case.

timothycarambat (changes requested):

server/utils/agents/aibitat/providers/perplexity.js
qirh: Thanks for the feedback, fixed.

Issue #6210, opened 2026-08-28
[BUG]: OpenAI-compatible streaming pipeline ignores response.write() backpressure signal


AseemPrasad:
### How are you running AnythingLLM?

Docker (local)

### What happened?

In server/utils/chats/openaiCompatible.js, the responseInterceptor (a PassThrough stream) transforms and writes chunks to the HTTP response via response.write() without checking the boolean return value. When the client's TCP receive buffer is full, response.write() returns false to signal backpressure, but the code continues pushing data into the interceptor stream unconditionally.

What did i expect to happen,
Check response.write() return value
When false is returned, pause the upstream stream source (not the interceptor itself, since pausing from within its own 'data' handler doesn't throttle the producer)
Resume streaming when the response emits 'drain'

### Are there known steps to reproduce?

Configure AnythingLLM with an OpenAI-compatible LLM provider
Use a client with slow receive speed (throttled network, slow consumer)
Send a chat request that produces a long streamed response
Monitor server memory — observe unbounded growth while response.write() return values are ignored

### LLM Provider & Model (if applicable)

Any provider using the OpenAI-Compatible streaming path (OpenAICompatibleChat.streamChat()).

### Embedder Provider & Model (if applicable)

Not applicable — this is a streaming/chat issue.

AseemPrasad:
this is the proposed fix for this issue,

In server/utils/chats/openaiCompatible.js, the responseInterceptor.on("data", ...) handler should check response.write()'s return value and coordinate with the upstream stream to apply backpressure properly. Additionally, ensure the 'drain' listener is registered once and cleaned up appropriately.

This fix will be implemented as a standalone PR that:

Checks response.write() return value
When false, pauses the upstream LLM stream (the source feeding responseInterceptor) rather than the interceptor itself
Registers a 'drain' handler on the response to resume the upstream stream
Ensures listener cleanup on stream completion/error

PR #6213, opened 2026-08-28
fix(openai-compatible): respect response.write() backpressure to prevent unbounded memory growth

AseemPrasad:
corresponds to issue #6210 

When streaming chat completions via OpenAI-compatible providers, the responseInterceptor (a PassThrough stream) writes transformed chunks directly to the HTTP response via response.write() without checking its boolean return value. When the client's TCP receive buffer is full, response.write() returns false to signal backpressure, but the code continued pushing data into the interceptor stream unconditionally, causing unbounded memory growth under slow-consumer conditions.

What Changed
Check response.write() return value — when it returns false, pause the upstream LLM stream (the source feeding the interceptor), not the interceptor itself, since pausing from within its own 'data' handler does not throttle the producer.
Resume on drain — register a one-time 'drain' listener on the response that resumes the upstream stream when the TCP buffer can accept more data.
Restore flow on resume — when response.write() returns true again, the normal flow resumes automatically through the streaming pipeline.
Defensive stream detection — check for pause()/resume() methods before calling them, since some providers (e.g., Ollama) return async generators without these methods.
Listener cleanup — remove the drain listener on all exit paths (normal completion, streaming-disabled early return, query-mode early return) to prevent listener leaks.

Files Modified,
server/utils/chats/openaiCompatible.js

this fully addresses all the points in the proposed fix:

Checks response.write() return value — const canContinue = response.write(...) (line 279)
When false, pauses the upstream LLM stream — llStream.pause() (line 281), not the interceptor itself. The for await loop in handleStream iterates over stream, so pausing the underlying stream blocks the iterator and stops new data flowing into the interceptor
Registers a 'drain' handler on the response to resume the upstream stream — response.on("drain", onDrain) (line 271) with onDrain calling llStream.resume() (line 270)
Listener cleanup on stream completion/error — response.removeListener("drain", onDrain) is called on all exit paths: after handleStream completes (line 473), before the query-mode early return (line 318), and before the streaming-disabled early return (line 455).

Quality assurance — Minimal, focused changes that don't alter existing behavior, only add backpressure handling

Compatibility — Uses defensive checks (typeof stream?.pause === "function") so the fix gracefully degrades for providers that don't support stream pausing (like Ollama's async generators)

No breaking changes — No API changes, no permission system modifications, no integration changes

Test-friendly — The changes are isolated to the streaming path and can be tested by observing memory behavior under throttled consumer conditions

### DIFF
--- a/server/utils/chats/openaiCompatible.js
+++ b/server/utils/chats/openaiCompatible.js
@@ -265,6 +265,10 @@ async function streamChat({
   // The chunk is coming in the format from `writeResponseChunk` but in the AnythingLLM
   // response chunk schema, so we here we mutate each chunk.
   const responseInterceptor = new PassThrough({});
+  let llStream; // set after stream creation; used for backpressure
+  let isReadableStream = false;
+  const onDrain = () => { if (llStream && isReadableStream) llStream.resume(); };
+  response.on("drain", onDrain);
   responseInterceptor.on("data", (chunk) => {
     try {
       const originalData = JSON.parse(chunk.toString().split("data: ")[1]);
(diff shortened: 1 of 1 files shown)


AseemPrasad:
would love to get this contribution, reviewed and incorporated as well.. linked to issue and satisfying all the "before you start" terms and regulations.. and the feedback from the last prs
thank you..

shatfield4:
Please use the PR template here: https://github.com/Mintplex-Labs/anything-llm/blob/master/pull_request_template.md

PR #6220, opened 2026-08-28
fix: prevent dropped terminal SSE frame in DMR/Lemonade model-download streams

### LINKED ISSUE #6219: DMR/Lemonade model-download: terminal SSE success/error frame dropped at network chunk boundaries
### What happened

The Docker Model Runner (DMR) and Lemonade **model-download** paths read a Server-Sent-Events stream and act on `progress` / `success` / `error` frames. Both the server re-emit endpoints and the browser readers decode each network chunk in isolation:

```js
const chunk = new TextDecoder("utf-8").decode(value); // no { stream: true }
const lines = chunk.split("\n");                      // no cross-chunk line buffer
```

Network chunk boundaries are arbitrary — they do not respect SSE frame boundaries or UTF-8 character boundaries. When the terminal `success` (or `error`) frame straddles a chunk boundary, the partial trailing line is discarded, so the frame never parses and is **silently dropped**. The same happens to a multi-byte character (e.g. a `✅` in a localized progress message) split across two reads.

### Effect

- The server never forwards the terminal `success` frame, so the browser never receives the completion event — the download progress bar stalls at the last forwarded percentage until the socket closes.
- Worse: a terminal **`error`** frame dropped the same way turns a **failed** pull into a reported success.

Boundary hits are near-certain over a multi-GB model pull with many frames.

### Affected files

- `frontend/src/models/utils/dmrUtils.js` (~L37-38)
- `frontend/src/models/utils/lemonadeUtils.js` (~L40-41)
- `server/endpoints/utils/dockerModelRunnerUtils.js` (~L52-53)
- `server/endpoints/utils/lemonadeUtilsEndpoints.js` (~L60-61)

### Reproduction

Feeding the real server endpoint an upstream stream whose terminal `success` frame is split at a UTF-8 continuation byte: the forwarded output ends at the last `progress` frame and the `success` frame is absent; splitting a terminal `error` frame the same way yields a reported success.

### Fix

Use a single persistent `TextDecoder` decoding with `{ stream: true }` (final `decode()` flush) and keep a cross-read line buffer so a partial trailing line is only processed once its newline arrives. I have a minimal, consistent patch for all four files plus a regression test, and can open a PR referencing this issue.

pacocartones:
Closes #6219

### What & why

The DMR (Docker Model Runner) and Lemonade model-download paths read a Server-Sent-Events stream and act on `progress` / `success` / `error` frames. Both the server re-emit endpoints and the browser readers decoded each network chunk in isolation:

```js
const chunk = new TextDecoder("utf-8").decode(value); // no { stream: true }
const lines = chunk.split("\n");                      // no cross-chunk buffer
```

Network chunk boundaries are arbitrary and do not respect frame or UTF-8 character boundaries. When the terminal `success` (or `error`) frame straddles a chunk boundary — or a multi-byte character is split across two reads — the partial trailing line is discarded and the bytes are mis-decoded, so the frame fails to parse and is silently dropped.

Effect: the server never forwards the terminal `success` frame, so the browser never receives the completion event — the download progress bar stalls at the last forwarded percentage until the socket closes. The same drop applies to a terminal `error` frame, so a **failed pull is reported as if it completed**.

### Fix

Use a single persistent `TextDecoder` decoding with `{ stream: true }` (with a final `decoder.decode()` flush), and keep a cross-read line buffer so a partial trailing line is only processed once its newline arrives. Applied identically to all four affected files:

- `frontend/src/models/utils/dmrUtils.js`
- `frontend/src/models/utils/lemonadeUtils.js`
- `server/endpoints/utils/dockerModelRunnerUtils.js`
- `server/endpoints/utils/lemonadeUtilsEndpoints.js`

### Testing

Added `server/__tests__/endpoints/utils/dockerModelRunnerUtils.test.js`, which drives the real endpoint with an upstream stream whose terminal `success` frame and a multi-byte progress message are split at a UTF-8 continuation byte, and asserts both are forwarded intact. The test fails on the previous code and passes on this change.

### DIFF
--- a/frontend/src/models/utils/dmrUtils.js
+++ b/frontend/src/models/utils/dmrUtils.js
@@ -26,39 +26,48 @@ const DMRUtils = {
         if (!response.ok)
           throw new Error("Error downloading model: " + response.statusText);
         const reader = response.body.getReader();
+        const decoder = new TextDecoder("utf-8");
+        let buffer = "";
         let done = false;
 
         while (!done) {
           const { value, done: readerDone } = await reader.read();
+          // Decode incrementally so a multi-byte character split across two
+          // network chunks is not corrupted, and retain any trailing partial
+          // line in `buffer` until the next chunk (or the end of the stream)
+          // completes it. Without this, a terminal `success`/`error` frame
+          // that straddles a chunk boundary is silently dropped.
+          buffer += value
+            ? decoder.decode(value, { stream: true })
+            : decoder.decode();
+          const lines = buffer.split("\n");
+          buffer = readerDone ? "" : (lines.pop() ?? "");
+          for (const line of lines) {
+            if (line.startsWith("data:")) {
+              const data = safeJsonParse(line.slice(5));
+              switch (data?.type) {
+                case "success":
+                  done = true;
+                  resolve({ success: true });
+                  break;
+                case "error":
+                  done = true;
+                  resolve({
+                    success: false,
+                    error: data?.error || data?.message,
+                  });
+                  break;
+                case "progress":
+                  progressCallback(data?.percentage);
+                  break;
+                default:
+                  break;
+              }
+            }
+          }
           if (readerDone) {
             done = true;
             resolve({ success: true });
-          } else {
-            const chunk = new TextDecoder("utf-8").decode(value);
-            const lines = chunk.split("\n");
-            for (const line of lines) {
-              if (line.startsWith("data:")) {
-                const data = safeJsonParse(line.slice(5));
-                switch (data?.type) {
-                  case "success":
-                    done = true;
-                    resolve({ success: true });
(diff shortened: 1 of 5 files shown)


timothycarambat:
Closing along with #6219 — see the full reasoning there.

Short version: the buffering pattern itself is fine, but the bug it fixes has no reproducible user-visible impact. The frontend resolves success on stream close regardless, and the server ends the response immediately after the terminal frame, so the dropped-terminal-frame scenario described in the issue cannot produce the claimed stall or false-success in practice. The added test only passes/fails against a synthetic stream deliberately split at a byte boundary — a condition we've never observed in real usage.

If we decide the defensive hardening is worth having, we'll apply the minimal server-side buffering directly. Thanks anyway.

PR #6222, opened 2026-08-30
Implement LocalAI image editing via ref_images

TheGoodMorty:
### What

Implements `editImage` for the LocalAI image generation provider so reference
images are actually sent to LocalAI instead of being dropped. Editing uses
LocalAI's `ref_images` parameter on the same `/v1/images/generations` endpoint
as generation - LocalAI does not implement `/v1/images/edits`.

### Why

The current `editImage` override logs "LocalAI does not support image editing",
**drops all reference images**, and returns a prompt-only generation with a
notice claiming the references were ignored. LocalAI does support editing
(documented: https://localai.io/docs/features/image-generation/ - "Flux Kontext
with stable-diffusion.cpp"), so the override produces a misleading failure.

The existing comment in this file already anticipated this:
> LocalAI supports `ref_images` on /v1/images/generations for Flux Kontext
> models, which could enable img2img here without needing /v1/images/edits.
> Untested.

This PR implements and tests it.

### How

- Sends `POST {baseURL}/v1/images/generations` with `ref_images` populated.
- Each reference is the **raw base64** of the image bytes (no `data:` URI
  prefix). Two LocalAI specifics make this the only reliable format:
  1. `data:image/...;base64,...` strings fail LocalAI's base64 decoding and the
     reference is **silently dropped** (no client error, backend log shows
     `ref_images_count: 0`, and the model generates from the prompt only).
  2. Non-public `http(s)://` URLs (localhost / private ranges) are rejected by
     LocalAI's SSRF validation, so URL references can't be used for locally
     hosted servers.
- Raw `fetch` mirrors the base class's default `editImage` implementation
  (same response normalization: `b64_json` or `url` -> Buffer).
- Honors `size` / `IMAGE_GEN_SIZE_PREF`; sends `Authorization` only when an
  API key is configured; keeps telemetry parity with the base class.

### Testing

Verified end-to-end against **LocalAI v4.9.0** (binary install, no Docker) with
the `flux.1-kontext-dev` model on the `stablediffusion-ggml` backend:

- Before: attach a reference image + `/img` edit prompt -> reference dropped
  (`ref_images_count: 0` in LocalAI's backend log) -> fresh generation returned
  with the "reference images were ignored" notice.
- After: same request -> `ref_images_count: 1` in the backend log, and the
  returned image is an actual edit of the reference.


timothycarambat:
And what happens when you try to pass edits to a non-flux-context model? Does it crash and how does it crash?

The implementation does not guard on model name or ability. it would be great if there was a check for `image-editing` for capability detection because otherwise someone is going to pass in refs for image generation and instead of seeing a "references dropped" reply with their image they will just get a bad response that did not use their image.

We do this on Ollama, who do not allow edits, so at least users do not sit and wonder why their image generation did not take into account reference images.

Also, all PRs need an associated issue for tracking: https://github.com/Mintplex-Labs/anything-llm/blob/master/CONTRIBUTING.md#before-you-start

TheGoodMorty:
my bad, I'm very very new to GitHub and this is my first attempt to contribute to anything.
It does not crash, it just seems to give back a bad image based on the input image. I'll change it so that editing is checked first. When it's not possible, the image will be dropped, and a message stating the reference image was dropped will appear.
Unless it would be better to simply drop the whole generation request, I feel like if I wanted an edit and it proceeded to generate an image without the reference, I'd rather have just cancelled the attempt and change the model myself or figure out how to get the image I want with prompt alone. But if you want it consistent with the Ollama behavior then that's fine.

timothycarambat:
Keeping in mind an "edit" can also just be a new generation with some random references in it - that still is technically an "edit".

Eg: I want to combine two existing images + a prompt to drive output. That is, according to API specs, an edit - so in cases like this with Ollama [we simply tell the user the references were dropped.](https://github.com/Mintplex-Labs/anything-llm/blob/f2b1b13c749b54e1635c7d1f5d2eb44ffc9dd90f/server/utils/ImageGenerators/ollama/index.js#L23-L32)

That being said, I do think simply aborting the image here makes the most sense so that you are not wasting compute on an image generation that is bound to be bad since it didnt take into account references.

So for this, we can actually gate based on the `process.env.IMAGE_GEN_MODEL_PREF` since **only** `flux.1-kontext-dev` supports editing with `ref_images`.

Maybe we do the same here for ollama as well just to prevent the illusion of the process working only to find out post-gen you wasted the time waiting.

PR #6241, opened 2026-08-31
fix: mobile reset-chat clears the whole workspace instead of the named thread

### LINKED ISSUE #6239: [BUG]: mobile reset-chat clears every chat in the workspace instead of the named thread

All versions

### What happened?

`reset-chat` from a paired mobile device marks every chat in the workspace as excluded, not just the thread that was asked for.

`server/endpoints/mobile/utils/index.js`:

```js
const threadId = threadSlug
  ? await prisma.workspace_threads.findFirst({
      where: {
        workspace_id: workspace.id,
        slug: String(threadSlug),
        ...(user ? { user_id: user.id } : {}),
      },
    })?.id
  : null;
```

`await` binds looser than member access, so this parses as `await (findFirst({...})?.id)`. `findFirst` returns a Promise, `.id` on a Promise is `undefined`, and `await undefined` is `undefined`. `threadId` is always `undefined` whenever `threadSlug` is supplied.

That value goes straight into the where clause:

```js
await WorkspaceChats.markThreadHistoryInvalidV2({
  workspaceId: workspace.id,
  ...(user ? { user_id: user.id } : {}),
  thread_id: threadId, // if threadId is null, this will reset the default thread.
});
```

`markThreadHistoryInvalidV2` hands the object to `prisma.workspace_chats.updateMany` unchanged, and Prisma drops a `where` field whose value is `undefined`. The `thread_id` filter disappears, so the update matches every chat in the workspace.

The inline comment shows the intent was `null`. `null` filters to the default thread. `undefined` removes the filter entirely, which is the whole difference.

Against a real sqlite database, with one chat in thread A, one in thread B, and one in the default thread, then resetting thread B:

```
BEFORE:
   ws=1 thread=1 include=true
   ws=1 thread=2 include=true
   ws=1 thread=null include=true

threadId computed by the current code = undefined

AFTER reset-chat on thread-b:
   ws=1 thread=1 include=false
   ws=1 thread=2 include=false
   ws=1 thread=null include=false
```

All three are gone. With the parentheses corrected the same request computes `threadId = 2` and clears only that thread.

Steps:
1. Pair and approve a mobile device.
2. In a workspace, create a thread and send messages in it, and send messages in the default thread.
3. `POST /api/mobile/send/reset-chat` with `{"workspaceSlug":"<ws>","threadSlug":"<thread>"}`.
4. Every chat in that workspace disappears from the UI, not only the named thread's.

Other workspaces are unaffected, since `workspaceId` is still in the clause. In multi-user mode it clears that user's chats in that workspace.

There is a second path to the same data loss in the same block. Because `null` is how this handler addresses the default thread, a `threadSlug` that matches no thread must not fall through to `null` either, or an unknown slug silently clears the default thread and returns `{"success": true}`. Only an absent slug should mean the default thread.

The same `await ... ?.id` expression shape appears once more, at `server/endpoints/system.js` in `/system/update-password`. I will report that separately since the consequence there is different.

marmar9615-cloud:
### Pull Request Type

- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6239

### Description

`reset-chat` from a paired mobile device clears every chat in the workspace instead of the one thread it was asked for.

The thread is resolved like this:

```js
const threadId = threadSlug
  ? await prisma.workspace_threads.findFirst({ ... })?.id
  : null;
```

`await` binds looser than member access, so this is `await (findFirst({...})?.id)`. `findFirst` returns a Promise, `.id` on a Promise is `undefined`, and `await undefined` is `undefined`. Whenever `threadSlug` is supplied, `threadId` is `undefined`.

`markThreadHistoryInvalidV2` passes its argument straight to `prisma.workspace_chats.updateMany`, and Prisma drops a `where` field whose value is `undefined`. The `thread_id` filter is removed and the update matches the whole workspace.

The existing inline comment, `if threadId is null, this will reset the default thread`, shows `null` was intended. `null` filters to the default thread; `undefined` filters to nothing. That is the entire bug.

Adding the parentheses is the fix. `server/endpoints/workspaces.js` already uses the correct form, `(await WorkspaceThread.get({...}))?.id ?? null`.

### Additional Information

Reproduced against a real sqlite database with one chat in thread A, one in thread B, and one in the default thread, then resetting thread B:

| | thread A | thread B | default |
|---|---|---|---|
| before | include=true | include=true | include=true |
| after, current code | include=false | **include=false** | **include=false** |
| after, with this fix | include=true | include=false | include=true |

Other workspaces are untouched either way, because `workspaceId` stays in the clause.

Scope of the change: one expression in one file. I checked the other two places in the repo where `await` is followed by an optional-chained call. `server/utils/vectorDbProviders/qdrant/index.js` is correctly parenthesized already. `server/endpoints/system.js` in `/system/update-password` has the same defect with a different consequence, and I am keeping that separate rather than mixing two fixes here.

Adding the parentheses alone was not enough. `null` is how this handler addresses the default thread, so letting an unresolved slug fall through to `null` just trades one silent data loss for a smaller one: an unknown slug would clear the default thread and still answer `{"success": true}`. The handler now resolves a supplied slug and returns `400 Thread not found` when it does not exist, leaving an absent slug as the only way to reach the default thread.

I used 400 because every error in this command handler is a 400, including the `Workspace not found` immediately above it. The public thread endpoints in `server/endpoints/api/workspaceThread/index.js` use 404 for this, so say the word if you would rather this match those.

The sibling `stream-chat` command in this file does treat an unresolved slug as the default thread. That is harmless there because it writes a message; doing the same here erases history the caller never named.

Noticed while tracing the sibling resets, not changed here: this handler does not pass `api_session_id: null`, but the Telegram reset does, and the deprecated `markHistoryInvalid` carries the comment `API session chats also live on the default thread - a UI reset must not clear them`. So a mobile reset that falls through to the default thread also clears API-session chats. That is pre-existing on both branches of the ternary and unaffected by this change, so I have left it alone rather than mix two fixes. Happy to open it separately if you want it.

### Developer Validations

### Test

`server/__tests__/endpoints/mobile/utils/index.test.js` asserts `markThreadHistoryInvalidV2` receives `thread_id: 42` exactly, and that omitting `threadSlug` still sends `thread_id: null`.

Three tests, written to fail the near misses as well as the current code:

| implementation | result |
| --- | --- |
| current code | 2 of 3 fail |
| parentheses added, unresolved slug falls through to `null` | 1 of 3 fails |
| this change | 3 pass |

The middle row is the one that matters. It looks correct and clears the reported bug, but it still erases the default thread on an unknown slug, so the suite has to reject it.

Full suite: 49 suites, 655 tests, all passing. Baseline on master is 48 suites, 652 tests. `cd server && yarn lint:check` clean.


timothycarambat:
Thanks for the thorough investigation and writeup — the operator-precedence bug is real and the parentheses fix is correct.

One change requested before merge: we need to drop the `400 Thread not found` branch. Returning an error on an unresolved slug will break existing mobile clients that may call `reset-chat` with a stale or out-of-sync `threadSlug` — for them a hard 400 on a reset action is a bigger problem than the fall-through. An unresolved slug should resolve to `threadId = null` (default thread), matching how the sibling `stream-chat` command in this same file already handles it.

I understand that's the exact "middle row" your test suite was written to reject — this is an intentional product decision to keep the mobile command handlers consistent and non-breaking, not an oversight since mobile is not updated often.

marmar9615-cloud:
Understood, and thank you for spelling out the reason. Stale or out-of-sync `threadSlug` values from clients you cannot update is not something I could have seen from the code, and a hard 400 on a reset is clearly the worse failure for those users. Your call, and it is the right one with that context.

I see you have already pushed `57c8c9c5`. I verified it rather than assume: 3 of 3 pass on your version, and against the original code the suite still fails 2 of 3, so the precedence bug itself stays guarded.

Worth saying that your test is better than mine on the point that actually matters. Mine pinned a 400 policy; yours pins `thread_id: null` against `undefined` on both the resolved and unresolved paths, which is the distinction the bug turns on, since Prisma drops an `undefined` where field and widens the update to the whole workspace. The policy was the part I should not have encoded in a test.

I have not pushed anything to this branch and will not without asking.

PR #6245, opened 2026-08-31
fix: report a failed password update instead of always answering success

### LINKED ISSUE #6244: [BUG]: /system/update-password always reports success, and exits the process when the update rejects

All versions

### What happened?

`POST /system/update-password` always answers `{"success": true}`, and when the underlying update rejects, the server process exits instead of returning an error.

`server/endpoints/system.js`:

```js
error = await updateENV(
  {
    AuthToken: newPassword,
    JWTSecret: v4(),
  },
  true
)?.error;

response.status(200).json({ success: !error, error });
```

`await` binds looser than member access, so this parses as `await (updateENV(...)?.error)`. `updateENV` is `async`, so `.error` is read off the pending Promise rather than the resolved value, and is always `undefined`. `!error` is therefore always `true`.

The more damaging half is that the Promise is never actually awaited, so a rejection does not reach the handler's `catch`. It surfaces as an `unhandledRejection`, and `@ladjs/graceful` installs a listener for that:

```js
process.on('unhandledRejection', (err) => {
  // we don't want to log here, we want to throw the error
  // so that processes exit or bubble up to middleware error handling
  throw err;
});
```

`BackgroundWorkers` calls `graceful.listen()`, and both `bootHTTP` and `bootSSL` start it, so this is installed on a normal server. The re-thrown error lands in the sibling `uncaughtException` handler, which logs and calls `process.exit(1)`.

`updateENV` calls `dumpENV()` in production, which writes the `.env` file, so an unwritable storage directory is enough to turn a password change into a process exit rather than a 500.

For completeness on the first half: a truthy `error` is not reachable from the shipped UI today, because both call sites pre-validate with the same character regex the server enforces, and with `force = true` that check is the only thing that sets `error` for these keys. It is reachable from any non-browser client, and onboarding shipped without that regex until #5037, so the guard is one drift away from mattering again.

`git log` shows this was a regression: `bfc9a96e` replaced a destructured result with the inline optional read.

For the always-succeeds half, with any client that does not pre-validate:

```
POST /api/system/update-password
{"usePassword": true, "newPassword": "my.pass"}

-> 200 {"success": true}
```

`.` is outside the allowed character set, so `updateENV` returns an error and never assigns `process.env.AUTH_TOKEN`. The caller is told it worked. The Security settings page then shows "Your page will refresh in a few seconds", clears `AUTH_USER`, `AUTH_TOKEN` and `AUTH_TIMESTAMP` from local storage and reloads, so the user is signed out while the old password is still in force.

For the exit half, make the storage directory unwritable so `dumpENV` throws, then post a valid password change. The request never returns a 500; the process exits.

### Related, and not fixed by correcting the above

`updateENV` filters out any value containing `******`:

```js
const ENV_KEYS = Object.keys(newENVs).filter(
  (key) => validKeys.includes(key) && !newENVs[key].includes("******")
);
```

`*` is inside the allowed password character set, so a password containing six or more consecutive asterisks is silently dropped with no error. `JWTSecret` still rotates, so the user is signed out and the old password still works. That is a separate defect and correcting the `await` above does not address it. Happy to open it separately.

marmar9615-cloud:
### Pull Request Type

- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6244

### Description

`/system/update-password` reads its result like this:

```js
error = await updateENV({ AuthToken: newPassword, JWTSecret: v4() }, true)?.error;
```

`await` binds looser than member access, so this is `await (updateENV(...)?.error)`. `updateENV` is `async`, so `.error` comes off the pending Promise and is always `undefined`, and `{ success: !error }` is always `true`.

The half that matters on a running server is the rejection. Because the Promise is never awaited, a rejection does not reach the handler's `catch`. It becomes an `unhandledRejection`, and `@ladjs/graceful` has a listener for that which deliberately re-throws so the process exits. `BackgroundWorkers` calls `graceful.listen()` and both `bootHTTP` and `bootSSL` start it, so it is installed on a normal server. The re-thrown error hits the sibling `uncaughtException` handler, which logs and calls `process.exit(1)`.

`updateENV` calls `dumpENV()` in production, which writes `.env`, so an unwritable storage directory turns a password change into a process exit rather than a 500.

Assigning the awaited result to a local and reading `.error` from that fixes both halves.

### Additional Information

This is a regression rather than original behaviour. `bfc9a96e` replaced a destructured result with the inline optional read.

Being precise about reach, since it affects how you may want to prioritise this: the always-succeeds half is **not** reachable from the shipped UI today. Both call sites pre-validate with the same character regex the server enforces, and with `force = true` that regex check is the only thing that sets `error` for these keys. It is reachable from any non-browser client, and onboarding shipped without that regex until #5037, so the server-side guard is one drift away from mattering again. The rejection half is reachable regardless of client.

One thing this does **not** fix, so the success flag still should not be read as proof the password changed: `updateENV` drops any value containing `******`, and `*` is inside the allowed password character set, so a password with six or more consecutive asterisks is silently ignored with no error while `JWTSecret` still rotates. That is a separate defect. I have described it in the issue and can open it separately if you want it.

I kept `update?.error` rather than destructuring. `updateENV` never resolves nullish, so the optional read is belt and braces; sibling handlers in this file destructure instead, and I am happy to match that if you prefer.

### Developer Validations

### Test

`server/__tests__/endpoints/system.test.js` captures the registered handler and drives it directly. It covers the reported failure being reported, the success path still reporting success, and a rejection reaching the handler's `catch` and answering 500.

All three fail against current code, and they fail cleanly:

```
✕ reports the failure when updateENV rejects the new password
✕ reports success when updateENV accepts the new password
✕ handles a rejected updateENV inside its own catch instead of leaking it
Tests: 3 failed, 3 total
```

That "cleanly" is deliberate. The rejection is pre-handled in the test, because without it, running this suite against the unfixed code terminates the jest worker with an unhandled rejection and reports no results at all, which would make a future revert look like a crash rather than a failed assertion.

Full suite: 49 suites, 655 tests, all passing, against a 48 suite and 652 test baseline on master. `cd server && yarn lint:check` clean.


timothycarambat:
Worth noting this was only possible by hitting the frontend API directly via curl and not from the UI - which did always report the proper error state and had form validations preventing this from being possible by users anyway.

Added simply to normalize and harden the requirement serverside.

PR #6249, opened 2026-08-31
Add variable categories to agent flow variables

### LINKED ISSUE #5467: [FEAT]: Clear variable categories in Agent Flows
### Problem/Use Case

While working with Agent Flows, I ran into difficulties managing variables due to the lack of clear distinctions between different variable types. Because of that, the llm model tend to not use the workflow correctly and modify variables that shouldn’t.

### Proposed Solution

I think, it would simplify the development of Agent Flow by spiting the variables in categories. The categories I see so far are:

- **Required variables**: Variables that must be set by the LLM in order to execute the workflow.
- **Optional variables**: Variables that have default values but can be overridden by the LLM.
- **Internal variables**: Variables used only within the workflow. These should not be exposed to or modified by the LLM.
- **Configuration variables**: Variables defined by the user in the Agent Flow settings. These are fixed at runtime and should not be visible to the LLM.

### Benefits

- Improves readability and maintainability of workflows
- Reduces ambiguity when designing flows
- Improve the stability/predictability of the workflows
- Provides clearer separation of responsibilities between user configuration, internal logic, and LLM interaction

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #5467

### Description

- add a category to each flow variable: required, optional, or static
- required variables must be provided by the LLM to run the flow (enforced in the tool schema and at execution time)
- optional variables keep their default value but can be overridden by the LLM
- static variables are fixed at design time and are never exposed to or modifiable by the LLM
- add a per-variable description input that is passed to the LLM in the tool schema
- LLM-provided args for non-exposed variables are dropped before flow execution
- existing flows are unaffected: variables without a category behave as optional (the previous behavior)

### Visuals (if applicable)

Light mode UI:
Screenshot 2026-08-31 at 6 22 30 PM

Dark mode UI:
Screenshot 2026-08-31 at 6 22 44 PM

### Additional Information

- No changes to community hub needed since we send over the entire json and store in community hub when uploading

### Developer Validations


timothycarambat (changes requested):
- Can we get some UI screenshots + validation in light mode
- Since we are doing some logic changes here can we confirm:
   - no impact to existing flows that do not have these configs
   - Tests that validate the expected path and test adversarial configs to ensure that the filtering on vars works as expected and we dont break it.

shatfield4:
> * Can we get some UI screenshots + validation in light mode
> * Since we are doing some logic changes here can we confirm:
>   
>   * no impact to existing flows that do not have these configs
>   * Tests that validate the expected path and test adversarial configs to ensure that the filtering on vars works as expected and we dont break it.

- Added screenshots for UI for both dark and light modes (only difference is in the variable options dropdown where they can select variably category)
- Added tests to ensure previous flows still work (all variables without a variable category fallback to optional)

PR #6252, opened 2026-08-31
fix: stop citation cleanup mangling scraped code and numbered links

### LINKED ISSUE #6254: [BUG]: scraping deletes bracketed indexes from code and eats numbered link text

All versions

### What happened?

Scraping a link stores corrupted text when the page contains bracketed numbers in code, numeric link or image labels, or bracketed URL segments.

`collector/processLink/helpers/htmlToMarkdown.js` runs this over the whole converted document to strip citation markers:

```js
markdown = markdown.replace(/\[[\d]+\]/g, "");
```

Turndown escapes brackets in prose text to `\[1\]`, which that pattern cannot match. The confirmed matches instead occur in generated Markdown contexts that retain literal brackets—including code, numeric link/image labels, and bracketed URL segments—and the cleanup corrupts all of them.

**Code.** Turndown leaves brackets literal inside code spans and fenced blocks:

| input | stored |
| --- | --- |
| `<pre><code>const second = items[1];</code></pre>` | `const second = items;` |
| `<p>Use <code>arr[0]</code></p>` | ``Use `arr` `` |

The failure is silent. The link imports normally and the corrupted snippet is what the model later cites.

**Numbered links.** `compactLinks` builds an unescaped `[text](href)`, so a link whose visible text is a number loses its text and strands the URL:

```
<p>See note<sup><a href="#fn1">1</a></sup> end.</p>   ->  See note(#fn1) end.
<a href="/page/1">1</a>                               ->  (https://example.com/page/1)
```

Footnote markers and pagination controls are exactly the numbered links a scraped article tends to have.

**Images and URLs.** The same replacement damages generated Markdown outside link labels:

```
<img src="/diagram.png" alt="1">                     ->  !(https://example.com/diagram.png)
<a href="https://example.com/?a[1]=x">query</a>       ->  [query](https://example.com/?a=x)
```

It likewise removes `[1]` from an image URL such as `image[1].png`.

Reference superscripts are removed before conversion by `stripCitations`, which drops `sup.reference` and `sup.noprint`, so citation handling does not depend on this regex.

Expected: citation cleanup should preserve source code and valid generated Markdown, including link text, image alt text, and URLs.

1. Add a website link containing an HTML code block such as `<pre><code>const second = items[1];</code></pre>`
2. Open the stored document
3. The index is gone from the code

The same path can be reproduced with a numbered link, numeric image alt text, or a URL containing a bracketed numeric segment.

marmar9615-cloud:
### Pull Request Type

- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6254

### Description

`htmlToMarkdown` ran this over the whole converted document to strip citation markers:

```js
markdown = markdown.replace(/\[[\d]+\]/g, "");
```

It never cleanly removed one. Turndown escapes brackets in prose text to `\[1\]`, which that pattern cannot match, so every context it did match was one it should have left alone.

**Inside code**, Turndown leaves brackets literal, so the regex ate them:

| input | before | after |
| --- | --- | --- |
| `<pre><code>const second = items[1];</code></pre>` | ``` const second = items; ``` | ``` const second = items[1]; ``` |
| `<p>Use <code>arr[0]</code></p>` | ``Use `arr` `` | ``Use `arr[0]` `` |

Scraping a page that documents an array index stored the wrong snippet, and the failure is silent: the link imports fine, and the corrupted text is what the model later cites.

**In markdown syntax**, Turndown and the converter emit unescaped brackets, and the pattern ate those too. `compactLinks` builds an unescaped `[text](href)`, so a link whose visible text is a number lost its text and stranded the URL:

```
<p>See note<sup><a href="#fn1">1</a></sup> end.</p>
  before: See note(#fn1) end.
  after:  See note[1](#fn1) end.

<a href="/page/1">1</a>
  before: (https://example.com/page/1)
  after:  [1](https://example.com/page/1)
```

Footnote markers and pagination controls are exactly the numbered links a scraped article tends to have.

Two further cases, same cause:

```
<p><img src="/a.png" alt="1"></p>
  before: !(/a.png)                  after: ![1](/a.png)

<p><a href="/docs/[1]/page">link</a></p>
  before: [link](/docs//page)        after: [link](/docs/[1]/page)
```

The second corrupts the URL itself rather than the text around it.

Reference superscripts are removed before conversion by `stripCitations`, which drops `sup.reference` and `sup.noprint`, so citation handling does not depend on this regex.

### Additional Information

The sibling `\[#cite...\]` and `\[edit\]` cleanups on the next two lines have the same shape and the same problem. `<pre><code>menu[edit] = true;</code></pre>` still stores `menu = true;` after this change. I left them alone to keep this reviewable, and can send them separately.

I considered scoping the regex to non-code regions instead of deleting it. That means writing a fence and span parser inside a cleanup line, and it still would not fix the numbered-link case, since those are real link syntax. Since every confirmed match corrupted source content or generated Markdown syntax, deleting it is the smaller and more honest change.

### Developer Validations

### Test

Six tests in a new `collector/__tests__/processLink/helpers/htmlToMarkdown.test.js`.

Three states rather than two:

| implementation | result |
| --- | --- |
| this change | 6 pass |
| current code | 5 fail, the citation test passes |
| narrower `/\[\d+\](?!\()/g` | 3 fail, 3 pass |

The citation test passing under current code is deliberate. It is a control: it shows `stripCitations` already handles reference superscripts independently, so the deletion is not what would break citation removal. It would fail the wrong fix that removed `stripCitations` instead.

Full suite: 51 suites, 685 tests, all passing, against a 50 suite and 679 test baseline on master measured in the same tree. `cd collector && yarn lint:check` clean.

### DIFF
--- a/collector/__tests__/processLink/helpers/htmlToMarkdown.test.js
+++ b/collector/__tests__/processLink/helpers/htmlToMarkdown.test.js
@@ -1,0 +1,38 @@
+/* eslint-env jest, node */
+process.env.STORAGE_DIR = "test-storage"; // needed for tests to run
+const {
+  htmlToMarkdown,
+} = require("../../../processLink/helpers/htmlToMarkdown");
+
+describe("htmlToMarkdown citation stripping", () => {
+  it("keeps a numeric index inside a fenced code block", async () => {
+    const markdown = await htmlToMarkdown(
+      "<pre><code>const second = items[1];</code></pre>"
+    );
+    expect(markdown).toContain("items[1]");
+  });
+
+  it("keeps a numeric index inside inline code", async () => {
+    const markdown = await htmlToMarkdown(
+      "<p>Use <code>arr[0]</code> to read the first element.</p>"
+    );
+    expect(markdown).toContain("`arr[0]`");
+  });
+
+  it("keeps a footnote link whose visible text is a number", async () => {
+    const markdown = await htmlToMarkdown(
+      '<p>See note<sup><a href="#fn1">1</a></sup> end.</p>',
+      "https://example.com"
+    );
+    // compactLinks emits an unescaped [1](#fn1), so the citation regex ate the
+    // link text and left a dangling (#fn1) behind.
+    expect(markdown).toBe("See note[1](#fn1) end.");
+  });
+
+  it("still removes a reference superscript from prose", async () => {
+    const markdown = await htmlToMarkdown(
+      '<p>As shown in the study<sup class="reference">[1]</sup>, results vary.</p>'
+    );
+    expect(markdown).toBe("As shown in the study, results vary.");
+  });
+});
--- a/collector/__tests__/processLink/helpers/htmlToMarkdown.test.js
+++ b/collector/__tests__/processLink/helpers/htmlToMarkdown.test.js
@@ -4,51 +4,220 @@ const {
   htmlToMarkdown,
 } = require("../../../processLink/helpers/htmlToMarkdown");
 
-describe("htmlToMarkdown citation stripping", () => {
-  it("keeps a numeric index inside a fenced code block", async () => {
-    const markdown = await htmlToMarkdown(
-      "<pre><code>const second = items[1];</code></pre>"
-    );
-    expect(markdown).toContain("items[1]");
+describe("htmlToMarkdown", () => {
+  describe("input handling", () => {
+    it("returns an empty string for null, undefined, and non-string input", async () => {
+      expect(await htmlToMarkdown(null)).toBe("");
+      expect(await htmlToMarkdown(undefined)).toBe("");
+      expect(await htmlToMarkdown(12345)).toBe("");
+      expect(await htmlToMarkdown("")).toBe("");
+    });
+
+    it("trims surrounding whitespace from the output", async () => {
+      expect(await htmlToMarkdown("<div>  <p>x</p>  </div>")).toBe("x");
+    });
   });
 
-  it("keeps a numeric index inside inline code", async () => {
-    const markdown = await htmlToMarkdown(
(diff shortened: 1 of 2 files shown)


marmar9615-cloud:
That `run-script` failure is not coming from this branch. The failing test is `server/__tests__/utils/helpers/modelPricing/index.test.js`, and this PR touches two files, both under `collector/`.

I checked out `2a55c00b` and ran it locally:

- that test file alone: 41 passed
- the full suite: 51 suites, 706 tests, all passing, which is the same 706 the CI run counted

It looks like a fixed-sleep race that predates this branch. The `ModelPricing` constructor starts `#refresh()` without awaiting it (`server/utils/helpers/modelPricing/index.js:151-159`), and the test helper waits a flat 25 ms before asserting the refresh has landed:

```js
async function flushRefresh() {
  await new Promise((resolve) => setTimeout(resolve, 25));
}
```

The failing assertion is `expect(pricing.isCacheStale).toBe(false)` immediately after writing `garbage-timestamp` into `.cached_at`. If the background refresh has not rewritten that file inside 25 ms, `Number("garbage-timestamp")` is still `NaN`, `#isCacheStale()` returns `true` on the `Number.isFinite` guard, and you get the reported `Expected: false, Received: true`. Lowering that wait to 1 ms reproduces failures in the file deterministically on my machine, so the file depends on the sleep being long enough rather than on a signal.

The run's own warnings point the same way: "Cannot log after tests are done ... Attempted to log `⚡Pre-cached model pricing data`" is that background promise resolving after teardown, and "A worker process has failed to exit gracefully".

A re-run should clear it, and I do not have rights to trigger one. I am happy to send the deterministic-wait fix as its own change with an issue attached if you want it, but putting a `server/__tests__` edit inside a collector PR seemed like the wrong call.

PR #6253, opened 2026-08-31
fix: score Lance chunks at or past orthogonal as 0, not a perfect match

### LINKED ISSUE #6255: [BUG]: LanceDB scores chunks at or past orthogonal as a perfect match, bypassing the similarity threshold

All versions

### What happened?

With LanceDB, the workspace similarity threshold can admit chunks at or past orthogonal and show them as perfect matches.

Both Lance search paths query with `.distanceType("cosine")`, so `_distance` runs `[0, 2]`: 0 identical, 1 orthogonal, 2 opposite. `distanceToSimilarity` in `server/utils/vectorDbProviders/lance/index.js` returns `1`, the top of the scale, for anything at or above 1.0.

Running the real class:

| cosine distance | similarity | clears a 0.75 threshold? |
| --- | --- | --- |
| 0.25 | 0.75 | yes |
| 0.5 | 0.5 | no |
| 0.99 | 0.01 | no |
| 1.0 | **1** | **yes** |
| 1.5 | **1** | **yes** |
| 2.0 | **1** | **yes** |

The mapping is non-monotonic: a chunk 0.99 away is dropped by a 0.75 threshold while a chunk 1.5 away, further from the query, clears every selectable threshold.

`similarityThreshold` is applied to this value in both paths, so any returned candidate with `_distance >= 1` can pass the threshold and reach the model as context. In the standard path the same value is the citation `score`, so it is displayed as a perfect match. The reranked path normally reports `rerank_score`, so there the effect is admission rather than the displayed number.

This was already the stated intent. #2986 says:

> Also fixed `if (distance >= 1.0) return 0;` because if distance should always be a value between 0 and 1 and if distance is greater than 1 then similarity should be 0

Its diff changed the `distance <= 0` branch and left this one returning `1`, so that half never landed.

Chroma and PGVector carry the same branch while also configuring cosine distance. I have kept this report to Lance so a fix can be verified per engine.

Expected: similarity should not rise as cosine distance increases; at or past orthogonal it should floor at 0 rather than pass a positive workspace threshold.

1. Use LanceDB, the default vector database
2. Set a workspace similarity threshold, for example 0.75
3. Ask something unrelated to the embedded documents
4. If the returned candidate set contains a chunk with `_distance >= 1`, standard search admits it and cites it at score 1

The condition can be reproduced deterministically with controlled vectors; an unrelated natural-language query alone does not guarantee a candidate at that distance.

marmar9615-cloud:
### Pull Request Type

- [x] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6255

### Description

Both Lance search paths query with `.distanceType("cosine")`, so `_distance` runs `[0, 2]`: 0 identical, 1 orthogonal, 2 opposite. `distanceToSimilarity` returned `1` for anything at or above 1.0, which is the top of the scale.

That makes the mapping non-monotonic. Running the real class:

| cosine distance | similarity | clears a 0.75 threshold? |
| --- | --- | --- |
| 0.25 | 0.75 | yes |
| 0.5 | 0.5 | no |
| 0.99 | 0.01 | no |
| 1.0 | **1** | **yes** |
| 1.5 | **1** | **yes** |
| 2.0 | **1** | **yes** |

A chunk 0.99 away is dropped by a 0.75 threshold while a chunk 1.5 away, further from the query, clears every threshold. A chunk pointing the opposite way scores a flawless 1.

`similarityThreshold` is applied to this value in both paths, so whenever the returned candidate set contains a row at `_distance >= 1`, that row is admitted no matter how strict the threshold is. In the standard path the same value is also the citation `score`, so it is displayed as a perfect match. The reranked path normally reports `rerank_score` instead, so there the effect is admission rather than the displayed number.

This was already the intended behaviour. #2986 says:

> Also fixed `if (distance >= 1.0) return 0;` because if distance should always be a value between 0 and 1 and if distance is greater than 1 then similarity should be 0

Its diff changed the `distance <= 0` branch but left this one returning `1`, so the stated half never landed.

### Additional Information

Flooring at 0 is the minimal monotone extension of the existing `1 - distance` mapping. Every score below orthogonal is unchanged, and the range stays `[0, 1]`, which is what the workspace threshold and the citation UI expect.

I deliberately did not rewrite the mapping to something like `1 - distance / 2` to use the full cosine range. That would change every existing score below orthogonal and shift what a saved workspace threshold means, which is a behaviour change rather than a bug fix.

Chroma and PGVector carry the same `return 1` branch while also configuring cosine distance. I left them out so this stays one engine with its own tests, and can send them separately.

### Developer Validations

### Test

Four tests in a new `server/__tests__/utils/vectorDbProviders/lance/index.test.js`.

One of them drives `similarityResponse` itself with a stub collection rather than asserting arithmetic, so it pins the behaviour a workspace threshold is supposed to produce: given rows at distance 0.2 and 1.5 and a 0.75 threshold, only the near row survives.

| implementation | result |
| --- | --- |
| this change | 4 pass |
| current code | 3 fail |

The fourth test, that near chunks still score as before, passes in both states on purpose. It is a control showing this does not touch scoring below orthogonal, not a guard for the fix.

Full suite: 51 suites, 683 tests, all passing, against a 50 suite and 679 test baseline on master measured in the same tree. `cd server && yarn lint:check` clean.


marmar9615-cloud:
Thanks for taking this, and for carrying the same floor across to Chroma and PGVector.

One thing I wanted to ask about rather than assume. The follow-up commit dropped `server/__tests__/utils/vectorDbProviders/lance/index.test.js`, so `distanceToSimilarity` is now unguarded in all three providers. I restored it verbatim onto current master to check whether it had gone stale against your version, and it passes 4 of 4, so it does not look like it was failing.

If dropping it was deliberate, that is fine and I will leave it alone. If it just went along with the squash, I am happy to send a small test-only PR that restores it and adds the same coverage for Chroma and PGVector. Happy to file an issue for it first either way.

timothycarambat:
If we add a test, we want to add a test to the full module. Given that class shares a ton of base methods with others I decided instead of making a dozen tests for each class I decided its best we not do it. The classes are stable anyway and are due for a later refactor by core team.

marmar9615-cloud:
Understood, that makes sense and I will leave it alone. Good to know the vector DB classes are due for a refactor before I touch that area again.

PR #6272, opened 2026-09-02
feat: describe uploaded images with a vision model so they are queryable

### LINKED ISSUE #3175: [FEAT]: RAG for images files

I would like to suggest to allow to upload into RAG system, the JPG/PNG/etc. files, because if I use a multimodal embedder model like MiniCPM-o-2_6, I saw that it can manage file like mp4 or mp3, it can describe, reading from my RAG, the video end the audio. I think it could be useful to allow the embedding of images files too.
My 2 cents ;-)

ArslanRasheed60:
resolves #3175

## What

Uploading an image only ever ran OCR. `collector/processSingleFile/convert/asImage.js` called `OCRLoader().ocrImage()` and, when OCR came back empty, deleted the file and returned `No text content found in <filename>.` — so a diagram, screenshot or photo with no printed text was **rejected outright** rather than ingested.

This adds an optional vision-describe step at ingest, which is the shape you described on the issue:

> Uploading an image, having it described and embedding that then later asking queries about **that** makes more sense for at least our use case.

## Implementation

- **`collector/utils/VisionProviders/`** — a new sibling directory modelled on `WhisperProviders/`, with `OpenAiVision` and `GenericOpenAiVision` (any OpenAI-compatible endpoint: Ollama, LM Studio, vLLM, LocalAI). Same class shape, constructor-options style and `{content, error}` contract as the Whisper providers. `openai` was already a collector dependency.
- **`asImage.js`** — a `VISION_PROVIDERS` map mirroring `asAudio.js`'s `WHISPER_PROVIDERS`. OCR still runs first and is unchanged.
- **Provenance.** The generated text is embedded behind a literal `[Machine-generated image description]` line and reflected in `docSource`/`description`, so model-written prose is never silently read back as OCR ground truth.
- **ENV: four keys.** `VISION_PROVIDER` (`none`/`openai`/`generic-openai`, default `none`) plus `VISION_GENERIC_OPEN_AI_{BASE_URL,API_KEY,MODEL}`. The OpenAI provider reuses the existing `OPEN_AI_KEY` rather than adding another. Config reaches the collector through the existing `#attachOptions()` channel.
- **Settings page** mirroring the transcription preference page, with the provider select and conditional fields.

### Behaviour when disabled

With no provider set, no provider object is constructed and ingestion is byte-for-byte what it is today, including the original failure message and the `trashFile()` call. A failing or unreachable vision endpoint logs and falls back to the OCR-only path — a vision outage cannot break image ingestion.

## Testing

`npx jest` from the repo root — **57 suites / 777 tests pass** (54/748 before this change). Lint clean in `collector/`, `server/` and `frontend/`; `yarn build` succeeds; `yarn translations:verify` passes.

Unit tests cover provider selection, the composed `pageContent` and its provenance marker, the empty-OCR-plus-description case that previously hard-failed, both-empty still failing with the original message, provider-throws falling back to OCR, and the `updateENV` validator. Network is mocked throughout.

Verified end-to-end on a running stack (server + collector + frontend) against a real vision model — Ollama's OpenAI-compatible endpoint with `moondream`:

- `VISION_PROVIDER` unset: a no-text PNG is still rejected with `No text content found in notext.png.`, and no provider is constructed.
- Enabled: the same file ingests, and `pageContent` carries the marker plus a correct description of the shapes in the image.
- Settings page saves and persists across reload; the masked API key round-trips without clobbering the stored value.
- Base URL pointed at a dead port: a text-bearing image still ingests via OCR, with the failure logged.
- Text-bearing image with vision on: OCR text and the marked description are both present and delimited.
- Non-image ingestion unaffected.

Two gaps I'd rather state than imply: the Docker/production env-file path wasn't exercised (in dev the config reaches the collector via the `#attachOptions()` payload, which is the intended route), and the hosted `openai` provider is unit-tested only — I had no key to run it live.

The model for the OpenAI provider is currently pinned the way `OpenAiWhisper` pins `whisper-1`; happy to expose it as a fifth key if you'd prefer that configurable.

### DIFF
--- a/collector/__tests__/processSingleFile/convert/asImage.test.js
+++ b/collector/__tests__/processSingleFile/convert/asImage.test.js
@@ -1,0 +1,226 @@
+/* eslint-env jest, node */
+process.env.STORAGE_DIR = "test-storage";
+
+const mockOcrImage = jest.fn();
+const mockOpenAiDescribe = jest.fn();
+const mockGenericDescribe = jest.fn();
+
+jest.mock("../../../utils/OCRLoader", () =>
+  jest.fn().mockImplementation(() => ({ ocrImage: mockOcrImage }))
+);
+jest.mock("../../../utils/VisionProviders/OpenAiVision", () => ({
+  OpenAiVision: jest
+    .fn()
+    .mockImplementation(() => ({ describeImage: mockOpenAiDescribe })),
+}));
+jest.mock("../../../utils/VisionProviders/GenericOpenAiVision", () => ({
+  GenericOpenAiVision: jest
+    .fn()
+    .mockImplementation(() => ({ describeImage: mockGenericDescribe })),
+}));
+jest.mock("../../../utils/files", () => ({
+  createdDate: jest.fn(() => "1/1/2026"),
+  trashFile: jest.fn(),
+  writeToServerDocuments: jest.fn(({ data }) => data),
+}));
+
+const OCRLoader = require("../../../utils/OCRLoader");
+const {
+  OpenAiVision,
+} = require("../../../utils/VisionProviders/OpenAiVision");
+const {
+  GenericOpenAiVision,
+} = require("../../../utils/VisionProviders/GenericOpenAiVision");
+const { trashFile } = require("../../../utils/files");
+const asImage = require("../../../processSingleFile/convert/asImage");
+
+const FULL_FILE_PATH = "/hotdir/diagram.png";
+const FILENAME = "diagram.png";
+const MARKER = "[Machine-generated image description]";
+
+function convert(options = {}, metadata = {}) {
+  return asImage({
+    fullFilePath: FULL_FILE_PATH,
+    filename: FILENAME,
+    options,
+    metadata,
+  });
+}
+
+describe("asImage", () => {
+  beforeEach(() => {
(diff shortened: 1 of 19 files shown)


ArslanRasheed60:
@timothycarambat when you have a moment — this implements the approach you described on #3175 (describe at ingest, embed the description) rather than image-similarity search.

The on-device OCR path is untouched: with no provider configured nothing new is constructed and ingestion behaves exactly as it does today, including the original rejection message. It's a separate `VisionProviders/` directory alongside `WhisperProviders/`, not a change to the OCR flow.

Two calls I'd like your read on: I kept it to four ENV keys (reusing `OPEN_AI_KEY` for the OpenAI provider) to avoid cloning the six `WHISPER_*` vars, and the OpenAI model is pinned the way `OpenAiWhisper` pins `whisper-1` — happy to make it configurable if you'd rather.

timothycarambat:
I am not sure what the issue was (3175 does not exist?) so I cannot recall the core bug was but in general it would be much easier to not take this approach and instead run a small pipeline that can check the current workspace/global LLM `getModelCapabilities` and check for vision - then if supported we could use the model to describe the image and store that as essentially a more capable CLIP/Caption model instead of it being **another** service vendor to config and manage.

PR #6275, opened 2026-09-02
fix(collector): treat a blank published value as absent, not as the epoch

### LINKED ISSUE #6274: [BUG]: A null published on /v1/document/raw-text dates the document to the epoch

Local development

### What happened?

`GET /v1/document/metadata-schema` documents the `published` key as `"epoch timestamp in ms | nullable"`. Sending it as `null` on `POST /v1/document/raw-text` dates the document to the Unix epoch instead of treating it as absent.

`collector/processRawText/index.js:42`

```js
published: ({ published }) => {
  if (isNaN(Number(published))) return new Date().toLocaleString();
  return new Date(Number(published)).toLocaleString();
},
```

`Number(null)` is `0`, as are `Number("")` and `Number(" ")`, so the `isNaN` guard does not catch them and the document is stamped `1/1/1970` (`12/31/1969` west of UTC). Omitting the key entirely works correctly, because `Number(undefined)` is `NaN`.

Expected: a nullable field sent as null behaves the same as omitting it.

The date is not only cosmetic. `published` is plucked into `chunkHeaderMeta` (`server/utils/TextSplitter/index.js:73-76`) and `stringifyHeader()` prepends it as literal `<document_metadata>` text on every chunk (`:135-147`), which all eight vector providers write. So the wrong date is embedded in the text the model reads, not just shown in the UI.

Against a local development server (`yarn dev:server`, default `SERVER_PORT=3001`; routes mount at `/api`, `server/index.js:81`). Generate an API key under Settings, Tools, Developer API.

1. Send the raw text with `published` explicitly null. `addToWorkspaces` is optional here, so nothing else is needed:

```bash
curl -sS -X POST http://localhost:3001/api/v1/document/raw-text \
  -H "Authorization: Bearer $ANYTHINGLLM_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"textContent":"hello world","metadata":{"title":"doc.txt","published":null}}' \
  | jq '.documents[0].published'
```

2. Compare against the same request with the key omitted entirely:

```bash
curl -sS -X POST http://localhost:3001/api/v1/document/raw-text \
  -H "Authorization: Bearer $ANYTHINGLLM_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"textContent":"hello world","metadata":{"title":"doc.txt"}}' \
  | jq '.documents[0].published'
```

The endpoint responds `{ success, error, documents }` (`server/endpoints/api/document/index.js:636`), so `.documents[0].published` is the field to read. The same value appears on the file row's hover tooltip in the document picker.

Observed: `12/31/1969, 6:00:00 PM` in a UTC-negative timezone, `1/1/1970` at or east of UTC.
Expected: the current time, the same as omitting `published`.

Same result for `""`, `" "` and `[]`. `published: 0` is a genuine epoch timestamp and should keep resolving to 1970.

Not applicable, the defect is in the collector's metadata normalization.

Not applicable.

marmar9615-cloud:
### Summary

Resolves #6274.

`GET /v1/document/metadata-schema` documents `published` as `"epoch timestamp in ms | nullable"`, but sending it as `null` stamps the document with the Unix epoch.

`collector/processRawText/index.js:42`
```js
if (isNaN(Number(published))) return new Date().toLocaleString();
return new Date(Number(published)).toLocaleString();
```

`Number(null)`, `Number("")` and `Number(" ")` are all `0`, so the `isNaN` guard never fires and the value becomes `1/1/1970` (`12/31/1969` west of UTC). Omitting the key works, because `Number(undefined)` is `NaN`.

### Why this is worth more than a wrong date column

`published` is plucked into `chunkHeaderMeta` (`server/utils/TextSplitter/index.js:73-76`) and `stringifyHeader()` prepends it as literal `<document_metadata>` text on **every chunk** (`:135-147`), called unconditionally by all eight vector providers. So the epoch date is embedded into the text the model reads. That path is traced through the source rather than exercised here: the tests in this PR assert the collector's `published` value, not the rendered chunk header. In the UI it only surfaces in the file row's hover tooltip (`Directory/FileRow/index.jsx:12-17`), which is the smaller half of the problem.

### Change

Five inserted lines in one file, three of them comment. Guard on a trimmed string so `null`, `""`, whitespace and `[]` are all absent.

`published: 0` is a genuine epoch timestamp and still resolves to 1970. No clause is needed for `undefined`: `Number(undefined)` is already `NaN`, so it falls to the existing `isNaN` branch. I checked that rather than adding a clause that does nothing.

### Tests

Nine cases in `collector/__tests__/processRawText/index.test.js`, run against three variants in `git archive` exports so the working tree was never mutated:

| variant | result |
| --- | --- |
| `master` | **5 failed** on `null`, `""`, `" "`, `"\t"`, `[]` |
| the shorter `if (!published)` | **4 failed**: swallows the real `published: 0`, and still misses `" "`, `"\t"`, `[]` |
| this PR | **9 passed** |

That middle row is why the guard trims rather than testing truthiness. `!published` looks equivalent and is not: it treats a legitimate `0` epoch as absent while letting whitespace through.

The four cases that pass on `master` are controls that must keep passing: an omitted key, a numeric epoch, a string epoch, and `0`.

`npx jest collector` on this branch: **157 passed, 10 suites**, against a **148 passed, 9 suites** baseline on `master`. Prettier clean.

The nine new tests discriminate rather than merely pass:

| implementation | result |
| --- | --- |
| `master` (unfixed) | 5 of 9 fail |
| `if (!published)` instead of the trimmed-string guard | 4 of 9 fail, since `" "`, `"\t"` and `[]` are all truthy and fall through to `Number()` |
| this branch | 9 of 9 pass |

Repository-wide, `yarn test` is red on `master` before this change: 21 suites fail on Node 26 where `jsonwebtoken` reaches for the removed `SlowBuffer`. That is unrelated to this PR and this PR does not touch it, which is why the numbers above are scoped to `npx jest collector`.

### One case deliberately left

`published: false` still resolves to 1970, because `Number(false)` is `0` and `` `${false}`.trim() `` is `"false"`. A boolean is outside the documented type for this field, and separating it from `0` needs a `typeof` check for an input nothing sends. Happy to add it if you would rather the guard be total.


timothycarambat:
Thanks for this, the bug report and the fix are both spot on. I checked the tests out locally and confirmed they discriminate: 5 of 9 fail on `master`, 9 of 9 pass on the branch.

I pushed a follow-up commit on top of yours that widens the coverage rather than changing your approach:

**`published`**
- Booleans now count as absent. `Number(false)` is `0` and `Number(true)` is `1`, so both were still landing on the epoch. You flagged `false` in the description; it was a one-line `typeof` so I closed it.
- Out-of-range values like `1e20` or `"Infinity"` were passing the `isNaN` guard and rendering the literal string `"Invalid Date"`, which then gets baked into every chunk header the same way the 1970 date did. The Date is now constructed once and checked with `getTime()`, which also folds in the original `isNaN(Number())` branch.

**Input guards**
- Non-string `textContent` (a number, array, object) now returns the existing empty-content failure instead of throwing on `.split`.
- Missing or non-string `metadata.title` returns a structured failure instead of throwing inside `stripAndSlug`. The server API validates title before calling the collector, but the collector's own `/process-raw-text` route does not, so this was reachable.

timothycarambat:
Thanks for this, the bug report and the fix are both spot on. I checked the tests out locally and confirmed they discriminate: 5 of 9 fail on `master`, 9 of 9 pass on the branch.

I pushed a follow-up commit on top of yours that widens the coverage rather than changing your approach:

**`published`**
- Booleans now count as absent. `Number(false)` is `0` and `Number(true)` is `1`, so both were still landing on the epoch. You flagged `false` in the description; it was a one-line `typeof` so I closed it.
- Out-of-range values like `1e20` or `"Infinity"` were passing the `isNaN` guard and rendering the literal string `"Invalid Date"`, which then gets baked into every chunk header the same way the 1970 date did. The Date is now constructed once and checked with `getTime()`, which also folds in the original `isNaN(Number())` branch.
- Moved the explanatory comments into a JSDoc block on the handler so the coverage is documented in one place.

**Input guards**
- Non-string `textContent` (a number, array, object) now returns the existing empty-content failure instead of throwing on `.split`.
- Missing or non-string `metadata.title` returns a structured failure instead of throwing inside `stripAndSlug`. The server API validates title before calling the collector, but the collector's own `/process-raw-text` route does not, so this was reachable.

**Tests**
Grew your suite from 9 to 106 cases in the same file, using the same harness. New blocks cover input validation, title slugging (including path traversal and absolute paths staying flat), url scheme filtering (`ftp:`, `javascript:`, `file:`, `data:` all rejected), the four string-metadata fallbacks against seven non-string types each, and the document assembly/writer contract. A couple of existing quirks are pinned as current behavior rather than changed: `"..."` slugs to `.txt`, and `wordCount` splits on single spaces only.

Full `jest collector` is green at 254 tests. Merging once CI passes.

marmar9615-cloud:
Thanks, and agreed on all of it. The `1e20` and `"Infinity"` case is a real catch, that string would have landed in every chunk header the same way the 1970 date did. I ran your head locally on Node 26.5: the processRawText file passes 106 of 106 and `jest collector` passes all 254. Nothing further from my side.

PR #6280, opened 2026-09-03
Make web search use You.com as the default provider

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #

### Description

- promote web search from a configurable skill to a default-enabled skill
- default the search provider to You.com's keyless free tier when no provider is configured
- fall back to DuckDuckGo automatically on any You.com failure or empty response
- use all You.com result snippets and published dates for richer search context
- prevent stale skill settings from reenabling a skill an admin has disabled
- keyless You.com tier allows 100 queries/day per IP ([docs](https://you.com/docs/quickstart)) and responds `402` once users hit limit

### Developer Validations


server/utils/agents/aibitat/plugins/web-browsing.js
```diff
@@ -1395,29 +1400,32 @@ const webBrowsing = {
...
             });
 
-            if (data.length === 0)
-              return `No information was found online for the search query.`;
+            if (error || data.length === 0) {
+              this.super.handlerProps.log(
+                `You.com Search ${error ? "failed" : "returned no results"} - falling back to DuckDuckGo.`
+              );
+              return await this._duckDuckGoEngine(query);
+            }
```
timothycarambat: Only on error should we failover to DDG. Hitting DDG with no results probably wont help us here. If we get zero results from You, then there just might be zero results. If the zero comes from a 429 or 402 - that should be an error.

server/utils/agents/aibitat/plugins/web-browsing.js
```diff
@@ -1395,29 +1400,32 @@ const webBrowsing = {
...
                 Array.isArray(snippets) && snippets.length > 0
-                  ? snippets[0]
+                  ? snippets.join("\n")
                   : description;
               if (!url && !title) return;
               data.push({
                 title: title || "",
                 link: url || "",
                 snippet: snippet || "",
+                ...(page_age ? { published: page_age } : {}),
```
timothycarambat, line 1444: `page_age` was the only real cool attrib we get from You provider? Nothing else we could possibly pass the LLM that might enhance the results?

timothycarambat (changes requested):

server/utils/agents/aibitat/plugins/web-browsing.js
shatfield4: Checked the response object and these are the only attributes we get back: `url`, `title`, `description`, `snippets`, `page_age`, `thumbnail_url` and `favicon_url`. We already pass everything except the thumbnail and favicon URLs so `page_age` is the only real new attribute we can use.

PR #6285, opened 2026-09-04
Apply workspace temperature to agents and omit the parameter when unset

### LINKED ISSUE #6091: [BUG]: Workspace LLM Temperature is not applied to agent sessions, and no agent-level temperature setting exists

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6091

### Description

- omit the temperature parameter from provider requests entirely when the workspace has no value set, letting the provider use its own default (temperature input is now optional and clearable)
- validate the value once at connector construction and carry it as a provider instance property instead of passing it through every chat handler
- apply the workspace temperature to agent sessions using the same model-specific guards already in place for the chat provider entrypoints (OpenAI o*/gpt-5, Anthropic/Bedrock no-temperature models, Azure reasoning deployments) - the agent providers just reuse them
- preserve the forced temperature 0 default for Mistral and Cerebras chat when unset

### Developer Validations


timothycarambat (changes requested):
Router workspaces lose the OpenAI-compatible API `temperature` override.

`resolveProviderConnector` only honors the new `temperature` arg on the non-router branch; the router branch builds the delegate from `workspace.openAiTemp` and ignores it. 

Since `openaiCompatible.js` no longer passes temperature into `getChatCompletion`, a router workspace hit via `/v1/openai/chat/completions` with `temperature: 0` will now go out with the workspace value. Easiest fix is probably to pass `temperature` into `AnythingLLMModelRouter` (or into its `getLLMProvider` call) so both branches behave the same. I think that will work, but needs to be tested.

---

Please remove the "close live agent sessions on workspace save" change.

It's a good change feature, but it fires from every `Workspace.update` caller (settings save, developer API, community hub import, Telegram model picker), and it aborts an in-flight agent run and closes the invocation. Lets just uncheck it for now from at least this PR.

Then make an issue for autosave on edit of workspace fields and lets handle that UX convenience and revisit the abort handler on save later. When you leave the page it should abort the session anyway so that shouldnt change anything and you cannot change the config "on the fly" anyway.

---

Smaller asks:
- Add a unit test for `resolveTemperature` (null, "", "0", negative, non-numeric, and one guarded model per class)
- Rebase to master

timothycarambat (changes requested):
Referring to the [comment from OR](https://github.com/Mintplex-Labs/anything-llm/pull/6337#pullrequestreview-5154920680)

> Anytime we are potentially pushing a key:value and value is undefined, we should spread it via a class util so that we are not ever sending key: undefined values as undefined over REST is not good practice ([RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259)). null is valid, undefined is not. Serialization of payload bodies also makes this a bad idea in general while spreading {} just omits it before/during serialization

We are passing potentially undefined here for this.temperature and it should have a shared utils that we can reasonable spread this on

PR #6287, opened 2026-09-04
fix(qdrant): use client.query after js-client-rest 1.19 dropped search

### LINKED ISSUE #6270: [BUG]: Qdrant vector search always fails — client.search() was removed in @qdrant/js-client-rest 1.19.0

AnythingLLM desktop app

### What happened?

Using Qdrant as vector database and for the Embedder option: "Generic OpenAI embedding model" Nemotron-3-Embed-8B-GGUF

Every RAG retrieval against a Qdrant vector database fails. In normal chat/query
mode this silently shows up as "There is no relevant information in this
workspace to answer your query". In agent mode the error surfaces:

```
[AgentHandler] memory.search raised an error. e.search is not a function

TypeError: e.search is not a function
    at Wg.similarityResponse (resources\backend\server.js:875:60730)
```

The workspace, collection and embedder are all configured correctly — the
collection exists, the vector count shows up in the workspace's Vector Database
tab, and the embedder runs fine (the log shows `GenericOpenAiEmbedder
Initialized` and the query prefix being applied right before the crash).

AnythingLLM 1.16.1 ships `@qdrant/js-client-rest` **1.19.0**. That release
removed `search` from the REST client. From the qdrant-js changelog for 1.19.0:

> The REST client dropped `search`, `searchBatch`, `searchPointGroups`,
> `recommend`, `recommendBatch`, `recommend_batch`, `recommendPointGroups`,
> `discoverPoints` and `discoverBatchPoints`. Use `query`, `queryBatch` and
> `queryGroups` instead

But `server/utils/vectorDbProviders/qdrant/index.js` still calls the old method
in `similarityResponse`:

```js
const responses = await client.search(namespace, {
  vector: queryVector,
  limit: topN,
  with_payload: true,
});
```

I confirmed the method is gone in the shipped package:

```
> node -e "const p=require('@qdrant/js-client-rest'); console.log(typeof p.QdrantClient.prototype.search, typeof p.QdrantClient.prototype.query)"
undefined function
```

Because the TypeError is thrown client-side, no search request is ever sent.
The Qdrant server log for a full chat attempt shows only the connection and the
namespace check:

```
"GET /cluster HTTP/1.1" 200 78 "-" "qdrant-js/1.19.0"
"GET /collections/my_collection HTTP/1.1" 200 591 "-" "qdrant-js/1.19.0"
```

There is never a `POST /collections/my_collection/points/search`. That is what
makes this confusing to diagnose: the connection works, the settings page reads
the vector count correctly, and it looks like an embedding or configuration
problem.

Since `similarityResponse` is shared by all chat modes, this affects Chat, Query
and Agent equally. It is not model-dependent — I reproduced it with several
different LLMs.

1. Set the vector database to Qdrant (I used Qdrant server 1.19.0 in Docker).
2. Create a workspace and embed any document, or point the workspace slug at an
   existing collection.
3. Ask anything in the chat.
4. Nothing is retrieved. `storage/logs/backend-*.log` shows
   `TypeError: e.search is not a function at ... similarityResponse`, and the
   Qdrant server log shows no search request.

### Suggested fix

Migrate the call to `query()`, which is the documented replacement. It returns
`{ points: [...] }` instead of a bare array:

```js
const { points: responses } = await client.query(namespace, {
  query: queryVector,
  limit: topN,
  with_payload: true,
});
```

It would probably also be worth pinning `@qdrant/js-client-rest` rather than
using a caret range, since this was a breaking change picked up automatically at
build time.

### Workaround

For anyone stuck on 1.16.1: adding a small `search()` shim back to
`node_modules/@qdrant/js-client-rest/dist/cjs/qdrant-client.js` restores
retrieval. It just forwards to `query()` and unwraps `.points`:

```js
async search(collection_name, { vector, filter, params, limit = 10, offset,
                                with_payload = true, with_vector = false,
                                score_threshold, consistency, timeout, shard_key } = {}) {
    const response = await this.query(collection_name, {
        consistency, timeout, shard_key,
        query: vector, filter, params,
        score_threshold, limit, offset,
        with_vector, with_payload,
    });
    return response?.points ?? [];
}
```

After that the `POST /points/search` requests show up in the Qdrant log and
retrieval works normally. Note this gets wiped by every app update.
```


any open weight LLM with tool use capabilities.

Nemotron-3-Embed-8B-GGUF

leilei3167:
## Summary
`@qdrant/js-client-rest` 1.19 removed `client.search()`. Desktop/chat retrieval still called it in `similarityResponse`, so Qdrant workspaces hit `TypeError: e.search is not a function` and returned no chunks.

## Changes
- Call `client.query(namespace, { query, limit, with_payload })` and unwrap `{ points }`
- Regression test with a 1.19-shaped mock (query present, search undefined)

Fixes #6270

### DIFF
--- a/server/__tests__/utils/vectorDbProviders/qdrant/index.test.js
+++ b/server/__tests__/utils/vectorDbProviders/qdrant/index.test.js
@@ -1,0 +1,160 @@
+/* eslint-env jest, node */
+
+jest.mock("@qdrant/js-client-rest", () => ({
+  QdrantClient: jest.fn(),
+}));
+
+jest.mock("../../../../utils/TextSplitter", () => ({
+  TextSplitter: jest.fn(),
+}));
+
+jest.mock("../../../../models/systemSettings", () => ({
+  SystemSettings: {},
+}));
+
+jest.mock("../../../../utils/files", () => ({
+  storeVectorResult: jest.fn(),
+  cachedVectorInformation: jest.fn(),
+}));
+
+jest.mock("uuid", () => ({
+  v4: jest.fn(() => "test-uuid"),
+}));
+
+jest.mock("../../../../utils/helpers", () => ({
+  toChunks: jest.fn(),
+  getEmbeddingEngineSelection: jest.fn(),
+}));
+
+jest.mock("../../../../utils/chats", () => ({
+  sourceIdentifier: jest.fn((payload) => {
+    if (!payload?.title || !payload?.published) return "anon-source";
+    return `title:${payload.title}-timestamp:${payload.published}`;
+  }),
+}));
+
+const { QDrant } = require("../../../../utils/vectorDbProviders/qdrant");
+
+describe("QDrant.similarityResponse", () => {
+  let provider;
+
+  beforeEach(() => {
+    provider = new QDrant();
+    jest.spyOn(provider, "logger").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    jest.restoreAllMocks();
+  });
+
+  it("uses client.query (not search) and maps {points} for @qdrant/js-client-rest 1.19+", async () => {
+    const queryVector = [0.1, 0.2, 0.3];
+    const query = jest.fn().mockResolvedValue({
+      points: [
+        {
+          id: "hit-1",
+          score: 0.91,
+          payload: {
+            text: "relevant chunk",
+            title: "doc.md",
+            published: "2026-01-01",
+          },
+        },
+        {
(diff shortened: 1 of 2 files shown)


leilei3167:
@timothycarambat ready for review — `@qdrant/js-client-rest` 1.19 dropped `search()`, so Qdrant RAG was throwing `e.search is not a function`. This switches `similarityResponse` to `query()` and unwraps `{points}`, with a mock regression test covering a search-less client. Happy to adjust if anything looks off.

PR #6333, opened 2026-09-08
Update index.js

MaddipatlaChetan24:
### Pull Request Type

### Relevant Issues

resolves #

### Description

Optimizes `processSingleFile` in the collector's file-processing pipeline:

- Replaces the synchronous `fs.existsSync` check with the non-blocking
  `fs.promises.access(...)`. `processSingleFile` is already `async` and can
  run concurrently across many files during directory-watch processing;
  a sync filesystem call there blocks Node's single event loop thread for
  every other in-flight request while the stat completes. The promise-based
  version preserves identical semantics (any access failure — missing file,
  permission denied, etc. — is treated as "doesn't exist").
- Moves the `RESERVED_FILES.includes(targetFilename)` check to the top of
  the function, before path resolution/normalization, since it only depends
  on the raw filename and never needed those steps to run first. Skips
  unnecessary path work when a reserved filename is rejected early.

No change to the function's return shape, success/failure semantics, or
which files get accepted/rejected — purely a performance/ordering cleanup.

### Visuals (if applicable)

N/A — no UI changes.

### Additional Information

N/A

### Developer Validations

### DIFF
--- a/collector/processSingleFile/index.js
+++ b/collector/processSingleFile/index.js
@@ -10,6 +10,7 @@ const {
   normalizePath,
   isWithin,
 } = require("../utils/files");
+
 const RESERVED_FILES = ["__HOTDIR__.md"];
 
 /**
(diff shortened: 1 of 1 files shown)


MaddipatlaChetan24 (commented):
can you please check this

timothycarambat:
Read the rules of https://github.com/Mintplex-Labs/anything-llm/blob/master/CONTRIBUTING.md before making self-assigned PRs please

PR #6337, opened 2026-09-09
Add OpenRouter service tier selection

### LINKED ISSUE #6306: [FEAT]: Support `service_tier` (OpenRouter / OpenAI flex & priority tiers) for LLM requests

**Summary.** Let users select a **service tier** (`flex` / `priority`) for the OpenRouter and OpenAI providers, sent as the `service_tier` field of every chat-completion request — in the normal chat path and in the agent (aibitat) path.

**Why.** OpenRouter and OpenAI expose cheaper/faster processing tiers as a top-level request parameter: `"service_tier": "flex"` runs on OpenAI's best-effort Flex tier at **half the token price** (e.g. GPT-5.6 Luna Pro: $0.10/$0.60 per M vs $0.20/$1.20), `"priority"` is the faster/pricier tier. OpenRouter docs: https://openrouter.ai/docs/guides/features/service-tiers. The tier is per request only — there is no account- or key-level default — so the client has to send it. Today AnythingLLM never sends it, so the only way to reach flex is OpenRouter's `:floor` model suffix, which has two problems inside AnythingLLM:

1. `:floor` variants are not in OpenRouter's `/api/v1/models` catalog, so the model dropdown cannot show/select them (the saved value is not in the list and the first model appears selected; saving the page silently overwrites the preference).
2. `OpenRouterLLM.promptWindowLimit()` resolves the context window from the cached catalog (`models()[model]?.maxLength || 4096`), so an unknown id gets a 4096-token window — which also makes `web-scraping` summarize pages that exceed it.

I verified both mechanisms directly against OpenRouter (generation metadata reports `service_tier`): a plain `openai/gpt-5.6-luna-pro` request billed at 1.00×; the same with `"service_tier": "flex"` → `service_tier=flex`, 0.50×; `"priority"` → 2.00×.

**Proposal.**
- A "Service tier" option (default / flex / priority) in the OpenRouter provider settings (and ideally the OpenAI provider, which supports the same field), stored as e.g. `OPENROUTER_SERVICE_TIER`, and added to the request body in:
  - `server/utils/AiProviders/openRouter/index.js` (`getChatCompletion` / `streamGetChatCompletion`, next to `include_reasoning`),
  - the agent path: `server/utils/agents/aibitat/providers/openrouter.js` and the shared `providers/helpers/tooled.js` (`chat.completions.create` in `tooledStream` / `tooledComplete`).
- Alternatively / additionally: extend the extra-request-params passthrough proposed in #6048 (`storage/config/llm/<provider>.json`) to the **OpenRouter** provider and to the shared tooled agent path, so `service_tier` (and any other provider-specific field) can be set there without a dedicated UI control. #6048 currently covers only `generic-openai`.
- Nice to have: an option to retry once without the tier on a flex capacity error (OpenRouter does not fall back from flex to the default tier by itself: a 429 surfaces instead).

**Workaround I use today.** A tiny local proxy in front of OpenRouter that injects `service_tier: "flex"` and serves `/v1/models` from the catalog, with AnythingLLM's Generic OpenAI provider pointed at it. It works (all calls report `service_tier=flex`, 50 % cost) but it is one more process to keep alive.

**Environment.** AnythingLLM Desktop 1.16.1 (Windows 11), OpenRouter provider with `openai/gpt-5.6-luna-pro`; agent workspace with MCP tools (native tool calling).

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6306 

### Description

- add a Service Tier dropdown under the OpenRouter advanced controls, stored as `OPENROUTER_SERVICE_TIER`
- offer every tier OpenRouter has in docs for the request body: `auto`, `default`, `fast`, `flex`, `priority`, `scale` (omitting `null` since it functions as if we set this to `auto` or `default`)
- send `service_tier` on every OpenRouter chat completion: normal chat, native tool calling agents, and the UnTooled agent fallback
- spread the param in so the key is never sent as `undefined` and is omitted entirely when no tier is set
- add tests to ensure no regressions

### Developer Validations


server/utils/AiProviders/openRouter/index.js
```diff
@@ -514,6 +518,40 @@ class OpenRouterLLM {
...
+      const tiers = new Set();
+      for (const { tag = "" } of data?.endpoints ?? []) {
+        const tier = tag.split("/").pop();
+        if (tier === "flex") tiers.add("flex");
+        if (tier === "fast" || tier === "priority") tiers.add("priority");
+      }
+      return SERVICE_TIERS.filter((tier) => tiers.has(tier));
+    })
+    .catch(() => []);
+}
```
timothycarambat: Some questions and clarifications are required:

OR Docs show this https://openrouter.ai/docs/api/api-reference/chat/create-a-chat-completion#body-service-tier-one-of-0
Image

This endpoint seems to indicate that the available endpoints might vary between model - which makes sense. In the docs about service tiers:

> The service_tier parameter. For priority, matching endpoints are tried first (sorted by throughput), with fallback to other endpoints if none succeed; billing always follows the endpoint actually used, so a priority request that falls back off-tier is charged at that endpoint’s standard rate, not the tier rate. For flex, routing is restricted to flex endpoints (sorted by price). Flex never falls back to a default-tier endpoint, since that would cost more than the tier you requested, so a flex capacity error surfaces instead. 
> 
> **If the pool contains no flex endpoints at all (for example, the model has no flex-capable provider), the request routes normally at standard rates.** Combine with [allow_fallbacks: false](https://openrouter.ai/docs/guides/routing/provider-selection#disabling-fallbacks) to route only to the top endpoint of that tier.

Since we do not set `allow_fallbacks` I think we can take a much more simple approach of just static values in the picker since if the user picks the wrong value it is inconsequential apparently. Given this is an "advanced" feature for OR for powerusers since most people will not configure this - I dont think we need all this infra for this functionality.

server/utils/agents/aibitat/providers/openrouter.js
```diff
@@ -47,6 +49,7 @@ class OpenRouterProvider extends InheritMultiple([Provider, UnTooled]) {
       .create({
         model: this.model,
         messages,
+        service_tier: this.serviceTier,
```
timothycarambat: Here we could possibly be sending `undefined` in the payload, which maybe OR strips but it is bad practice as it is not an accepted value in their docs.

If the parameter is not set it should be omitted.

server/utils/AiProviders/openRouter/index.js
```diff
@@ -16,6 +16,7 @@ const cacheFolder = path.resolve(
     ? path.resolve(process.env.STORAGE_DIR, "models", "openrouter")
     : path.resolve(__dirname, `../../../storage/models/openrouter`)
 );
+const SERVICE_TIERS = ["flex", "priority"];
```
timothycarambat: This is not true, since the service tiers can be `auto`, `default`, `fast`, `flex`, `priority`, `scale`, and `null` according to docs endpoint

server/utils/helpers/updateENV.js
```diff
@@ -498,6 +498,15 @@ const KEY_MAPPING = {
     envKey: "OPENROUTER_TIMEOUT_MS",
     checks: [],
   },
+  OpenRouterServiceTier: {
+    envKey: "OPENROUTER_SERVICE_TIER",
+    checks: [
+      (input) =>
+        ["default", "flex", "priority"].includes(input)
```
timothycarambat: This array is defined here and in `server/utils/AiProviders/openRouter/index.js` but in this current PR it is:
1. Available values are dependent on model endpoint parsing
2. These are not all possible acceptable values
3. This is testing explicitly the `alises` only

frontend/src/components/LLMSelection/OpenRouterOptions/index.jsx
```diff
@@ -140,3 +160,65 @@ function OpenRouterModelSelection({ settings }) {
...
+  useEffect(() => {
+    if (!model) return;
+    let stale = false;
+    System.openRouterServiceTiers(model).then((tiers) => {
+      if (!stale) setTiers(tiers);
+    });
+    return () => {
+      stale = true;
+    };
+  }, [model]);
```
timothycarambat: If tier being invalid during API request is inconsequential there is no need to mark stale. It will load and re-eval on model change. Also since this is being live fetched it should have a loading state if doing stuff like this.

frontend/src/components/LLMSelection/OpenRouterOptions/index.jsx
```diff
@@ -22,7 +25,13 @@ export default function OpenRouterOptions({ settings }) {
           />
         </div>
         {!settings?.credentialsOnly && (
-          <OpenRouterModelSelection settings={settings} />
+          <>
+            <OpenRouterModelSelection
+              settings={settings}
+              onModelChange={setSelectedModel}
+            />
+            <ServiceTierSelection settings={settings} model={selectedModel} />
```
timothycarambat: Should be moved under advanced controls. People are not going to practically change this value and are normally just going to pick a single model.

server/utils/agents/aibitat/providers/helpers/tooled.js
```diff
@@ -222,6 +223,7 @@ async function tooledStream(
     stream_options: { include_usage: true },
     messages: formattedMessages,
     ...maxTokensParam(maxTokens),
+    service_tier: serviceTier,
```
timothycarambat: should be spread since it can be undefined (or not at all for cases where OR is the router and they upgrade to now have this change in codebase and have an undefined value)

timothycarambat (changes requested):
This change introduces a ton of complexity for what is clearly a very narrow feature that probably a very low % of people use. That being said I think we can rip out a ton of this and simplify things greatly.  Some considerations in their docs for tiers and fallbacks: https://openrouter.ai/docs/guides/features/service-tiers#comparing-tier-selection-options


Some things to address:

1. Can you just use any valid service_tier slug with any model even if the endpoint does not explicitly support it - or does it error?  If so, we can simplify this massively.

1a. If we can, just provide a static picker dropdown in the UI that offers a `default` or `auto` as the main/unknown value - whatever the default value is for the API request. We can then delete all the endpoint retrieval and complexity here.

2. This dropdown should be tucked under advanced controls/settings.

3. Anytime we are potentially pushing a `key:value` and value is undefined, we should spread it via a class util so that we are not ever sending `key: undefined` values as `undefined` over REST is not good practice ([RFC 8259](https://datatracker.ietf.org/doc/html/rfc8259)). `null` is valid, undefined is not. Serialization of payload bodies also makes this a bad idea in general while spreading `{}` just omits it before/during serialization

server/utils/AiProviders/openRouter/index.js
shatfield4: Testing this by using multiple models from different providers in OR has shown that `auto` and `default` are just aliases for each other and same with `fast` and `priority`. `scale` was not able to be used on any models I tested which is why it was omitted before but added this since their docs show it as an option.

server/utils/AiProviders/openRouter/index.js
shatfield4: See note on other review message here: https://github.com/Mintplex-Labs/anything-llm/pull/6337#discussion_r3973627717

Omitted `null` as an option since it functions the same as `auto` or `default`

server/utils/agents/aibitat/providers/openrouter.js
shatfield4: Done

frontend/src/components/LLMSelection/OpenRouterOptions/index.jsx
shatfield4: Removed along with all the backend endpoints for dynamically pulling tiers from OR.

server/utils/agents/aibitat/providers/helpers/tooled.js
shatfield4: Now uses the same spread helper function as everywhere else in the backend

```
...serviceTierParam(serviceTier)
```

frontend/src/components/LLMSelection/OpenRouterOptions/index.jsx
shatfield4: Done

server/utils/AiProviders/openRouter/index.js
```diff
@@ -514,6 +518,16 @@ class OpenRouterLLM {
...
+/**
+ * Build the `service_tier` request field, spread so the key is absent entirely
+ * when no tier is configured.
+ * @param {unknown} tier
+ * @returns {{service_tier?: string}}
+ */
+function serviceTierParam(tier) {
+  return tier ? { service_tier: tier } : {};
+}
+
```
timothycarambat: You have defined this exact same function twice here and in [tooled.js](https://github.com/Mintplex-Labs/anything-llm/pull/6337/changes#diff-ef907323ddbb957f9bd570028e6c45b9b3d33e956f24cbb9b25ad2c0ba518bbdR183-R194) and its pretty confusing as they both do the exact same thing 

The tooled one is in fact more defensive and better. Since we already have `maxTokensParam` in tooled lets keep it there and import it into the OR AiProvider.

Then to make it more clear that SERVICE_TIERS are specific to OR move that const to a `static` prop on the AiProvider or Aibitat OR class and use that when referencing `SERVICE_TIERS` instead

timothycarambat (changes requested):

server/utils/AiProviders/openRouter/index.js
shatfield4: Deduped helper and now reuses the one from tooled.js. Also made SERVICE_TIERS static on the OpenRouter LLM provider class.

PR #6343, opened 2026-09-10
Autosave workspace settings on edit

### LINKED ISSUE #6334: [FEAT]: Autosave on edit of workspace fields/options

- On save of workspace settings/fields changes we should autosave
- Debouce saving
- Investigate abort handler for restarting agent sessions if setting changed will not apply until agent session is restarted
- This can live on the backend model for `Workspace.update()` but we must keep in mind this means it may kill all sessions for all users if they are in the middle of chatting

shatfield4:
### Pull Request Type

- [x] ✨ feat (New feature)
- [ ] 🐛 fix (Bug fix)

### Relevant Issues

resolves #6334

### Description

- workspace settings fields now save themselves shortly after editing stops instead of requiring an explicit update button click
- rapid edits are debounced into a single save so typing in a field does not fire a request per keystroke
- leaving the page before the debounce fires saves the pending edit
- invalid values are not saved and the browser's native validation message is shown instead
- the update workspace buttons are removed from the general, chat, vector database, and agent settings tabs
- the existing success and error toasts remain as the save feedback

### Developer Validations


timothycarambat (changes requested):
This needs to be revisited here in terms of UX. I get the idea here and it practically speaking autosaving is working, but the UX is awful and we cannot ship this as is.

1. This spams the toast notification and it is very annoying to see every second when making edits or pausing during an input message for a few ms.
2. It saves in the middle of editing or doing something while the field is still in focus (eg: system prompt or temp)
3. The debounce does not fire instantly on change (eg: chat mode, when i click, that should be an instant save)
4. When I click on the provider dropdown for workspace it instantly saves even when model is not yet saved. Then when I select the model from dropdown is does not instantly save. When this field is an input it saves in the middle of typing.

Suggestions:
- Add ephemeral indicator to the label on these fields that can be shown with some kind of indication that the edit was saved. Like a small text or check or something so the user knows an edit saved since the button is going away.
- Do not save while the field is focused for a form. Consider on blur for input types
- On dropdowns or buttons, we should save on user click.
- For provider/model dont save on provider change, save on model selection or loading or something so that they both save at the same time. Something like that would be tolerable.

I know these UX changes will be involved and it will touch more code but if we are going to implement autosave it should feel and function like autosave in other applications. Given how settings are laid out it might make sense to even wrap the forms on a Context or something so they are easier to manager per-form and keep track of form state.

shatfield4:
> This needs to be revisited here in terms of UX. I get the idea here and it practically speaking autosaving is working, but the UX is awful and we cannot ship this as is.
> 
> 1. This spams the toast notification and it is very annoying to see every second when making edits or pausing during an input message for a few ms.
> 2. It saves in the middle of editing or doing something while the field is still in focus (eg: system prompt or temp)
> 3. The debounce does not fire instantly on change (eg: chat mode, when i click, that should be an instant save)
> 4. When I click on the provider dropdown for workspace it instantly saves even when model is not yet saved. Then when I select the model from dropdown is does not instantly save. When this field is an input it saves in the middle of typing.
> 
> Suggestions:
> 
> * Add ephemeral indicator to the label on these fields that can be shown with some kind of indication that the edit was saved. Like a small text or check or something so the user knows an edit saved since the button is going away.
> * Do not save while the field is focused for a form. Consider on blur for input types
> * On dropdowns or buttons, we should save on user click.
> * For provider/model dont save on provider change, save on model selection or loading or something so that they both save at the same time. Something like that would be tolerable.
> 
> I know these UX changes will be involved and it will touch more code but if we are going to implement autosave it should feel and function like autosave in other applications. Given how settings are laid out it might make sense to even wrap the forms on a Context or something so they are easier to manager per-form and keep track of form state.

Converted this over to a provider that now gives feedback from the `<SavedIndicator />` reusable component. Saves now happen on blur of text areas and input boxes instead of deboucing input fields. It also no longer spams toast success messages on save.

PR #6360, opened 2026-09-14
docs: add Easypanel to self-hosting deploy table

Ahson-Shaikh:
Following up from #295 (anythingllm-docs) — @timothycarambat mentioned this should be a PR against this repo instead. Adds Easypanel as an additional one-click deploy option in the Self-Hosting table. AnythingLLM has an official one-click template on Easypanel (https://easypanel.io/templates/anythingllm).


timothycarambat:
@Ahson-Shaikh FWIW, your template is using a very old version of anythingllm `mintplexlabs/anythingllm:1.12` - you can just use `latest` or at least our latest version tag `1.16`

Ahson-Shaikh:
Hi @timothycarambat - We do batch updates once a month. Can't use latest due to static configurations.

PR #6377, opened 2026-09-16
fix: give seven providers the vision treatment #6296 gave TogetherAI/SambaNova/Z.ai

L4XB:
## Pull Request Type

- [x] Bug fix

## What

Nine providers spread the raw chat history straight into the request, so the
internal `attachments` array went out in the body. That is the symptom. The gap
underneath it is that none of them had a `#generateContent`, so they were also
blind to the image on the **current** turn: an AnythingLLM user on a Fireworks
or Vertex vision model could attach an image and the model never saw it.

This applies the treatment #6296 gave TogetherAI, SambaNova and Z.ai.

| provider | change |
|---|---|
| Cohere, FireworksAI, GiteeAI, MiniMax, Perplexity, Vertex | OpenAI-shaped `#generateContent`, `formatChatHistory(chatHistory, this.#generateContent)`, and the current turn's `attachments` passed through it |
| DeepSeek | the Cerebras stub — no vision model is published, and the history still goes through the helper, which is what keeps `attachments` out of the body |
| Groq, PPIO | untouched |

All seven talk to an OpenAI-compatible endpoint through the `openai` SDK, so the
content shape is the same one `togetherAi`, `sambanova` and `zai` already emit:

```js
[{ type: "text", text: userPrompt },
 { type: "image_url", image_url: { url: attachment.contentString } }]
```

Groq is deliberately special-cased — `#conditionalPromptStruct` drops the system
prompt *and* the history for its vision models, with the reasons written out in
the file — and PPIO's `constructPrompt` documents `attachments` as unsupported.
Neither is changed.

`formatChatHistory` keeps its required formatter argument; every caller passes
one, so the default the first version of this branch added is not needed.

## Measurement

`server/__tests__/utils/AiProviders/visionChatHistory.test.js`, 27 tests over
the seven providers: an image on the current turn becomes an `image_url` part,
an image in the history becomes one too, assistant turns pass through untouched,
a message with no attachments stays a plain string, and no message ever carries
a key other than `role` and `content`. DeepSeek asserts the stub instead: the
attachments are dropped rather than sent as a part the model cannot read.

```
$ npx jest server/__tests__/utils/AiProviders/visionChatHistory
27 passed

# same tests, providers reverted to master
20 failed, 7 passed
```

The 7 that pass on master are the "plain string" cases, which are unchanged.

Full server suite:

```
master:      Test Suites: 48 passed   Tests: 665 passed
this branch: Test Suites: 49 passed   Tests: 692 passed
```

`eslint` clean, `prettier --check` clean.

## Developer Validations

- [x] Relevant documentation has been updated — none applies
- [x] Docker build succeeds locally — not run; provider files plus one test, no
      dependency or build change


timothycarambat (changes requested):
Thanks for the thorough writeup. The observation is correct: these nine providers spread the raw history and the internal `attachments` array goes out in the request body. I reproduced that on master.

I don't think dropping the attachments is the right fix though. Most of these providers (Fireworks, Vertex, Cohere, MiniMax, GiteeAI, Perplexity) serve vision models. The real gap is that they have no `#generateContent`, so they also ignore the image on the *current* turn. Stripping the history image makes the request smaller but leaves them just as blind.

What I'd like instead is the same treatment #6296 gave TogetherAI/SambaNova/Z.ai: an OpenAI-shaped `#generateContent` in each provider, `formatChatHistory(chatHistory, this.#generateContent)`, and the current-turn `attachments` passed through it. DeepSeek is text-only and can take the Cerebras stub. Groq and PPIO are intentionally special-cased and should stay as they are.

With that, the default-formatter change in `responses.js` isn't needed, since every caller passes one. I'd also drop the two test files. The `describe.each` over nine providers asserting key shape would need a rewrite for the vision behaviour anyway, and the helper test covers a default that won't exist.

Happy to review a revised version along those lines.

L4XB:
You're right, and thanks for reproducing it. Stripping the history image made the request smaller and left the model just as blind. Rewritten along the lines you asked for.

**Cohere, FireworksAI, GiteeAI, MiniMax, Perplexity, Vertex** each get the OpenAI-shaped `#generateContent`, `formatChatHistory(chatHistory, this.#generateContent)`, and the current turn's `attachments` passed through it — the same three lines as `togetherAi`/`sambanova`/`zai`. All seven talk to an OpenAI-compatible endpoint through the `openai` SDK, so the emitted shape is identical to theirs (`{type:"text"}` + `{type:"image_url", image_url:{url}}`, no `detail`).

**DeepSeek** takes the Cerebras stub, with the same `@dev-note` explaining that there is no vision model to fill it in for. Its history still runs through `formatChatHistory`, which is what keeps `attachments` out of the body.

**Groq and PPIO are untouched.** I had missed that both are deliberate: `#conditionalPromptStruct` drops the system prompt and the history for Groq's vision models with the reasoning written out in the file, and PPIO's `constructPrompt` documents attachments as unsupported.

**`responses.js` is back to master** — the default formatter isn't needed now that every caller passes one.

On the tests: I dropped both files. The helper test went with the default it covered, and you're right that the nine-provider key-shape `describe.each` needed a rewrite rather than a trim — so I rewrote it as one file asserting the vision behaviour instead of the key shape. 27 tests over the seven providers: current-turn image becomes an `image_url` part, history image becomes one too, assistant turns pass through untouched, a message without attachments stays a plain string, and DeepSeek asserts the opposite — attachments dropped rather than sent as a part the model cannot read. 20 of the 27 fail with the providers reverted to master; the 7 that pass are the plain-string cases that were never broken. Happy to drop that file too if you'd rather not carry it.

PR #6388, opened 2026-09-16
fix: give a merged table cell the columns it covers

L4XB:
Follow-up to #6387, which gave the scraper its table rules. It left one case open: a merged cell.

### What is wrong

`addTableRules` emits one `|` per `<th>`/`<td>` element, so a cell that spans columns or rows fills a single column. Every row carrying a span is then narrower than the header, and GFM reads the remaining values one column to the left.

On `master` (67d0f42b), a two-variant product table:

```html
<table>
  <tr><th>Product</th><th>Variant</th><th>Price</th></tr>
  <tr><td rowspan="2">Cable</td><td>1 m</td><td>9 EUR</td></tr>
  <tr><td>2 m</td><td>12 EUR</td></tr>
</table>
```

| today | with this change |
|---|---|
| `\| Product \| Variant \| Price \|` | `\| Product \| Variant \| Price \|` |
| `\| --- \| --- \| --- \|` | `\| --- \| --- \| --- \|` |
| `\| Cable \| 1 m \| 9 EUR \|` | `\| Cable \| 1 m \| 9 EUR \|` |
| `\| 2 m \| 12 EUR \|` | `\|  \| 2 m \| 12 EUR \|` |

`2 m` lands under **Product** and `12 EUR` under **Variant**, so the second variant is indexed as a product named `2 m` that costs nothing. `colspan` fails the same way in the other direction: a `<td colspan="3">total</td>` leaves its row two cells short, and a `<th colspan="2">` makes the delimiter row narrower than the body.

Merged cells are not exotic on the pages this scraper is pointed at — price tables, spec sheets and comparison tables use a `rowspan` for the product column and a `colspan` for a section heading routinely.

### The change

`addTableRules` lays the table out on a grid the way a browser does: each cell claims `colspan × rowspan` slots, and a cell that follows a claimed slot emits the empty cells for the columns it has to skip. Then

- a row is padded to the width of the widest row, which also covers a ragged row that has nothing to do with spans;
- the delimiter row counts the header columns by their spans rather than by how many `<th>` elements there are.

Markdown cannot merge cells, so the value is not repeated down the span — an empty cell is the closest honest rendering, and it is what markdownify does for `colspan` too. What it buys is that every remaining value stays under its own column.

The grid is measured once per `<table>` and cached in a `WeakMap`, so a page with many tables does not re-walk one for every cell.

Both copies of `addTableRules` change identically — `collector/processLink/helpers/htmlToMarkdown.js` and `open-computer/services/interface-service/utils/html-to-markdown.js` — and the two blocks are byte-identical after the change, as they were before it.

### Tests

Four cases added to `collector/__tests__/processLink/helpers/htmlToMarkdown.test.js` inside the existing `tables` block: colspan padding, rowspan padding, a header counted by its spans, and a row short of the widest row.

```
# source change stashed, tests kept
✕ pads the columns a colspan covers
✕ holds the column a rowspan covers open in the rows below it
✕ counts the header columns by their spans
✕ pads a row that is short of the widest row
Tests: 4 failed, 44 passed, 48 total

# with the change
Tests: 48 passed, 48 total
```

The 44 that pass on both sides include all seven table cases #6387 added, so the existing behaviour is pinned unchanged.

`npx jest __tests__/` in `collector/` goes from 72 to 76 tests with the same 5 failures and 10 unresolved suites before and after — those are `resyncPaperlessNgx` and a set of `Cannot find module` errors in my checkout, present on a clean `master` too.

Prettier clean on all three files.


L4XB:
Thanks @timothycarambat for a496e21 — that trailing-rowspan gap was real, and it was in the other three places this layout lives too (QwenLM/qwen-code#12039, CherryHQ/cherry-studio#20644, cline/cline#14190); I ported your rule and your test case there, crediting this PR.

On top of your commit, 58c1f716 fixes three things a reviewer measured in the identical rules on qwen-code, all reachable from any page the collector scrapes:

- **Unclamped spans.** `colspan="1000000"` produced ten million characters; `rowspan="65534" colspan="1000"` threw `RangeError: Set maximum size exceeded` after 3.4 s, so the fallback returned the whole page as the text `xxx`. `colspan` is now clamped to HTML's 1..1000 and `rowspan` to the rows left, and a table is padded to a full grid only while the grid stays within 64 + 8 cells per real cell and under 40,000; past that it is written row by row, one column per cell.
- **A row with no cells.** Turndown writes a blank `<tr>` as a blank line instead of calling the row rule. Mid-table that ended the table; as the first row it left the table with no delimiter. Blank lines between rows are dropped, and the header is the first row that has cells, found once per table.
- **A pipe inside a code span.** `<code>a\|b</code>` already carries its own backslash, so escaping it again made an even run, which GFM splits at. Only an even run gets one more; outside code spans the output is unchanged.

Your rule is kept as is: the width a row emits cells for is recorded before the trailing claimed slots are skipped. Swapping that order back fails your `holds a rowspan open when it covers the row's last columns`, so it guards the new layout too.

`collector/__tests__/processLink/helpers/htmlToMarkdown.test.js`: 54 passed. On a496e21 the five new cases fail and everything else passes. Both copies of `addTableRules` are byte-identical again.

L4XB:
Pushed `cbc34977`, one more gap in the same rules, found in review of the identical code elsewhere. When the spans are over budget, the delimiter counted only the header's own cells, so a body row wider than the header fell outside the table: `<tr><th>A</th><th>B</th></tr>` over a `colspan="1000"` row and a three-cell row gave a two-column delimiter. The header is now padded to the widest row, in both converters. Body rows are left as they are, because GFM fills a short row with empty cells and padding every row would grow with the square of the table. Two new cells, both red without the change; `htmlToMarkdown.test.js` 56 passed.

L4XB:
Follow-up for `rowspan="0"`.

A `rowspan="0"` was read as a span of one row: parsing the attribute gives 0, and only a positive value counted. Chrome reports `element.rowSpan === 0` for it and lays the cell out to the end of its row group, so this table

```html
<table><tbody>
<tr><td rowspan="0">A</td><td>b1</td></tr>
<tr><td>b2</td></tr>
<tr><td>b3</td></tr>
</tbody></table>
```

came out of the converter as

```
| A | b1 |
| --- | --- |
| b2 | |
| b3 | |
```

with `b2` and `b3` in the first column instead of the second, while the same table written with `rowspan="3"` was already correct.

A `0` now reaches the rows left in the cell's own `<thead>`, `<tbody>` or `<tfoot>`, or in the table when the rows are not wrapped in a group, so the output matches the `rowspan="3"` version exactly:

```
| A | b1 |
| --- | --- |
| | b2 |
| | b3 |
```

`colspan="0"` stays one column, which is where HTML5 left it, and a value like `colspan="12abc"` still keeps its leading digits. The grid budget is unchanged: `rowspan="0" colspan="1000"` repeated over 401 rows produces the same 3513-character span-free fallback as before, and a plain 300-row table converts byte-for-byte identically.

Two of the four new tests in `collector/__tests__/processLink/helpers/htmlToMarkdown.test.js` fail on the previous commit (2 failed, 58 passed) and pass now (60 passed); the other two guard `colspan="0"` and the budget. Full suite: 62 suites / 1077 tests passing, and `yarn lint:check` in `collector` is clean. The same change is in the `open-computer` copy of the converter, which this branch keeps byte-identical to the collector one; that copy is outside jest, so I checked it by piping the table above through its CLI.

timothycarambat:
Next time you submit a PR please ensure that you have addressed **all possible** changes. It is a massive time sink to review several PRs all doing the same thing instead of 1 PR that covers all the cases. Reviewing takes a lot of time and failing to capture to full scope of an issue is a big issue and looks like PR farming.

This is the **third** PR around just this chunk of code. All of which have to do with the exact same class of issue.

L4XB:
Fair, and I'll take it. After your a496e213 I pushed three more rounds at `addTableRules` — 58c1f716, cbc34977, bef4997d — each sent over as the case surfaced instead of held until I had swept the rule end to end. Three separate pushes onto a PR you were already reviewing is the same drain, and that part is mine.

Where they came from, since it bears on the fix: 58c1f716 was three defects found in review of the identical rules on qwen-code, cbc34977 came out of review of the same code elsewhere, and `rowspan="0"` in bef4997d is the one I went looking for myself. Porting each one here the day it surfaced instead of batching them is what turned one review into several — my call, and the wrong one.

The section order and the nested table in 3329e36d are yours; I hadn't covered either. I'm not going to claim that finished the surface, because I still haven't swept this rule end to end and don't know what else is in it. Next time the sweep — every attribute the rule reads, both copies, ragged and hostile input — goes before the PR.

Nothing further is queued against this repo. If something does turn up in this converter it goes in an issue, not a PR.

PR #6399, opened 2026-09-17
fix: move default datetime to the latest user message to preserve prefix caching

### LINKED ISSUE #6398: [BUG]: Default `{datetime}` in system prompt breaks prefix cache reuse across minute boundaries
### What happened?

#6016 (addressing #6014) added this sentence to the default system prompt:

> The current date and time is {datetime}.

`{datetime}` expands to `moment().format("LLLL")`, which changes at every minute boundary. The expansion runs on every request in both the chat and agent system-prompt paths.

Prompt caching is prefix-based (Anthropic `cache_control`, OpenAI automatic prefix caching, KV-cache reuse in llama.cpp / vLLM / Ollama). Because the timestamp sits near the beginning of the system prompt — the first content of every request — a minute change invalidates the cached prefix from that position onward: the rest of the system prompt, RAG context, memories, and the entire chat history. Consecutive turns rarely land in the same minute, so in practice every turn is a miss for every workspace on the default prompt. With `ANTHROPIC_CACHE_CONTROL` enabled, each turn additionally pays the cache-write premium with effectively no hits.

#6014 noted that "the small amount of tokens should have very little impact". That holds for token *count*; the problem is their *position*.

Relevant code:
- `server/models/systemSettings.js` — `saneDefaultSystemPrompt`
- `server/models/systemPromptVariables.js` — `{datetime}` expansion
- `server/utils/chats/index.js` — `chatPrompt()` (chat path)
- `server/utils/agents/aibitat/providers/ai-provider.js` — `Provider.systemPrompt()` (agent path)

### Steps to reproduce

1. Use a workspace on the default system prompt.
2. Build its system prompt with `chatPrompt(workspace)` (or `Provider.systemPrompt({ workspace })` for the agent path).
3. Advance the clock across a minute boundary and build it again.
4. Compare: the two strings differ near the beginning, even though the workspace configuration is unchanged.

This demonstrates the changing prefix; actual cache behavior can be confirmed with provider-side cache metrics.

### Expected behavior

Everything before the latest user message stays byte-identical between turns, while the model still receives the current date and time.

One way to achieve this is to append the current datetime to the latest user message instead of embedding it in the system prompt — the last user message is new on every turn anyway.

### Affected workspaces

- Workspaces whose `openAiPrompt` is `null` and resolve to the default at runtime.
- Workspaces created since #6016 without a custom prompt: `Workspace.new()` persists the default text into `openAiPrompt` (`server/models/workspace.js`), so changing the default constant alone does not fix existing records.

Custom prompts that explicitly use `{datetime}` or other time variables are out of scope: that is an opt-in the user made, and #6016 did not change it.

cyclopentadiene:
Fixes #6398

## Problem

#6016 added `The current date and time is {datetime}.` near the beginning of the default system prompt. When the minute changes, the expanded timestamp changes, preventing reuse of the previously cached prefix from that point onward — including otherwise unchanged instructions, context, and conversation history. See the issue for details.

## Fix

Keep the default system prompt static and append the current datetime to the latest user message instead. This preserves the stable prefix before that message while retaining the existing datetime format.

The suffix is not persisted. On the next turn, replayed history omits the previous timestamp, so prefix reuse can stop at the previous user message's suffix position. Preserving that suffix across turns is outside this change.

- New `server/utils/helpers/chat/prompt.js` owns the default prompt text plus `resolveSystemPrompt()`, `getPromptDatetime()` and `appendPromptDatetime()`.
- `saneDefaultSystemPrompt` drops the `{datetime}` sentence. Both system-prompt entry points (`chatPrompt()` and `Provider.systemPrompt()`) go through `resolveSystemPrompt()`.
- Chat call sites (`stream.js`, `apiChatHandler.js`, `embed.js`, `openaiCompatible.js`, `telegramBot/chat/stream.js`) wrap only the `userPrompt` passed to `compressMessages()`. The stored prompt, vector search input and replayed history are untouched. `openaiCompatible` skips the suffix when the client supplies its own system message.
- Agent path: `AIbitat.reply()` appends the datetime to the outgoing user turn in the same block that already injects parsed-file context. Existing `_chats` entries are never mutated, so persisted history and regenerate keep the original text; the value is computed once per reply so recursive tool calls share the prefix.

## Behavior notes

- **Custom prompts are unchanged.** A workspace with a custom `openAiPrompt` gets no suffix; explicit `{datetime}` / `{time}` variables keep expanding in place as before #6016.
- **Workspaces that already have the #6016 default persisted in `openAiPrompt`** (`Workspace.new()` stores the default verbatim) are recognised at runtime by `resolveSystemPrompt()` — no data migration, no rewrite of user settings. Those workspaces may still appear as "modified" in settings until their stored prompt is reset or updated to the current default.

## Tests

- `server/__tests__/utils/helpers/chat/prompt.test.js` (new): chat and agent system prompts are byte-identical across a minute boundary for `null`, the new default, the #6016 default and whitespace-padded variants; a custom prompt containing `{datetime}` still changes (control); history passed to the model is unchanged and only the final user message carries the timestamp; `AIbitat.reply()` unit tests with mocked execution cover routing input, undecorated history, the decorated last turn and unchanged existing `_chats` entries for `undefined` / `null` / custom workspaces.
- `server/__tests__/utils/chats/apiChatHandler.test.js`: `chatSync` / `streamChat` persist the original prompt and send the suffixed one.
- `server/__tests__/utils/agents/defaults.test.js`: removed a now-dead `{datetime}` replacement in a mock.

Full server suite: 49 suites / 677 tests passing.

## Out of scope

- Replaying the timestamp for earlier turns in history.
- Rewriting persisted #6016 defaults and the "modified" indicator on the settings page.


timothycarambat:
Thanks for the thorough writeup and tests. The analysis in #6398 is correct, but after tracing where the prefix actually breaks we don't want to take this as a standalone fix.

The datetime is one of **several** things that rewrite the system message on every turn. RAG context is appended to the system prompt in nearly every provider's `constructPrompt`, memories are appended there too, and the Anthropic `cache_control` breakpoint sits on that same block. Any workspace with documents or memories misses the cache regardless of the timestamp, so this change only helps pure-chat workspaces on the default prompt, while adding the same two-call pattern to seven call sites.

We'd rather fix the structure once. Tracking that in #6401. Closing this, but the tests and the "recognize persisted defaults" idea are useful and will carry over.

